Issue #29, opened 2023-03-07
Implement the RPC to interact with Starknet wrapped blocks


abdelhamidbakhta:
# Description

Implement the RPC server to query the Starknet wrapped blocks.

See:
- [frontier](https://github.com/paritytech/frontier)
- [moonbeam](https://github.com/PureStake/moonbeam)

See also this issue: [#20](https://github.com/keep-starknet-strange/kaioshin/issues/20)

PR #97, opened 2023-03-28
feat: expose rpc server 

drspacemn:
Implement the starknet RPC spec:
https://github.com/starkware-libs/starknet-specs/tree/master/api

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #29 

# What is the new behavior?

- declare and implement runtime APIs to expose the StarkNet pallet
- define a starknet JSON-RPC api
- implement the starknet JSON-RPC api

- [ ] Yes
- [x] No


drspacemn:
Test RPC endpoints:
`curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "starknet_blockHashAndNumber"}' http://localhost:9933/`

`curl -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "starknet_blockNumber"}' http://localhost:9933/`

crates/pallets/starknet/src/lib.rs
```diff
@@ -485,6 +489,16 @@ pub mod pallet {
 
     /// The Starknet pallet internal functions.
     impl<T: Config> Pallet<T> {
+        /// Get current block number.
+        ///
+        /// # Returns
+        ///
+        /// The current block number.
+        #[inline(always)]
+        pub fn current_block_number() -> Option<U256> {
```
EvolveArt, line 498: I would rather stay with only `current_block` and do the mapping in the runtime `lib.rs`

crates/pallets/starknet/Cargo.toml
```diff
@@ -30,6 +30,7 @@ frame-system = { workspace = true }
 frame-benchmarking = { workspace = true, optional = true }
 
 # Substrate primitives
+sp-api = { default-features = false, git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.40" }
```
EvolveArt, line 33: we should probably add it to the parent `cargo.toml` like we do for the other crates

crates/pallets/starknet/src/lib.rs
drspacemn: moved to runtime lib

crates/pallets/starknet/Cargo.toml
drspacemn: moved to parent `cargo.toml`

EvolveArt (approved):
Did you check with @tdelabro ? If he is starting from this branch then let's merge it, lgtm

tdelabro:
I use part of the code but in other files. So no need to merge. I took what was good and it will show up in my PR

PR #101, opened 2023-03-28
Use storage to define ETH RPC URL

DavideSilva:
# Pull Request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: N/A

# What is the new behavior?

-
-
-

- [ ] Yes
- [ ] No

# Other information

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -38,6 +38,8 @@ pub use self::pallet::*;
 
 pub(crate) const LOG_TARGET: &str = "runtime::starknet";
 
+pub const ETHEREUM_MAINNET_RPC: &[u8] = b"starknet::ETHEREUM_MAINNET_RPC";
+
 // syntactic sugar for logging.
 #[macro_export]
 macro_rules! log {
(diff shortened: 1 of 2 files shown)


DavideSilva:
Hey @abdelhamidbakhta, just to check this was the intended result of setting RPC configuration in offchain storage

Next step is to replace the hardcoded `StorageValueRef::persistent(ETHEREUM_MAINNET_RPC).set(&"https://ethereum-mainnet-rpc.allthatnode.com")` with an RPC call that sets the value

EvolveArt:
Probably we should add a `NETWORK` option such that we can easily switch between `MAINNET` and `GOERLI` RPC urls. wdyt ?

DavideSilva:
I think adding a way to switch between networks is a good idea but I'm not sure at what level that switch should be made.
I'm guessing changing the network at runtime is a no go, right? So, should that decision depend on the config?

abdelhamidbakhta:
> 

Let's not consider this option for now and stick with the configuration of the URL directly.
I suggest we go with:
`ETHEREUM_EXECUTION_RPC` and `ETHEREUM_CONSENSUS_RPC`.

Issue #93, opened 2023-03-26
dev: Remove `pallet::without_storage_info` 🦀


EvolveArt:
The `ContractClassWrapper` uses a `BTreeMap` to store entrypoints.

We should switch to a `[BoundedBTreeMap](https://crates.parity.io/frame_support/storage/bounded_btree_map/index.html)` as recommended by the substrate repo [here](https://github.com/paritytech/substrate/issues/8629).

On a general note, we should clearly state in some docs the boundaries for bounded values e.g `MaxProgramSize`, `MaxCalldataSize` etc

PR #104, opened 2023-03-29
fix: entry_points_by_type 🦀

EvolveArt:
# Pull Request type

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #93 

# What is the new behavior?

While working on benchmark and e2e testing scripts in Typescript with PolkadotJS. I saw that sending `entrypoint` as bytes from the JSON just like we do for the `program` in the contract class works, whereas trying to send a BTreeMap would always fail for `MAX_SAFE_INTEGER` boundaries reasons in polkadot.js
Anyway this behavior is just more consistent and overall makes more sense.


abdelhamidbakhta (approved):
lgtm

PR #106, opened 2023-03-29
feat: Transaction Events ✨

EvolveArt:
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

No events are emitted on transaction's execution.

Issue Number: NaN

# What is the new behavior?

- Pallet Events would be emitted within which there would be a `Starknet` type event `keys`/`data`/`from_address`
- Event data is taken from the execution `CallInfo`

- [ ] Yes
- [x] No

# Other information

[Events Starknet Documentation](https://docs.starknet.io/documentation/architecture_and_concepts/Events/starknet-events/)


crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -40,19 +45,60 @@ impl Event {
...
             keys: BoundedVec::try_from(vec![one, one]).unwrap(),
             data: BoundedVec::try_from(vec![one, one]).unwrap(),
-            from_address: one,
+            from_address: ContractAddressWrapper::from(one),
+        }
+    }
+}
+
+impl From<Event> for EventWrapper {
+    fn from(event: Event) -> Self {
```
abdelhamidbakhta, line 75: please add test for this function

crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -40,19 +45,60 @@ impl Event {
...
+                event.content.data.0.iter().map(|d| H256::from_slice(d.bytes())).collect::<vec::Vec<H256>>(),
+            )
+            .unwrap(),
+            from_address: event.from_address.0.key().bytes().try_into().unwrap(),
+        }
+    }
+}
+
+impl From<EventContent> for EventWrapper {
+    fn from(event: EventContent) -> Self {
```
abdelhamidbakhta, line 91: please add rust doc

crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -40,19 +45,60 @@ impl Event {
...
+        Self {
+            keys: BoundedVec::try_from(
+                event.keys.iter().map(|k| H256::from_slice(k.0.bytes())).collect::<vec::Vec<H256>>(),
+            )
+            .unwrap(),
```
abdelhamidbakhta, line 96: since it can fail it should be a try_from and return a result. we should avoid using unwrap

abdelhamidbakhta (changes requested):
Thanks. Left some comments

abdelhamidbakhta (approved):
lgtm

Issue #111, opened 2023-03-30
feat: test transaction helper 🧪 


EvolveArt:
# Feature Request

**Describe the Feature Request**

While writing tests for the `starknet` pallet, we have to manually create new transactions by using its constructor `Transaction::new`. This is very cumbersome and affects tests' readability overall.

**Describe Preferred Solution**

- Add a helper function/macro that would generate a `Transaction` object from a json file. 
- Add a `transactions` subfolder in `ressources`.
- Refractor related tests with the helper function.

**Related Code**

- [Example of cumbersome test](https://github.com/keep-starknet-strange/madara/blob/main/crates/pallets/starknet/src/tests.rs#L186)

Matth26:
I'll make a proposal for this one

abdelhamidbakhta:
> I'll make a proposal for this one

could, assigned it to you!

PR #143, opened 2023-04-03
Add transaction_from_json helper function and refactor StarkNet pallet tests

Matth26:
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

While writing tests for the starknet pallet, new transactions have to be manually created by using its constructor Transaction::new. This is very cumbersome and affects tests' readability overall.

Issue Number: #111 

# What is the new behavior?

- A new helper function, `transaction_from_json`, has been added. This function allows users to create Transaction objects directly from json files. 
- json transaction files are now stored in a dedicated `/resources/transactions` subfolder. 

- [ ] Yes
- [x] No

# Other information

All elements of the Transaction come from the json except for the `pub contract_class: Option<ContractClassWrapper>` field. 

The existing tests use a file from the blockifier crate, imported with the `include_bytes!` macro. So I opted to exclude it from the json and incorporate it as a function argument instead:

```Rust
pub fn transaction_from_json(json_str: &str, contract_content: &'static [u8])
```

If `contract_content` is set to `&[]`, the `contract_class` field will be set to `None`. Otherwise, we create the `ContractClassWrapper` from the `contract_content`.


EvolveArt (changes requested):
Thanks for this! A few changes I'd like to see tho : 
- You could put the code from `transaction_from_json` in the `starknet_serde` primitive module.
- No need to rewrite all the tests, just modifying the existing ones to use the helper function will be fine
- I would like to see tests only on serialization/derserialization. Edge cases where it fails should be covered there.
- Finally would be great to have more explicit names about what the transactions actually do instead of referring to test names

crates/primitives/starknet/src/starknet_serde/mod.rs
```diff
@@ -1 +1,321 @@
...
+use blockifier::test_utils::get_contract_class;
+use frame_support::BoundedVec;
+use hex::{FromHex, FromHexError};
+use serde::{Deserialize, Serialize};
+use sp_core::{H256, U256};
+
+use crate::execution::{CallEntryPointWrapper, ContractAddressWrapper, ContractClassWrapper, EntryPointTypeWrapper};
+use crate::transaction::types::{EventWrapper, MaxArraySize, Transaction};
+
+fn remove_prefix(input: &str) -> &str {
```
abdelhamidbakhta, line 15: please add rust doc

abdelhamidbakhta (changes requested):
left few comments

EvolveArt (approved):
lgtm

abdelhamidbakhta (approved):

Issue #92, opened 2023-03-26
feat: Implement Contract Salt 🧂


EvolveArt:
# Feature Request

**Describe the Feature Request**

Contract Salt is currently hardcoded and should rather be sent within the `deploy` transaction.

**Additional Context**

https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/transactions/#deploy_account_transaction

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

- [ ] Yes
- [ ] No

PR #167, opened 2023-04-06
feat: contract salt

drspacemn:
Allow Contract Salt to be included in Deploy transactions instead of hard coded to default.

# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

contract address salt is hardcoded 

Issue Number: #92 

# What is the new behavior?

- contract address salt can be set by the Deploy transaction

- [ ] Yes
- [x] No


EvolveArt (approved):
lgtm

abdelhamidbakhta:
Ready to merge when conflict is resolved @drspacemn

abdelhamidbakhta (approved):
lgtm

crates/pallets/starknet/src/mock.rs
```diff
@@ -70,14 +70,16 @@ pub fn new_test_ext() -> sp_io::TestExternalities {
...
+    let arg_account_class = get_contract_class(include_bytes!("../../../../ressources/account_compiled.json"));
     let test_class = get_contract_class(include_bytes!("../../../../ressources/test.json"));
     let l1_handler_class = get_contract_class(include_bytes!("../../../../ressources/l1_handler.json"));
 
     // ACCOUNT CONTRACT
     let contract_address_str = "02356b628D108863BAf8644c945d97bAD70190AF5957031f4852d00D0F690a77";
     let contract_address_bytes = <[u8; 32]>::from_hex(contract_address_str).unwrap();
 
-    let class_hash_str = "025ec026985a3bf9d0cc1fe17326b245dfdc3ff89b8fde106542a3ea56c5a918";
+    let class_hash_str = "025ec026985a3bf9d0cc1fe17326b245dfdc3ff89b8fde106542a3ea56c5a919";
```
drspacemn, line 82: @EvolveArt this change isn't very descriptive. I'd like to change it before you merge. 

This hash doesn't actually apply to the contract located @ `ACCOUNT_CONTRACT_PATH` so this value essentially just serves as a lookup key, but incrementing it by one is confusing. 

I'll make this more transparant.

EvolveArt: Right I think we do not need to have such complicated values but rather simple ones like the [blockifier](https://github.com/tdelabro/blockifier/blob/no_std-support/crates/blockifier/src/test_utils.rs#L34) does.

drspacemn: Okay attempted to simplify the testing and use actual contract class hashes where applicable.

One thing I don't understand is why blockifier testing isn't using the actual contract_class_hash if they can.

EvolveArt: Fair question to be honest not sure I know exactly why

abdelhamidbakhta (approved):
Lgtm

PR #217, opened 2023-04-11
fix: change ts package name ✍️

EvolveArt:


tests/package.json
```diff
@@ -31,7 +31,7 @@
   },
   "scripts": {
     "test-with-logs": "mocha --printlogs -r ts-node/register 'tests/**/test-*.ts'",
-    "setup-typescript-api": "rm -rf node_modules/@madara/api-augment && cd ../typescript-api && npm run build && cp -r build ../tests/node_modules/@madara/api-augment",
+    "setup-typescript-api": "rm -rf node_modules/@keep-starknet-strange/madara-api-augment && cd ../typescript-api && npm run build && cp -r build ../tests/node_modules/@keep-starknet-strange/madara-api-augment",
```
abdelhamidbakhta, line 34: this should not be needed anymore, right ? unless we want to keep it for local development of the typescript API before publishing, but in this case I would make it explicit that it is for local dev purposes

abdelhamidbakhta (approved):
left a minor comment, otherwise looks good to me

Issue #188, opened 2023-04-10
feat: RPC - starknet_getBlockWithTxHashes


abdelhamidbakhta:
Get block information with transaction hashes given the block id

dubzn:
Hey Abdel, can i take this one?

abdelhamidbakhta:
> Hey Abdel, can i take this one?

Sure, assigned to you

PR #274, opened 2023-04-19
feat: RPC - `starknet_getBlockWithTxHashes`

dubzn:
# Pull Request type

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #188

# What is the new behavior?

- add starknet_getBlockWithTxHashes rpc endpoint.

- [ ] Yes
- [x] No

# Other information

Response example:
image


crates/client/rpc/src/lib.rs
```diff
@@ -196,13 +199,81 @@ where
...
+        let sequencer_address =
+            String::from_utf8(hex::encode(block.header().sequencer_address).into()).map_err(|e| {
+                error!("SequencerAddress: Failed to convert 'sequencer_address' attribute to String: {e}");
+                StarknetRpcApiError::InternalServerError
+            })?;
+
+        let block_with_tx_hashes = BlockWithTxHashes {
+            transactions: txs_hashes,
+            // TODO: Status hardcoded, get status from block
+            status: BlockStatus::Pending,
```
dubzn, line 246: Where could I get it from? I think it is not implemented, right?

crates/client/rpc-core/src/lib.rs
```diff
@@ -39,6 +39,54 @@ pub struct FunctionCall {
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum MaybePendingBlockWithTxHashes {
+    Block(BlockWithTxHashes),
+    PendingBlock(PendingBlockWithTxHashes),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BlockWithTxHashes {
+    pub status: BlockStatus,
+    pub block_hash: FieldElement,
```
EvolveArt, line 52: rustdoc

crates/client/rpc/src/lib.rs
```diff
@@ -196,13 +199,81 @@ where
...
+        let block = self
+            .overrides
+            .for_block_hash(self.client.as_ref(), substrate_block_hash)
+            .current_block(substrate_block_hash)
+            .unwrap_or_default();
+
+        let txs_hashes = block.transactions_hashes().iter().map(|tx| tx.to_string()).collect();
+        let sequencer_address =
+            String::from_utf8(hex::encode(block.header().sequencer_address).into()).map_err(|e| {
+                error!("SequencerAddress: Failed to convert 'sequencer_address' attribute to String: {e}");
```
EvolveArt, line 239: can we use `try_into` here ?

crates/client/rpc/src/lib.rs
EvolveArt: I don't think so, basically if it is in `current_block` mapping then it means it's been "accepted on L2". Pending transactions are stored in the "pending" storage map. Probably we could add an internal function in the pallet that constructs the pending block out of this mapping or you can do it in a helper function client-side.

EvolveArt (changes requested):
Thanks a lot for this contribution @dubzn ! Left a few comments, overall just need to add more integration tests too

crates/client/rpc/src/lib.rs
dubzn: Sure! its done :heavy_check_mark:

crates/client/rpc-core/src/lib.rs
dubzn: :heavy_check_mark:

crates/client/rpc/src/lib.rs
dubzn: That makes sense. If you think it's necessary, I can add that method to the pallet and the runtime. It would return all pending transactions that have not yet been assigned a block, and return the `PendingBlockWithTxHashes` struct:

```
pub struct PendingBlockWithTxHashes {
    /// The hashes of the transactions included in this block
    pub transactions: Vec<FieldElement>,
    /// The time in which the block was created, encoded in Unix time
    pub timestamp: u64,
    /// The Starknet identity of the sequencer submitting this block
    pub sequencer_address: FieldElement,
    /// The hash of this block's parent
    pub parent_hash: FieldElement,
}
```

abdelhamidbakhta:
hey @dubzn what is the status on this PR ?

dubzn:
> hey @dubzn what is the status on this PR ?

hey Abdel, thanks for ask! I'm currently working on this, but I'm stuck with the integration tests. I've been talking to @EvolveArt and @tdelabro to see how I can make progress, but I haven't been able to finish it yet. The 'problem' I'm having [is explained here](https://github.com/keep-starknet-strange/madara/pull/274/files?diff=unified&w=0#diff-324c72286b679b662255c8c6d3db890cb89784828b64cf7c34fe1d8b12ffc50aR86), but to summarize: when creating a block with transactions, for some reason, it's not found when requested through the provider. I understand that for each Substrate block, we have a corresponding Starknet block. I think where I'm failing is that when I create a block, it gives me the hash of a Substrate block, but when I call getBlockWithTxsHashes(), I need the hash of the corresponding Starknet block, but I couldn't find an example of the conversion. Today, I was planning to work on implementing the conversion of StarknetBlock to SubstrateBlock (already implemented in .rs but not in .ts) and vice versa. The tests that are currently implemented are basic because they simply check that the response is not 'undefined'.

abdelhamidbakhta:
> > hey @dubzn what is the status on this PR ?
> 
> hey Abdel, thanks for ask! I'm currently working on this, but I'm stuck with the integration tests. I've been talking to @EvolveArt and @tdelabro to see how I can make progress, but I haven't been able to finish it yet. The 'problem' I'm having [is explained here](https://github.com/keep-starknet-strange/madara/pull/274/files?diff=unified&w=0#diff-324c72286b679b662255c8c6d3db890cb89784828b64cf7c34fe1d8b12ffc50aR86), but to summarize: when creating a block with transactions, for some reason, it's not found when requested through the provider. I understand that for each Substrate block, we have a corresponding Starknet block. I think where I'm failing is that when I create a block, it gives me the hash of a Substrate block, but when I call getBlockWithTxsHashes(), I need the hash of the corresponding Starknet block, but I couldn't find an example of the conversion. Today, I was planning to work on implementing the conversion of StarknetBlock to SubstrateBlock (already implemented in .rs but not in .ts) and vice versa. The tests that are currently implemented are basic because they simply check that the response is not 'undefined'.

ok thanks for the detailed answer. @EvolveArt please help @dubzn tomorrow if you can. if you cant I will have a look later this week and help you

LucasLvy (approved):
lgtm

PR #286, opened 2023-04-22
feat(pallet_starknet): use a more precise state reader

tdelabro:
# Pull Request type

- [x] Feature

The whole blockchain state is loaded in the RAM each time we interact with it through the blockifier `StateReader`.

- only the values that are actually accessed are loaded into RAM, on demand

- [x] No

# Other information

Tx per block went from: 12.56 to 19.33 on my machine


crates/pallets/starknet/src/tests/blockifier_state_reader.rs
```diff
@@ -0,0 +1,88 @@
...
+        System::set_block_number(0);
+        run_to_block(2);
+
+        let none_origin = RuntimeOrigin::none();
+        let (account_addr, _, _) = account_helper(TEST_ACCOUNT_SALT);
+
+        let contract_class = get_contract_class(ERC20_CONTRACT_PATH);
+        let erc20_class = ContractClassWrapper::from(contract_class.clone());
+        let erc20_class_hash =
+            <[u8; 32]>::from_hex("057eca87f4b19852cfd4551cf4706ababc6251a8781733a0a11cf8e94211da95").unwrap();
```
Eikix, line 29: should this be a test util constant? It seems it'll be reused in many tests?

crates/pallets/starknet/src/tests/blockifier_state_reader.rs
```diff
@@ -0,0 +1,88 @@
...
+            ..Transaction::default()
+        };
+
+        assert_ok!(Starknet::declare(none_origin, transaction));
+
+        let mut state_reader = BlockifierStateReader::<Test>::new();
+        let retrieved_contract_class =
+            (*state_reader.get_contract_class(&ClassHash(StarkFelt(erc20_class_hash))).unwrap()).clone();
+
+        assert!(retrieved_contract_class.abi.is_none());
```
Eikix, line 51: why is the ABI none:) ?

Issue #204, opened 2023-04-10
feat: RPC - starknet_pendingTransactions


abdelhamidbakhta:
Returns the transactions in the transaction pool, recognized by this sequencer

ZePedroResende:
Hi, i would like to try and tackle this one !

abdelhamidbakhta:
> Hi, i would like to try and tackle this one !

Sure, assigned to you

PR #293, opened 2023-04-24
[WIP] feat: Pending Transaction RPC

ZePedroResende:
# Pull Request type

- [ ] Bugfix
- [ X] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #204 

# What is the new behavior?

- Add starknet_pendingTransactions rpc endpoint

- [ ] Yes
- [X] No

# Other information

This is still a WIP, its not on its final state and its missing tests and documentation.12

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -42,6 +42,8 @@ use sp_std::prelude::*;
 use sp_version::RuntimeVersion;
 /// Import the types.
 pub use types::*;
+use mp_starknet::transaction::types::Transaction;
+
 
 // Create the runtime by composing the FRAME pallets that were previously configured.
 construct_runtime!(
@@ -275,6 +277,13 @@ impl_runtime_apis! {
         fn call(address: ContractAddressWrapper, function_selector: H256, calldata: Vec<U256>) -> Result<Vec<StarkFeltWrapper>, DispatchError> {
             Starknet::call_contract(address, function_selector, calldata)
         }
+
+		fn pending_transactions() -> Vec<Transaction> {
+			let pending = Starknet::pending();
+            let transactions: Vec<Transaction> = pending.into_iter().map(|(transaction, _)| transaction).collect();
+
+			transactions
+		}
     }
 
     #[cfg(feature = "runtime-benchmarks")]
--- a/crates/client/rpc-core/src/lib.rs
+++ b/crates/client/rpc-core/src/lib.rs
@@ -63,4 +63,8 @@ pub trait StarknetRpcApi {
     /// Call a contract function at a given block id
     #[method(name = "call")]
     fn call(&self, request: FunctionCall, block_id: BlockId) -> RpcResult<Vec<String>>;
+	
+    /// Returns the transactions in the transaction pool, recognized by this sequencer
+	#[method(name = "pendingTransactions")]
+    fn pending_transactions(&self) -> RpcResult<Vec<String>>;
 }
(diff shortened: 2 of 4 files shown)


ClementWalter:
I'm taking up this work

tests/tests/test-rpc/test-starknet-rpc.ts
```diff
@@ -541,4 +541,39 @@ describeDevMadara("Starknet RPC", (context) => {
...
+  it("pendingTransactions", async function () {
+    await rpcTransfer(
+      providerRPC,
+      ARGENT_CONTRACT_NONCE,
+      ARGENT_CONTRACT_ADDRESS,
+      MINT_AMOUNT
+    );
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const txs: { type?: any; sender_address?: string; calldata?: string[] }[] =
```
ClementWalter, line 554: why not putting a type `string` or `enum` ? is the type returned from starknet-js wrong?

tests/tests/test-rpc/test-starknet-rpc.ts
```diff
@@ -541,4 +541,39 @@ describeDevMadara("Starknet RPC", (context) => {
...
+        hash.getSelectorFromName("transfer"),
+        0,
+        3,
+        3,
+        ARGENT_CONTRACT_ADDRESS,
+        MINT_AMOUNT,
+        0,
+      ].map(toHex)
+    );
+  });
```
ClementWalter, line 578: I'd rather have an `expected_results` or `expected_pending_txs` object and assert only only the full result against the full expected result. Otherwise we're only partially testing the output

ClementWalter (commented):
I cannot tell so much about the substrate part because I'm too new to the framework, but left 2 minor comments about the ts

crates/runtime/src/lib.rs
```diff
@@ -322,6 +323,13 @@ impl_runtime_apis! {
         fn get_hasher() -> Hasher {
             Starknet::get_system_hash().into()
         }
+
+        fn extrinsic_filter(xts: Vec<<Block as BlockT>::Extrinsic>) -> Vec<Transaction>{
```
abdelhamidbakhta, line 327: 2 questions:
- could this be done entirely client side ?
- we should include `DeclareTransaction` and `DeployAccountTransaction` extrinsics as well, not only `InvokeTransaction`, right ?

abdelhamidbakhta (commented):
left some comments

crates/runtime/src/lib.rs
abdelhamidbakhta: After discussion with @tdelabro, it's fine to keep this logic in the runtime. @tdelabro to document the rationale

ClementWalter: @ZePedroResende please see this comment and add some doc/comment

ZePedroResende: > * we should include `DeclareTransaction` and `DeployAccountTransaction` extrinsics as well, not only `InvokeTransaction`, right ?

Regarding this @abdelhamidbakhta @tdelabro, is this correct ?

ClementWalter (commented):
It looks like the state is not reset in between tests so the `expect(tx).to.deep.equal` doesn't work. Let's not block this PR because of of, so sorry for the comment, you can check only what is stable across run and add an issue

Furthermore, can you rebase and update the description of your test to be like the others
```
describe('pendingTransaction', async () => {
    it("should ...", async () => {
```

crates/runtime/src/lib.rs
abdelhamidbakhta: Yes we should add the two other types of transactions

abdelhamidbakhta (approved):
lgtm

PR #308, opened 2023-04-27
Clean code/no unwrap

tdelabro:
# Pull Request type

- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)

Too much logic hang in `lib.rs`, and inside the `pallet` module.

# What is the new behavior?

- offchain worker logic is in it's own module
- `transfer_fees` moved to the `starknet_fee` module
- `impl<T: Config> Pallet<T>` moved out of the `pallet` module in `lib.rs`
- less unwrap in `lib.rs`
- `query_eth` return raw `Vec<u8>` instead of converting it to String. It was not needed as we can serialize form bytes as well as from string
- use `TryFrom` rather than standalone method for type conversion

- [ ] Yes
- [x] No


LucasLvy (approved):
lgtm

Issue #192, opened 2023-04-10
feat: RPC - starknet_getTransactionByHash


abdelhamidbakhta:
Get the details and status of a submitted transaction

glihm:
@abdelhamidbakhta I can try to take this one to dive into how madara runtime works. 👍

abdelhamidbakhta:
> @abdelhamidbakhta I can try to take this one to dive into how madara runtime works. 👍

Sure thing, assigned to you.

glihm:
Quick question, as I am going through my PR for this issue, I was wondering what is the most efficient way to retrieve transaction by hash.

At the moment, we store the blocks with all their transactions. Does it mean I have no choice of iterating on all blocks using `getBlockWithTxHashes` to find the transaction the user asked for?

Do you think it is worthy to spend some time on finding an efficient way to keep somehow an index of transactions? Or did I miss it?

Thanks a lot for the feedbacks.

(For now I am doing a very naïve implementation).

glihm:
Also to confirm, for now I can return a `Transaction`, but once the storage is implemented for `TransactionReceipWrapper` this should be the returned object, right?

glihm:
@abdelhamidbakhta merged in #323. :+1:

PR #323, opened 2023-05-02
feat: RPC - `getTransactionByHash` - `getTransactionReceipt`

glihm:
# Pull Request type

- [ ] Bugfix
- [X] Feature
- [ ] Build-related changes
- [ ] Testing

Issue Number: #192 

# What is the new behavior?

Add support for `starknet_getTransactionByHash`.
This work is on progress for several reasons:
- Transaction receip is expected to be returned, but not yet present in the block.
- Still working on how data are gathered from the runtime to be sent to substrate client.

- [ ] Yes
- [X] No

New RPC endpoint.

# Other information

I will add some comment on the code, and I'll keep you posted for new content on the PR or if I am stuck.


crates/client/rpc-core/src/lib.rs
```diff
@@ -17,11 +17,13 @@ use hex::ToHex;
 use jsonrpsee::core::RpcResult;
 use jsonrpsee::proc_macros::rpc;
 use mp_starknet::execution::types::{ContractClassWrapper, EntryPointTypeWrapper, EntryPointWrapper, MaxEntryPoints};
+use mp_starknet::transaction::types::Transaction;
 use serde::{Deserialize, Serialize};
 
 pub type FieldElement = String;
 pub type BlockNumber = u64;
 pub type BlockHash = FieldElement;
+pub type TransactionHash = FieldElement;
```
glihm, line 26: Could it be better to propose a `pub type Hash = FieldElement` to not have `TransactionHash`, `BlockHash`, etc...?

crates/client/rpc-core/src/lib.rs
```diff
@@ -254,6 +256,14 @@ pub enum RPCContractClass {
     ContractClass(SierraContractClass),
 }
 
+/// Starknet transaction for RPC serving.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct RpcTransaction {
```
glihm, line 261: I did a new struct to encapsulate internal data and be sure we can have loose coupling between the struct sent in the RPC response, and internal organization. Is that ok?

crates/client/rpc/src/errors.rs
```diff
@@ -36,6 +36,11 @@ pub enum StarknetRpcApiError {
     InternalServerError = 500,
     #[error("Failed to fetch pending transactions")]
     FailedToFetchPendingTransactions = 38,
+
+    // Provided hash is not a valid H256 hash.
+    // This may be reused in several places to check RPC params that are hashes.
+    #[error("Invalid hash")]
+    InvalidHash = 39,
```
glihm, line 43: I propose a new `InvalidHash` that may be more generic for all future Hash related arguments. What do you think?

crates/client/rpc/src/lib.rs
```diff
@@ -214,6 +214,46 @@ where
...
+        &self,
+        tx_hash: TransactionHash,
+    ) -> RpcResult<RpcTransaction> {
+
+        let tx_h256 = H256::from_str(&tx_hash).map_err(|e| {
+            error!("GetTransactionByHash: Failed to convert '{tx_hash}' to H256: {e}");
+            StarknetRpcApiError::InvalidHash
+        })?;
+
+        let substrate_block_hash = self
```
glihm, line 229: Can you confirm that due to the subtrate architecture, this lines of codes are required to do a runtime call? Because several endpoints will not have the block as argument, and we can then make a thin wrapper to abstract this to them.

crates/runtime/src/lib.rs
```diff
@@ -291,6 +292,17 @@ impl_runtime_apis! {
         fn contract_class_by_class_hash(class_hash: ClassHashWrapper) -> Option<ContractClassWrapper> {
             Starknet::contract_class_by_class_hash(class_hash)
         }
+
+        /// Retrieves a transaction's receip from the storage based on the transaction hash.
+        fn get_transaction_by_hash(tx_hash: H256) -> Option<TransactionReceiptWrapper> {
+            // TODO: Naive approach.
```
glihm, line 298: Can you confirm that the block only contains `Transactions` and not the receips. Am I able to find receips directly going into the storage, or for now, I can stick with transactions and I'll update this once receips are present in the block?

tests/tests/test-rpc/test-starknet-rpc.ts
```diff
@@ -80,4 +80,15 @@ describeDevMadara("Starknet RPC", (context) => {
 
     expect(contract_class).to.not.be.undefined;
   });
+
+  it("getTransactionByHash", async function () {
+    // Test bad hash.
+    const tx = await providerRPC.getTransactionByHash("abcd");
```
glihm, line 86: I have an error here:

```
LibraryError: -32602: missing field `tx_hash` at line 1 column 27
```

I am not sure what I am missing into the `.ts` file.

glihm (commented):
Thank you for your time and answers. :pray:

crates/runtime/src/lib.rs
glihm: Sorry, I just seen the other PR for transaction receipt. I'll focus on Transaction.

crates/client/rpc/src/lib.rs
```diff
@@ -517,4 +519,80 @@ where
...
+                    let transaction = InvokeTransactionV1 {
+                        transaction_hash: h,
+                        max_fee: FieldElement::ZERO,
+                        signature: vec![FieldElement::ZERO],
+                        nonce: FieldElement::ZERO,
+                        sender_address: FieldElement::ZERO,
+                        calldata: vec![FieldElement::ZERO],
+                    };
+
+                    // Map the correct transaction base on the transaction type.
```
glihm, line 583: The mapping here is really heavy. I am looking for alternative, if any. And add the TxType rework if I can as suggested by @EvolveArt.

examples/rpc/starknet/starknet_addInvokeTransaction.hurl
```diff
@@ -8,7 +8,7 @@ Content-Type: application/json
             "type": "INVOKE",
             "max_fee":"0xDEADB",
             "version":"0x1",
-            "nonce":"0x1",
+            "nonce":"0x0",
```
glihm, line 11: I need nonce set to 0 to be able to send the transaction as the first transaction.
Not sure if it's a problem from my setup, or something I am missing.

glihm (commented):
Here the mapping in the client works well, and I am exploring the addition of receipts + how types converters must be done as @drspacemn did in his PR. It's a lot of mapping, investigating if there is an easier way to go here.

ClementWalter:
I think that this is blocked until #452 is merged

glihm:
> I think that this is blocked until #452 is merged

It is yes. :+1: 
#459 is the guy now, right?

glihm:
Unlocked by merge of #459, I'll tackle this after work today. :+1:

glihm:
@abdelhamidbakhta @tdelabro @EvolveArt 

On this PR, here is my current state:

1. I am waiting the `pendingTransactions` to be able to re-use the function they did to verify if the transaction is in the transaction pool.
2. On my tests, I have good behavior using hurl files. But in integration testing, I am doing exactly as `getTransactionByBlockIdAndIndex` to submit a transaction, but I can't retrieve it. Also I added a little code to get the tx hash back, but not sure it is the right way.

About the receipts, I have a situation.

1. The receipts are generated when a transaction is executed [here in the runtime](https://github.com/keep-starknet-strange/madara/blob/e664aeb879f957951f9331bc4e526ac0c8e547ce/crates/pallets/starknet/src/lib.rs#L509). After that, it is added to the `Pending` storage.
2. However, as we've seen, this storage is not populated until the block is init / execute txs / finalize. Our block interval time is 6s, this process occurs at like 5.8s. So it's super fast, and the probability we hit the RPC at this time is super small (at least for now).
3. So, during all this time, the transaction is in the transaction pool, where there is no receipt available.

I have talked with @drspacemn, here there is a confusion and I am not sure I'm getting it.

1. When the tx is in the transaction pool, it's status is supposed to be "Received".
2. When the tx is executed by the runtime (which for now it's a very short moment), and stored in `Pending` storage, status should be "Pending" (and a receipt is available).
3. When it's in a finalized block -> "accepted on L2".
4. (I'll skip "accepted on L1" for now)

I am not sure being in the right track as it's still a bit confused, please let me know your take on this.

But if I understood correctly substrate, we have here a transaction status that is missing. Because saying it's pending in the transaction pool, would be incorrect as no receipt is available. But we can stick to it right now, and the receipt will return "NoTxnHashFound".

tdelabro:
@LucasLvy @abdelhamidbakhta 

> But if I understood correctly substrate, we have here a transaction status that is missing. Because saying it's pending in the transaction pool, would be incorrect as no receipt is available.

I don't understand, how could there be a receipt when the tx is still in the pool? Is it how Starknet works right now?

crates/client/rpc/src/lib.rs
```diff
@@ -652,4 +652,113 @@ where
...
+                for (txn, _) in transactions {
+                    if txn.hash != transaction_hash.into() {
+                        continue;
+                    }
+
+                    return Ok(Transaction::try_from(txn.clone()).map_err(|e| {
+                        error!("{:?}", e);
+                        StarknetRpcApiError::InternalServerError
+                    })?);
+                }
```
tdelabro, line 701: Using iterators and `find` is probably more idiomatic. Is there a reason why not to use it?

crates/pallets/starknet/src/lib.rs
```diff
@@ -874,7 +874,13 @@ impl<T: Config> Pallet<T> {
         let sequencer_address = SEQUENCER_ADDRESS;
         let block_timestamp = Self::block_timestamp();
         let transaction_count = pending.len() as u128;
-        let transactions: Vec<Transaction> = pending.into_iter().map(|(transaction, _)| transaction).collect();
+
+        // TODO: need more experienced rust dev here to understand
+        // what is the correct way of doing the next two clones.
```
tdelabro, line 879: What is the problem here exactly?

Did you tried using `iter` rather than `into_iter` and only cloning in the end?

crates/client/rpc/src/lib.rs
```diff
@@ -652,4 +652,113 @@ where
...
+
+                Err(StarknetRpcApiError::TxnHashNotFound.into())
+            }
+            BlockTransactions::Hashes(_hashes) => {
+                // TODO: Here what can I do? I don't have all the info,
+                // so what's the expected result, TxnNotFount? Error?
+                Err(StarknetRpcApiError::TxnHashNotFound.into())
+            }
+        }
+    }
```
tdelabro, line 763: Can probably be factored with the previous one

crates/primitives/starknet/src/block/mod.rs
```diff
@@ -30,7 +30,7 @@ pub enum BlockTransactions {
     /// Only hashes
     Hashes(BoundedVec<Felt252Wrapper, MaxTransactions>),
     /// Full transactions
-    Full(BoundedVec<Transaction, MaxTransactions>),
+    Full(BoundedVec<(Transaction, TransactionReceiptWrapper), MaxTransactions>),
```
tdelabro, line 33: I'm really not sure about this.
https://alpha-mainnet.starknet.io/feeder_gateway/get_block?blockNumber=latest

The receipts seem to be in a separate field of the `Block` struct. Not stored together with the transaction.

@LucasLvy any insight?

tdelabro (changes requested):

crates/primitives/starknet/src/block/mod.rs
LucasLvy: Yes they are different and are not in the same field

LucasLvy:
> @LucasLvy @abdelhamidbakhta
> 
> > But if I understood correctly substrate, we have here a transaction status that is missing. Because saying it's pending in the transaction pool, would be incorrect as no receipt is available.
> 
> I don't understand, how could there be a receipt when the tx is still in the pool? Is it how Starknet works right now?

We introduced the pending state because blocks are really long on mainnet (could go up to 2 hours) so we needed a difference between tx in the mempool (received), tx processed but block not closed (pending) and tx included in a mined block (accepted on l2). As madara will have fast enough blocks we can ignore the pending state. The goal here is to return the receipt/information of mined txs. Once again let's completely ignore the pending state. Does that help ?

crates/client/db/src/mapping_db.rs
```diff
@@ -88,4 +88,41 @@ impl<B: BlockT> MappingDb<B> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `transaction_hash`: the transaction hash to be mapped. H256 is used here because it's a
+    ///   native type of substrate, and we are sure it's SCALE encoding is optimized and will not
+    ///   change.
+    ///
+    /// * `block_hash`: substrate block hash.
+    pub fn transaction_block_hashes_map(&self, transaction_hash: &H256, block_hash: &B::Hash) -> Result<(), String> {
+        let _lock = self.write_lock.lock();
```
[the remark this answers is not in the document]
glihm: Good point you mentioned.
It was solved with the refacto with everything in the save commitment function like frontier.

crates/client/mapping-sync/src/sync_blocks.rs
```diff
@@ -48,7 +50,9 @@ where
                             block_hash: substrate_block_hash,
                             starknet_block_hash: digest_starknet_block_hash.into(),
                         };
-                        backend.mapping().write_hashes(mapping_commitment)
+                        backend.mapping().write_hashes(mapping_commitment)?;
+
+                        commit_transactions_block_hashes_mapping(&substrate_block_hash, &digest_starknet_block, backend)
```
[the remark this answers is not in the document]
glihm: So much cleaner yes, thanks for pointing this out.

crates/client/rpc/src/lib.rs
glihm: I was limited by the fact that I didn't understand how to manage the possible error in the find. I have tried an implementation in the new commit with find.

crates/pallets/starknet/src/lib.rs
glihm: I have changed here to support the two fields `transactions` and `transaction_receipts` of the block. I have made one for loop to ensure we only iterate one time on `pending`. Once again, I didn't manage to get this with `iter` and `map`. Let me know in the new commit if it's ok. :+1:

crates/client/rpc/src/lib.rs
glihm: Refactored using find as you suggested.

glihm:
> > @LucasLvy @abdelhamidbakhta
> > > But if I understood correctly substrate, we have here a transaction status that is missing. Because saying it's pending in the transaction pool, would be incorrect as no receipt is available.
> > 
> > 
> > I don't understand, how could there be a receipt when the tx is still in the pool? Is it how Starknet works right now?
> 
> We introduced the pending state because blocks are really long on mainnet (could go up to 2 hours) so we needed a difference between tx in the mempool (received), tx processed but block not closed (pending) and tx included in a mined block (accepted on l2). As madara will have fast enough blocks we can ignore the pending state. The goal here is to return the receipt/information of mined txs. Once again let's completely ignore the pending state. Does that help ?

Helps a lot, thank you @LucasLvy.

Is this summary accurate then?

1. Tx in the mempool:
    * `getTransactionByHash` -> return the tx from the pool. (but no field to mentioned that it's "received").
    * `getTransactionReceipt` -> `TxnHashNotFound`? But how to say that's it's received? Or should I submit a PR to add a type like [PendingInvokeTransactionReceipt](https://github.com/xJonathanLEI/starknet-rs/blob/a6cbfa3be71e9f42532c393632bdf18ccb01f88a/starknet-core/src/types/codegen.rs#L757) to have one like `ReceivedTransactionReceipt`?
    
2. Tx in the block but block not finished -> skip for now as you said.

3. Tx in mined block -> Transaction + Receipt ok.

My only concern is about having the user to know that the transaction is received. Which, for now, I don't know how to do so.

Thanks again for all your feedbacks and comment guys, helped a lot.

crates/primitives/starknet/src/block/mod.rs
glihm: Last commit separates the two entities in the block. :+1:

LucasLvy:
1. For txs in the mempool you return an error `TxNotFound` or something like that
2. Yes also return an error of that kind
3. Yes exactly

glihm:
Waiting for #293 to check pending transactions.

crates/client/db/src/mapping_db.rs
```diff
@@ -84,8 +85,31 @@ impl<B: BlockT> MappingDb<B> {
...
 
         Ok(())
     }
+
+    /// Retrieves the substrate block hash
+    /// associated with the given transaction hash, if any.
+    ///
+    /// # Arguments
+    ///
+    /// * `transaction_hash`: the transaction hash to search for. H256 is used here because it's a
```
LucasLvy, line 106:
```suggestion
    /// * `transaction_hash` - the transaction hash to search for. H256 is used here because it's a
```

crates/client/rpc/src/lib.rs
```diff
@@ -731,4 +731,128 @@ where
...
+            .map_err(|e| {
+                error!("Failed to get transaction's substrate block hash from mapping_db: {e}");
+                StarknetRpcApiError::TxnHashNotFound
+            })?;
+
+        let substrate_block_hash = match block_hash_from_db {
+            Some(block_hash) => block_hash,
+            None => {
+                // Not found in the local mapping tx_hash <> block_hash, we
+                // then check into the transaction pool.
```
LucasLvy, line 759: don't check in the mempool, this is the job of `starknet_pendingTransaction`

LucasLvy (changes requested):

crates/client/db/src/mapping_db.rs
glihm: Good catch, thanks!

crates/client/rpc/src/lib.rs
glihm: Makes sense.. sorry not having this conclusion before.
Just pushed. :+1:

LucasLvy (approved):
lgtm

PR #352, opened 2023-05-04
Add RPC types from generated file

ClementWalter:
# Pull Request type

- Feature

Types for the Starknet RPC are iteratively added by the contributors.

# What is the new behavior?

The whole RPC spec has been generated using the @open-rpc/typings tools and contributors only need to make the implementation.

No

# Other information

I have generated the file once with the command added in the README, then made manual updates to have a comprehensive rust API spec that contrib can implement directly.

There are some open questions *to solve* before merge (type diff between the generated and the previous ones):

- `call`: `RpcResult<Vec<String>>` => `RpcResult<FieldElement>`
-  `get_block_transaction_count` : `u128` => `i64`
- `syncing` : removed `#[serde(rename = "sync_status")]`
- `BlockId` : removed the workaround with `BlockIdTaggedVariants`
- some `#[default]` removed


ClementWalter:
This won't work as easily, closing now to use starknet-rs work instead.
The [Starkware spec](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json) needs a significant rework imho

Issue #341, opened 2023-05-04
feat: RPC - starknet_addDeclareTransaction


abdelhamidbakhta:
Submit a new class declaration transaction

PR #380, opened 2023-05-11
feat: addDeclareTransaction RPC ✨

EvolveArt:

Resolves: #341 

## What is the new behavior?

Adds support for `Declare` Transaction V1 (Deprecated) one. 
Sierra Support will come in another PR.

No

### DIFF
--- a/crates/client/rpc-core/src/types/deprecated.rs
+++ b/crates/client/rpc-core/src/types/deprecated.rs
@@ -46,8 +46,14 @@ impl From<EntryPointWrapper> for DeprecatedCairoEntryPoint {
 
 impl From<DeprecatedCairoEntryPoint> for EntryPointWrapper {
     fn from(value: DeprecatedCairoEntryPoint) -> Self {
-        let selector = <[u8; 32]>::from_hex(&value.selector).unwrap();
-        let offset = value.offset.parse::<u128>().unwrap();
+        let selector = <[u8; 32]>::from_hex(format_hex(&value.selector)).unwrap();
+        let offset = u128::from_str_radix(&format_hex(&value.offset), 16).unwrap();
         Self { selector, offset }
     }
+}
+
+/// Removes the "0x" prefix from a given hexadecimal string and pads it with 0s
+#[inline(always)]
+fn format_hex(input: &str) -> String {
+    format!("{:0>64}", input.strip_prefix("0x").unwrap_or(input))
 }
--- a/crates/client/rpc-core/src/types/deprecated.rs
+++ b/crates/client/rpc-core/src/types/deprecated.rs
@@ -1,4 +1,4 @@
-use hex::ToHex;
+use hex::{FromHex, ToHex};
 use mp_starknet::execution::types::EntryPointWrapper;
 use serde::{Deserialize, Serialize};
 
@@ -40,6 +40,14 @@ impl From<EntryPointWrapper> for DeprecatedCairoEntryPoint {
         let selector = add_prefix(&selector);
         let offset: String = value.offset.to_be_bytes().as_slice().encode_hex();
         let offset = add_prefix(remove_leading_zeros(&offset));
+        Self { selector, offset }
+    }
+}
+
+impl From<DeprecatedCairoEntryPoint> for EntryPointWrapper {
+    fn from(value: DeprecatedCairoEntryPoint) -> Self {
+        let selector = <[u8; 32]>::from_hex(&value.selector).unwrap();
+        let offset = value.offset.parse::<u128>().unwrap();
         Self { selector, offset }
     }
 }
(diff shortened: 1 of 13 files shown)


crates/client/rpc-core/src/types/deprecated.rs
```diff
@@ -43,3 +43,17 @@ impl From<EntryPointWrapper> for DeprecatedCairoEntryPoint {
...
+impl From<DeprecatedCairoEntryPoint> for EntryPointWrapper {
+    fn from(value: DeprecatedCairoEntryPoint) -> Self {
+        let selector = <[u8; 32]>::from_hex(format_hex(&value.selector)).unwrap();
+        let offset = u128::from_str_radix(&format_hex(&value.offset), 16).unwrap();
+        Self { selector, offset }
+    }
+}
+
+/// Removes the "0x" prefix from a given hexadecimal string and pads it with 0s
+#[inline(always)]
```
abdelhamidbakhta, line 56: don't you think this function should live elsewhere ? in a utils primitives or something like that

crates/client/rpc-core/src/utils.rs
```diff
@@ -61,3 +61,38 @@ pub(crate) fn _to_deprecated_entrypoint_by_type(
...
     DeprecatedEntryPointsByType { constructor, external, l_1_handler }
 }
+
+/// Returns a hash map of entry point types to entrypoint from deprecated entry point by type
+pub fn to_btree_map_entrypoints(
+    entries: DeprecatedEntryPointsByType,
+) -> BTreeMap<EntryPointTypeWrapper, BoundedVec<EntryPointWrapper, MaxEntryPoints>> {
+    let mut entry_points_by_type: BTreeMap<EntryPointTypeWrapper, BoundedVec<EntryPointWrapper, MaxEntryPoints>> =
+        BTreeMap::new();
+    // We can unwrap safely as we already checked the length of the vectors
```
abdelhamidbakhta, line 71: where is this checked performed ? are you referring to a check outside of the function ? if yes, why do you assume the check will always be done, since the function is public ?

crates/client/rpc-core/src/utils.rs
```diff
@@ -61,3 +61,38 @@ pub(crate) fn _to_deprecated_entrypoint_by_type(
...
+        BoundedVec::try_from(
+            entries.l_1_handler.iter().map(|e| EntryPointWrapper::from(e.clone())).collect::<Vec<_>>(),
+        )
+        .unwrap(),
+    );
+    entry_points_by_type
+}
+
+/// Removes the "0x" prefix from a given hexadecimal string and pads it with 0s
+#[inline(always)]
```
abdelhamidbakhta, line 95: this function is defined twice

crates/client/rpc/src/lib.rs
```diff
@@ -477,10 +483,81 @@ where
...
 
-/// Removes the "0x" prefix from a given hexadecimal string and pads it with 0s
-#[inline(always)]
-fn format_hex(input: &str) -> String {
-    format!("{:0>64}", input.strip_prefix("0x").unwrap_or(input))
+    /// Declares a new contract class
+    ///
+    /// # Arguments
+    ///
+    /// * `declare_transaction` - The declare transaction
```
abdelhamidbakhta, line 491: missing block id param in rust doc

crates/client/rpc/src/lib.rs
```diff
@@ -477,10 +483,81 @@ where
...
+        // Program is send as gzip + base64 encoded, we need to decompress it
+        // Decode the base64 encoded string
+        let compressed_bytes = base64::decode(&tx.contract_class.program).unwrap();
+
+        // Create a GzipDecoder to decompress the bytes
+        let mut gz = GzDecoder::new(&compressed_bytes[..]);
+
+        // Read the decompressed bytes into a Vec<u8>
+        let mut decompressed_bytes = Vec::new();
+        gz.read_to_end(&mut decompressed_bytes).unwrap();
```
abdelhamidbakhta, line 526: could be worth extracting this in a function

crates/client/rpc/src/lib.rs
```diff
@@ -477,10 +483,81 @@ where
...
+
+        // Deserialize the decompressed bytes into a Rust struct using serde_json
+        let program: ProgramWrapper = serde_json::from_slice(&decompressed_bytes).unwrap();
+
+        // TODO: Add support for V2 Declare TXs
```
abdelhamidbakhta, line 531: please create issue and add reference to the issue

crates/client/rpc/src/lib.rs
```diff
@@ -477,10 +483,81 @@ where
...
+        // TODO: remove unwraps
```
abdelhamidbakhta, line 532: since this is a draft PR it's fine for now, but just want to make sure we don't forget it when you change the status of the PR, so I am adding a comment

abdelhamidbakhta (commented):
doing a first round, even if still in draft. Let me know if you prefer not to have reviews during the draft phase, and only a review when ready

Issue #364, opened 2023-05-05
Make the chain id configurable


LucasLvy:
# Description

The chain id is a constant for now, make it configurable.

# Acceptance criteria

* The chain id is configurable

chirag-bgh:
I would like to take this.

PR #383, opened 2023-05-13
feat: make chain_id configurable

chirag-bgh:
# Pull Request type

- Feature

Resolves: #364


crates/primitives/starknet/src/transaction/types.rs
```diff
@@ -22,6 +22,11 @@ use crate::execution::types::{CallEntryPointWrapper, ContractAddressWrapper, Con
 /// TODO: add real value (#250)
 pub type MaxArraySize = ConstU32<10000>;
 
+/// chain id for the oerli network.
+pub const SN_GOERLI: &str = "SN_GOERLI";
+/// chain id for the mainnet network.
+pub const SN_MAINNET: &str = "SN_MAINNET";
```
drspacemn, line 28: should be "SN_MAIN"

crates/primitives/starknet/src/transaction/types.rs
```diff
@@ -309,7 +314,7 @@ impl From<InvokeTransaction> for Transaction {
     fn from(value: InvokeTransaction) -> Self {
         Self {
             version: value.version,
-            hash: calculate_invoke_tx_hash(value.clone()),
+            hash: calculate_invoke_tx_hash(value.clone(), SN_GOERLI),
```
drspacemn, line 317: +335 should be configurable

crates/primitives/starknet/src/crypto/commitment/mod.rs
```diff
@@ -189,6 +192,7 @@ fn calculate_transaction_hash_common<T>(
     nonce: U256,
     version: u8,
     tx_prefix: &[u8],
+    chain_id: &str,
```
chirag-bgh, line 195: Since I guess we can't fetch the chain_id from the GenesisConfig here, I'd to add this as function parameter. But this would cause hardcoding the chain_id at some places like when casting `DeclareTransaction` to `Transaction`. 
https://github.com/keep-starknet-strange/madara/pull/383/files#diff-d1e7254b6a1f394f0f8f9fee94b5cfad347b50e492ef7a6d92e7238220852faeR340
How can I fix this?

LucasLvy:
Ok what I have in mind is: have an onchain storage for the chain id similar to that https://github.com/keep-starknet-strange/madara/blob/58fc213437664a1ead5c1ecabab94aca96e11504/crates/pallets/starknet/src/lib.rs#L249-L252 that would be set in the genesis config. Does that make sense ?

chirag-bgh:
> Ok what I have in mind is: have an onchain storage for the chain id similar to that
> 
> https://github.com/keep-starknet-strange/madara/blob/58fc213437664a1ead5c1ecabab94aca96e11504/crates/pallets/starknet/src/lib.rs#L249-L252
> 
> that would be set in the genesis config. Does that make sense ?

I think we already have this for GenesisConfig?

chirag-bgh:
Closing this due to too many conflicts. New PR at #494

LucasLvy:
> I think we already have this for GenesisConfig?
Yes but it's not used anywhere i think

Issue #359, opened 2023-05-05
dev: Use starknet-rs types for the Starknet RPC


ClementWalter:
The RPC (see [rpc-core](../tree/main/crates/client/rpc-core) and [rpc](../blob/main/crates/client/rpc) crates) defines custom types while the [starknet-rs repo](https://github.com/xJonathanLEI/starknet-rs) already did the job + generated them with a custom tool directly from the [RPC spec](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json).

Let us use them directly, as [Beerus](https://github.com/keep-starknet-strange/beerus/blob/a5fdd41dc47b989b4dee7f5aa8b0ae5d7907ba8d/crates/beerus-rpc/src/api.rs) does for instance.

ClementWalter:
ongoing work on this branch https://github.com/ClementWalter/madara/tree/cw/rpc-core/starknet-rs-types

on hold, waiting for starknet-rs to release wrapped types for request/response

Issue #381, opened 2023-05-11
bug: block hash out of range


xJonathanLEI:
# Bug Report

**Madara version:**

`349e7c1044eb9d715a6b06d259678ec9c4dd49fd`

**Current behavior:**

Block hashes (or maybe other hashes too; haven't tested) reported are full 256-bit hashes.

**Expected behavior:**

All hashes should be bounded by the prime to be valid field elements.

**Steps to reproduce:**

1. Run Maara with `--dev`:
   ```console
   cargo run -- --dev
   ```
2. Query the latest block hash:
   ```console
   curl "http://localhost:9933/" -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","id":1,"method":"starknet_blockHashAndNumber","params":[]}'
   ```
3. Check that the resulting block hash is 256-bit.

**Related code:**

N/A

**Other information:**

N/A

ClementWalter:
does this mean that hash return should be `hash % P` instead?

ClementWalter:
The issue is [here](../blob/main/crates/client/rpc/src/lib.rs#L137) if I'm right

```
        Ok(BlockHashAndNumber { block_hash: format!("{:#x}", block_hash), block_number })
```

tdelabro:
The problem is not RPC related. Our whole codebase uses U256 and H256 instead of Felts. We did it initially because it was more handy but is back firing now.

In order to fix this:
In `crates/primitives/starknet/src/block/header.rs`, we must change this function.
```
    #[must_use]
    pub fn hash(&self) -> H256 {
        H256::from_slice(frame_support::Hashable::blake2_256(&self.block_number.encode()).as_slice())
    }
```
Its signature should return a Felt and its implem should respect the starknet specification defined here: https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/header/#block_hash

Then we have to change everything that uses this type downstream: `BlockHeader` type definition, runtime storage, backend and rpc implem.

A problem we will have is that substrate expects some traits for the types it accepts in its storage. I think we will have to define our own `SubstrateFelt(Felt)` newtype.
I think such a type already exists with `Felt252Wrapper(pub U256);`  defined in `crates/primitives/starknet/src/execution/program_wrapper.rs`. I believe this is the type we should store in the runtime. But when using it in our program we should convert it to `cairo_vm::felt::Felt252`.

tdelabro:
[...]

PR #396, opened 2023-05-16
Use starknet-rs types in rpc

ClementWalter:
# Pull Request type

- Feature

The RPC crates define custom types while:
- we strive to have one single typing for the whole starknet X rust ecosystem
- it takes useless time to write them, especially when they are complex
- it's bug prone

Resolves: #359 

## What is the new behavior?

Use only types from starknet-rs.

Yes and No.
We need to do #381 to use these types in the runtime as well. In the current state, H256 cannot be converted to FieldElements, not sure about what's going on.


crates/client/rpc-core/Cargo.toml
```diff
@@ -28,3 +28,5 @@ sp-blockchain = { workspace = true, default-features = true }
 starknet_api = { workspace = true, default-features = false }
 frame-support = { workspace = true }
 sp-core = { workspace = true }
+starknet-providers = { git = "https://github.com/xJonathanLEI/starknet-rs", branch = "dev/jsonrpc_0_3_0", default-features = false }
```
abdelhamidbakhta, line 31: would be better to use `workspace = true` here to inherit the version defined in the workspace Cargo toml.

abdelhamidbakhta (dismissed):
left one very minor comment, otherwise looks good to me

ClementWalter:
Need to do https://github.com/keep-starknet-strange/madara/issues/381 ASAP to remove all the H256>FieldElement conversions

EvolveArt (approved):
lgtm

Issue #128, opened 2023-03-31
feat: block header hash computation 🦀


EvolveArt:
# Feature Request

**Describe the Feature Request**

The block hash should be computed correctly following the starknet specs defined [here](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/header/#block_hash).

You can find the code to change under `primitives/starknet/src/starknet_block/header.rs`

**Acceptance Criteria**

- Hash function used should be generic such that we can easily switch
- Tests should be added

dangush:
Hi there, I'd like to work on this!

abdelhamidbakhta:
> Hi there, I'd like to work on this!

let's go! I assigned it to you

dangush:
Hi there, just want to make sure I have the right approach. What is the case in which different hash functions would be used for producing block hashes? 

I see that there's a Hasher trait already implemented, should the header hash function accept a function implementing this trait?

EvolveArt:
> Hi there, just want to make sure I have the right approach. What is the case in which different hash functions would be used for producing block hashes? I'm noticing that the starknet pallet `SystemHash` is set to be `PedersenHasher`. Would this function be used for block headers too then?
> 
> Asking these questions to better understand how to make the hash function generic- should the header hash function signature be changed to include a function implementing trait Hasher as a parameter? Or should it be using whatever function is configured in the starknet pallet as the `SystemHash`?

Hello! Sorry for the late reply @dangush. Yes exactly the `SystemHash` should be used at least in a first step. In the future we might have two different hashers to comply with the current starknet specs.

And yes I believe you do have to change the function's signature to take a `Hasher` trait param too in order to make it work.

EvolveArt:
Hey @dangush still working on this ?

tdelabro:
@ClementWalter will take this issue in order to fix #381

ClementWalter:
I think that we also need to fix:

```
fn get_storage_at(address: ContractAddressWrapper, key: StorageKeyWrapper) -> Result<U256, DispatchError>;
fn call(address: ContractAddressWrapper, function_selector: H256, calldata: Vec<U256>) -> Result<Vec<StarkFeltWrapper>, DispatchError>;
```

to use `FieldElement` instead of `U256` or `H256`.

PR #400, opened 2023-05-18
feat: header block hash ✨

EvolveArt:

Resolves: #381 #128 

## What is the new behavior?

`Header.hash` is now generic on a hasher that implements the trait `Hasher` defined in the primitives.
This will make sure that block hashes are always in the range of a Field Element.

## Other information

In the next PR, we will change all runtime types from `H256/U256` to `Felt252Wrapper`


tests/util/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { number } from "starknet";
+import BN__default from "bn.js";
+
+export type BigNumberish = string | number | BN__default;
```
ClementWalter, line 4: isn't this type already defined in starknet-js?

EvolveArt: can't access it :(

abdelhamidbakhta (approved):
lgtm

Issue #285, opened 2023-04-21
bug: npm test-seq doesn't work within the docker instance


sparqet:
# Bug Report

**Current behavior:**

When building the project with `docker-compose -f infra/docker/docker-compose.yml up`, we are not able to run test with `npm run test-seq`

**Expected behavior:**

Should be able to run `npm run test-seq` after building madara with docker-compose

dbejarano820:
I'll look into this bug

PR #410, opened 2023-05-19
bug: docker instance not able to run npm scripts fix

dbejarano820:
The docker instance currently doesn't have the capability of running npm scripts since Node.js and npm aren't installed. This PR fixes that by adding a line to install them and removing a line which erases the bin directories.

# Pull Request type

- Bugfix

Docker instance isn't able to run npm scripts

Resolves: #285 

## What is the new behavior?
Added missing step to install Node.js and npm in the container, which are required to run `npm run test-seq`.

I'm 100% sure about this other change but I eliminated the lines which removes /usr/bin and /usr/sbin since it's affecting the ability to run Node.js and npm inside the container.

Another alternative could be creating a separate container to run the npm scripts, this way we could avoid removing the line. Let me know!

No

### DIFF
--- a/infra/docker/Dockerfile.dev
+++ b/infra/docker/Dockerfile.dev
@@ -1,0 +1,36 @@
+# This is a DEV base image to build substrate nodes
+FROM docker.io/paritytech/ci-linux:production as builder
+
+WORKDIR /madara
+COPY . .
+RUN cargo build --locked --release
+
+# This is the 2nd stage: a very small image where we copy the binary."
+FROM docker.io/library/ubuntu:20.04
+LABEL description="Multistage Docker image for Madara" \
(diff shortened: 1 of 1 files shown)


dbejarano820:
I will create a separate Dev Dockerfile instead of adding node dependencies to the production image

abdelhamidbakhta:
> I will create a separate Dev Dockerfile instead of adding node dependencies to the production image

Awesome, thanks

abdelhamidbakhta (changes requested):
Waiting for the separate docker file, as discussed.

abdelhamidbakhta (approved):
Lgtm

abdelhamidbakhta (approved):
Lgtm

PR #418, opened 2023-05-21
feat: refractoooor runtime types 🔥

EvolveArt:

Resolves: N/A

## What is the new behavior?

As discussed a lot lately, there is no need for cumbersome type conversions, both client-side and in the runtime.
We only need one type which here is `Felt252Wrapper` which can be used everywhere and that has `From`/`Into` implementations with main used types. 

This PR refractors the whole codebase to **ONLY** use this type, note that it can be slightly improved by adding more trait implementations to the type.

@glihm will also improve this PR by making `FieldElement` **SCALE** encodable and thus reducing again the overhead.

Yes


abdelhamidbakhta (approved):
Lgtm

PR #422, opened 2023-05-22
feat: Felt252Wrapper with SCALE encoding support

glihm:
<!--- Please provide a general summary of your changes in the title above -->

# Pull Request type

<!-- Please try to limit your pull request to one type; submit multiple pull requests if needed. -->

- Feature

At the moment, we are using `H256` and `U256` types to use Substrate SCALE encoding supported types.

## What is the new behavior?

We are now able to pass `FieldElement` from starknet-rs natively between client and runtime.

YES

## Other information

# Madara: quick overview of this type refactoring                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     

## SCALE encoding constraint
Madara is using substrate, which is using
[SCALE encoding](https://docs.substrate.io/reference/scale-codec/) for the
communication between the runtime and the client.

Native types from [sp_core](https://docs.rs/sp-core/latest/sp_core/) have
all those traits implemented.

Here are the traits (as there is few documentation about it) from
the
[`scale_codec`](https://github.com/paritytech/parity-scale-codec/tree/ad10bddc95ad5810489618f8c5c50cae506dece9/src)
repository and
[`scale_info`](https://github.com/paritytech/scale-info) repository where
I found the definition and some implementation examples.

```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
/// SCALE trait.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
impl Encode for Felt252Wrapper {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
    fn encode_to<T: Output + ?Sized>(&self, dest: &mut T) {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
        dest.write(/* write the type in a byte array */);                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
    }                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
/// SCALE trait. Required to enable `FullEncode`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
/// trait, which also support types like &Felt252Wrapper.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
impl EncodeLike for Felt252Wrapper {}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
/// SCALE trait.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
impl MaxEncodedLen for Felt252Wrapper {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
    /// Max size known in our case.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
    fn max_encoded_len() -> usize {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
        32                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
    }                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
/// SCALE trait.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
impl Decode for Felt252Wrapper {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
    fn decode<I: Input>(input: &mut I) -> Result<Self, Error> {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
            let mut buf: [u8; 32] = [0; 32];                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
            input.read(&mut buf)?;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
            // decode the type from byte array.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          
    }                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
/// SCALE trait.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
impl TypeInfo for Felt252Wrapper                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
{                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
    type Identity = Self;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
    // Define type info                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
    fn type_info() -> Type {                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
        Type::builder().path(...)                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
    }                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    
}                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      


## Starknet felt252
For Starknet, we have a `felt252` type which is know into several library
like `FieldElement` in
[starknet-rs](https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-ff/src/lib.rs#L26)
, `StarkFelt` in
[starknet-api](https://github.com/tdelabro/starknet-api/blob/4baf6db384c0e64c52eb97bd856c889e7d2cecaa/src/hash.rs#L50)
and finally `Felt252Wrapper` in
[Madara](https://github.com/keep-starknet-strange/madara/blob/62609f6808c882f8fe2301af2f4a14e56e06ebfb/crates/primitives/starknet/src/execution/program_wrapper.rs#L284)
.

The latter being the wrapper we want to adapt for SCALE encoding.
For now, it's using `U256` under the hood as this type is provided by
substrate and thus implements all SCALE encoding traits.

Madara is composing with all those implementations, each one in different
part of Madara.
**First question here**: Is this possible to homogenize the `felt252`
implementation or their use is to different?

Because we use starknet-rs under the hood for cryptography computation as
[commitments](https://github.com/keep-starknet-strange/madara/blob/62609f6808c882f8fe2301af2f4a14e56e06ebfb/crates/primitives/starknet/src/crypto/commitment/mod.rs#L206)
and hashes in the same module.
For starknet-api, it seems to be related to the `cairo-vm` stuff. But some
methods interesect starknet-rs (for computing class hash for instance).

But I don't have enough background to give more insights here.


## Types rework
This second rework, after the one done by @EvolveArt, is proposing
the use of `FieldElement` from starknet-rs as the value passing via
SCALE encoding between client and runtime.

However, as we also use `StarkFelt`, we still need at some places
to convert back to bytes array. Which is I think disminuing the possible
performance improvment from SCALE encoded `FieldElement`.

Anyway, for code refactoring, it's helping as convertions are easier
for every place that derives from `FieldElement`, which is at least
the runtime output to then go out for RPC responses.

### DIFF
--- a/crates/client/mapping-sync/src/sync_blocks.rs
+++ b/crates/client/mapping-sync/src/sync_blocks.rs
@@ -43,7 +43,7 @@ where
                         // Success, we write the Starknet to Substate hashes mapping to db
                         let mapping_commitment = mc_db::MappingCommitment {
                             block_hash: substrate_block_hash,
-                            starknet_block_hash: digest_starknet_block_hash,
+                            starknet_block_hash: digest_starknet_block_hash.into(),
                         };
                         backend.mapping().write_hashes(mapping_commitment)
                     }
@@ -68,7 +68,7 @@ where
     let block = client.runtime_api().current_block(substrate_block_hash).map_err(|e| format!("{:?}", e))?;
     let block_hash = block.header().hash(PedersenHasher::default());
     let mapping_commitment =
-        mc_db::MappingCommitment::<B> { block_hash: substrate_block_hash, starknet_block_hash: block_hash };
+        mc_db::MappingCommitment::<B> { block_hash: substrate_block_hash, starknet_block_hash: block_hash.into() };
     backend.mapping().write_hashes(mapping_commitment)?;
 
     Ok(())
--- a/crates/client/rpc-core/src/utils.rs
+++ b/crates/client/rpc-core/src/utils.rs
@@ -56,7 +56,9 @@ pub fn to_invoke_tx(tx: BroadcastedInvokeTransaction) -> Result<InvokeTransactio
                 invoke_tx_v1.signature.iter().map(|x| (*x).into()).collect::<Vec<Felt252Wrapper>>(),
             )
             .map_err(|e| anyhow!("failed to convert signature: {:?}", e))?,
-            sender_address: invoke_tx_v1.sender_address.to_bytes_be().into(),
+
+            // Safe to unwrap, starknet-core already parsed the FieldElement with jsonrpsee.
+            sender_address: Felt252Wrapper::try_from(&invoke_tx_v1.sender_address.to_bytes_be()).unwrap(),
             nonce: U256::from(invoke_tx_v1.nonce.to_bytes_be()),
             calldata: BoundedVec::try_from(
                 invoke_tx_v1.calldata.iter().map(|x| (*x).into()).collect::<Vec<Felt252Wrapper>>(),
@@ -87,17 +89,17 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         .try_into()
         .map_err(|_| anyhow!("failed to bound signatures Vec<H256> by MaxArraySize"))?;
 
-    let sender_address = calculate_contract_address(
-        ContractAddressSalt(StarkFelt(contract_address_salt)),
-        ClassHash(StarkFelt(account_class_hash)),
-        &Calldata(calldata.into()),
-        StarknetContractAddress::default(),
-    )
-    .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
-    .0
-    .0
-    .0
-    .into();
+    let sender_address = Felt252Wrapper::try_from(
+        &calculate_contract_address(
+            ContractAddressSalt(StarkFelt(contract_address_salt)),
+            ClassHash(StarkFelt(account_class_hash)),
+            &Calldata(calldata.into()),
+            StarknetContractAddress::default(),
+        )
+            .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
+            .0
+            .0
+            .0).unwrap(); // Ok to unwrap, starknet-core parsed type.
 
     let calldata = tx
         .constructor_calldata
@@ -116,7 +118,7 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         calldata,
         salt: U256::from(contract_address_salt),
         signature,
-        account_class_hash: account_class_hash.into(),
+        account_class_hash: Felt252Wrapper::try_from(&account_class_hash).unwrap(), // Ok to unwrap, starknet-core parsed type.
         nonce,
         max_fee,
     })
--- a/crates/client/rpc-core/src/utils.rs
+++ b/crates/client/rpc-core/src/utils.rs
@@ -58,7 +58,7 @@ pub fn to_invoke_tx(tx: BroadcastedInvokeTransaction) -> Result<InvokeTransactio
             .map_err(|e| anyhow!("failed to convert signature: {:?}", e))?,
 
             // Safe to unwrap, starknet-core already parsed the FieldElement with jsonrpsee.
-            sender_address: Felt252Wrapper::try_from(&invoke_tx_v1.sender_address.to_bytes_be()).unwrap(),
+            sender_address: Felt252Wrapper::from(invoke_tx_v1.sender_address),
             nonce: U256::from(invoke_tx_v1.nonce.to_bytes_be()),
             calldata: BoundedVec::try_from(
                 invoke_tx_v1.calldata.iter().map(|x| (*x).into()).collect::<Vec<Felt252Wrapper>>(),
@@ -76,7 +76,7 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
 
     let contract_address_salt = tx.contract_address_salt.to_bytes_be();
 
-    let account_class_hash = tx.class_hash.to_bytes_be();
+    let account_class_hash = tx.class_hash;
 
     let calldata =
         tx.constructor_calldata.iter().filter_map(|f| StarkFelt::new(f.to_bytes_be()).ok()).collect::<Vec<_>>();
@@ -89,19 +89,16 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         .try_into()
         .map_err(|_| anyhow!("failed to bound signatures Vec<H256> by MaxArraySize"))?;
 
-    let sender_address = Felt252Wrapper::try_from(
-        &calculate_contract_address(
-            ContractAddressSalt(StarkFelt(contract_address_salt)),
-            ClassHash(StarkFelt(account_class_hash)),
-            &Calldata(calldata.into()),
-            StarknetContractAddress::default(),
-        )
+    let sender_address = calculate_contract_address(
+        ContractAddressSalt(StarkFelt(contract_address_salt)),
+        ClassHash(StarkFelt(account_class_hash.to_bytes_be())),
+        &Calldata(calldata.into()),
+        StarknetContractAddress::default(),
+    )
         .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
         .0
         .0
-        .0,
-    )
-    .unwrap(); // Ok to unwrap, starknet-core parsed type.
+        .into();
 
     let calldata = tx
         .constructor_calldata
@@ -120,8 +117,7 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         calldata,
         salt: U256::from(contract_address_salt),
         signature,
-        account_class_hash: Felt252Wrapper::try_from(&account_class_hash).unwrap(), /* Ok to unwrap, starknet-core
-                                                                                     * parsed type. */
+        account_class_hash: account_class_hash.into(),
         nonce,
         max_fee,
     })
--- a/crates/client/rpc-core/src/utils.rs
+++ b/crates/client/rpc-core/src/utils.rs
@@ -95,10 +95,10 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         &Calldata(calldata.into()),
         StarknetContractAddress::default(),
     )
-        .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
-        .0
-        .0
-        .into();
+    .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
+    .0
+    .0
+    .into();
 
     let calldata = tx
         .constructor_calldata
--- a/crates/client/rpc-core/src/utils.rs
+++ b/crates/client/rpc-core/src/utils.rs
@@ -96,10 +96,12 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
             &Calldata(calldata.into()),
             StarknetContractAddress::default(),
         )
-            .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
-            .0
-            .0
-            .0).unwrap(); // Ok to unwrap, starknet-core parsed type.
+        .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
+        .0
+        .0
+        .0,
+    )
+    .unwrap(); // Ok to unwrap, starknet-core parsed type.
 
     let calldata = tx
         .constructor_calldata
@@ -118,7 +120,8 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         calldata,
         salt: U256::from(contract_address_salt),
         signature,
-        account_class_hash: Felt252Wrapper::try_from(&account_class_hash).unwrap(), // Ok to unwrap, starknet-core parsed type.
+        account_class_hash: Felt252Wrapper::try_from(&account_class_hash).unwrap(), /* Ok to unwrap, starknet-core
+                                                                                     * parsed type. */
         nonce,
         max_fee,
     })
--- a/crates/client/rpc-core/src/utils.rs
+++ b/crates/client/rpc-core/src/utils.rs
@@ -57,8 +57,7 @@ pub fn to_invoke_tx(tx: BroadcastedInvokeTransaction) -> Result<InvokeTransactio
             )
             .map_err(|e| anyhow!("failed to convert signature: {:?}", e))?,
 
-            // Safe to unwrap, starknet-core already parsed the FieldElement with jsonrpsee.
-            sender_address: Felt252Wrapper::from(invoke_tx_v1.sender_address),
+            sender_address: invoke_tx_v1.sender_address.into(),
             nonce: U256::from(invoke_tx_v1.nonce.to_bytes_be()),
             calldata: BoundedVec::try_from(
                 invoke_tx_v1.calldata.iter().map(|x| (*x).into()).collect::<Vec<Felt252Wrapper>>(),
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -378,7 +378,7 @@ where
             .unwrap_or_default();
 
         let transactions = block.transactions_hashes().into_iter().map(FieldElement::from).collect();
-        let blockhash = Felt252Wrapper::try_from(block.header().hash(PedersenHasher::default())).unwrap();
+        let blockhash = block.header().hash(PedersenHasher::default());
         let parent_blockhash = block.header().parent_block_hash;
         let block_with_tx_hashes = BlockWithTxHashes {
             transactions,
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -25,7 +25,7 @@ use sc_transaction_pool_api::{TransactionPool, TransactionSource};
 use sp_api::{ApiError, ProvideRuntimeApi};
 use sp_arithmetic::traits::UniqueSaturatedInto;
 use sp_blockchain::HeaderBackend;
-use sp_core::{H256, U256};
+use sp_core::{H256};
 use sp_runtime::generic::BlockId;
 use sp_runtime::traits::{Block as BlockT, Header as HeaderT};
 use starknet_core::types::FieldElement;
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -96,7 +96,7 @@ where
             .current_block(substrate_block_hash)
             .unwrap_or_default();
 
-        Ok(block.header().hash(PedersenHasher::default()))
+        Ok(block.header().hash(PedersenHasher::default()).into())
     }
 
     /// Returns the substrate block corresponding to the given Starknet block id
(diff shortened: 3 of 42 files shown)


glihm:
I pushed the PR to give @EvolveArt an overview of the refacto.

Except linters, I have an error related to the encoding. The `sender_address` on invoke transaction is `0x8000000000000000000000000000000000000000000000000000000000000000`, which is out of range.

It's good we can't pass out of range value, now I have to find why.

glihm:
To add some context:

When we want a felt from an array of bytes or U256, we always have to ensure that there is no overflow -> `Felt252Wrapper::try_from(...)`.

The same applies to the two methods from strings (hex and decimal), where we have to ensure
no invalid characters are passed, or out of range again.
`Felt252Wrapper::from_hex_be("0x1")`  
`Felt252Wrapper::from_dec_str("1234")`

Where overflow can't happen, if when we convert from `u128` or an other felt implementation.
`Felt252Wrapper::from(1991_u128)`  
`Felt252Wrapper::from(FieldElement::ZERO)`

EvolveArt (approved):
lgtm

crates/client/mapping-sync/src/sync_blocks.rs
```diff
@@ -68,7 +68,7 @@ where
     let block = client.runtime_api().current_block(substrate_block_hash).map_err(|e| format!("{:?}", e))?;
     let block_hash = block.header().hash(PedersenHasher::default());
     let mapping_commitment =
-        mc_db::MappingCommitment::<B> { block_hash: substrate_block_hash, starknet_block_hash: block_hash };
+        mc_db::MappingCommitment::<B> { block_hash: substrate_block_hash, starknet_block_hash: block_hash.into() };
```
ClementWalter, line 71: not related to _this_ change, but are we sure that we want to define the variables such that we have

```
{
    block_hash: substrate_block_hash,
    starknet_block_hash: block_hash.into()
}
```

and not 

```
{
    block_hash,
    starknet_block_hash: starknet_block_hash.into()
}
```

crates/client/rpc-core/src/utils.rs
```diff
@@ -56,7 +56,9 @@ pub fn to_invoke_tx(tx: BroadcastedInvokeTransaction) -> Result<InvokeTransactio
                 invoke_tx_v1.signature.iter().map(|x| (*x).into()).collect::<Vec<Felt252Wrapper>>(),
             )
             .map_err(|e| anyhow!("failed to convert signature: {:?}", e))?,
-            sender_address: invoke_tx_v1.sender_address.to_bytes_be().into(),
+
+            // Safe to unwrap, starknet-core already parsed the FieldElement with jsonrpsee.
+            sender_address: Felt252Wrapper::try_from(&invoke_tx_v1.sender_address.to_bytes_be()).unwrap(),
```
ClementWalter, line 61: there is a `impl From<FieldElement> for Felt252Wrapper` so why not just

```
sender_address: invoke_tx_v1.sender_address.into()
```

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,373 @@
...
+use starknet_ff::{FieldElement, FromByteSliceError, FromStrError};
+
+///
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct Felt252Wrapper(pub FieldElement);
+
+impl Felt252Wrapper {
+    /// Inits from '0x...' hex string.
+    pub fn from_hex_be(value: &str) -> Result<Self, Felt252WrapperError> {
```
ClementWalter, line 26: can't we use `DerefMut` ? https://doc.rust-lang.org/stable/std/ops/trait.DerefMut.html

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,373 @@
...
+use scale_codec::{Decode, Encode, EncodeLike, Error, Input, MaxEncodedLen, Output};
+use scale_info::build::Fields;
+use scale_info::{Path, Type, TypeInfo};
+use sp_core::{H256, U256};
+use starknet_ff::{FieldElement, FromByteSliceError, FromStrError};
+
+///
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct Felt252Wrapper(pub FieldElement);
```
ClementWalter, line 22: can't we use [`DerefMut`](https://doc.rust-lang.org/stable/std/ops/trait.DerefMut.html) to avoid all the re-implementations below?

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,373 @@
...
+    }
+
+    /// Inits from decimal string.
+    pub fn from_dec_str(value: &str) -> Result<Self, Felt252WrapperError> {
+        let ff = FieldElement::from_dec_str(value)?;
+        Ok(Self(ff))
+    }
+
+    /// Inits from zero constant value.
+    pub fn zero() -> Felt252Wrapper {
```
ClementWalter, line 38: I see that you use `Deref` below, so the only purposed of these is to change the case and type? (noob question)

crates/client/rpc-core/src/utils.rs
glihm: @ClementWalter it's a very nice catch and in the overall refacto I'll double check I didn't miss something like this.
As before it was `H256/U256` the conversion to bytes was needed. Now it's not, so it's a useless call to `to_bytes_be`, which can then be refact as you said. 👍

EvolveArt: > there is a `impl From<FieldElement> for Felt252Wrapper` so why not just
> 
> ```
> sender_address: invoke_tx_v1.sender_address.into()
> ```

I find this way, more cumbersome for sure but clearer aswell. Note that in this case it would actually be `.try_into().unwrap`

crates/client/mapping-sync/src/sync_blocks.rs
EvolveArt: Yes, cf [frontier](https://github.com/paritytech/frontier/blob/master/client/mapping-sync/src/lib.rs#L61)

crates/primitives/starknet/src/execution/felt252_wrapper.rs
EvolveArt: Tried to remove it but couldn't access inner type public constants, not sure why.

crates/primitives/starknet/src/execution/felt252_wrapper.rs
EvolveArt: Tried and doesn't work unfortunately, maybe I'm doing it wrong but seems that we still need these re-implementations.

```rust
impl DerefMut for Felt252Wrapper {
    fn deref_mut(&mut self) -> &mut Self::Target {
        return &mut self.0;
    }
}
```

crates/client/rpc-core/src/utils.rs
ClementWalter: I think that we should favor everywhere `(try_)into()` instead of `Type::(Try)From` because the latter is redundant and heavier to maintain. For example, it's possible that we later change the function signature and update if necessary the input type with a new `From`, but we won't have to rewrite everything. It seems to me that `into` is more versatile and more idiomatic. The compiler knows what to do, we don't need to write it as well

crates/primitives/starknet/src/execution/felt252_wrapper.rs
ClementWalter: @tdelabro an idea?

crates/primitives/starknet/src/execution/felt252_wrapper.rs
ClementWalter: @tdelabro an idea?

crates/client/rpc-core/src/utils.rs
```diff
@@ -87,17 +89,19 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
...
+        &calculate_contract_address(
+            ContractAddressSalt(StarkFelt(contract_address_salt)),
+            ClassHash(StarkFelt(account_class_hash)),
+            &Calldata(calldata.into()),
+            StarknetContractAddress::default(),
+        )
+        .map_err(|e| anyhow!("Failed to calculate contract address: {e}"))?
+        .0
+        .0
+        .0,
```
ClementWalter, line 102: current types are 
image
so we could stop at `StarkFelt` and `.into`

crates/client/rpc-core/src/utils.rs
```diff
@@ -116,7 +120,8 @@ pub fn to_deploy_account_tx(tx: BroadcastedDeployAccountTransaction) -> Result<D
         calldata,
         salt: U256::from(contract_address_salt),
         signature,
-        account_class_hash: account_class_hash.into(),
+        account_class_hash: Felt252Wrapper::try_from(&account_class_hash).unwrap(), /* Ok to unwrap, starknet-core
```
ClementWalter, line 123: if you look on top of the file, you'll see that

```
let account_class_hash = tx.class_hash.to_bytes_be();
```

so you may just remove the `.to_bytes_be()`, keep the `FieldElement` and `.into()`

crates/client/mapping-sync/src/sync_blocks.rs
EvolveArt: To have what ? Well here it's quite important we don't differ much from an audited and proven secure codebase that frontier is.

EvolveArt: @tdelabro has worked the most on this part and could help clarify there.

crates/client/rpc/src/lib.rs
```diff
@@ -179,11 +179,11 @@ where
         })?;
 
         let runtime_api = self.client.runtime_api();
-        let hex_address = contract_address.to_bytes_be();
-        let hex_key = H256::from_slice(&key.to_bytes_be()[..32]);
+        let hex_address = Felt252Wrapper::from(contract_address);
```
ClementWalter, line 182: similar to above, the `contract_address` is a `FieldElement` so you could remove this line and keep the `.into` in the `get_storage_at` call

crates/primitives/starknet/src/execution/felt252_wrapper.rs
tdelabro: https://doc.rust-lang.org/std/ops/trait.Deref.html

```
Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and [DerefMut](https://doc.rust-lang.org/std/ops/trait.DerefMut.html) were designed specifically to accommodate smart pointers. Because of this, Deref should only be implemented for smart pointers to avoid confusion.
```

We should not use deref just to access the inner type of a wrapper

tdelabro (changes requested):
Remove the deref implem for `Felt252Wrapper`

https://github.com/keep-starknet-strange/madara/pull/422/files/c2d4ca2beb31d55b0f8f9dcea08577fdf812455c#r1202064665

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,373 @@
...
+            Ok(ff) => Ok(Self(ff)),
+            Err(e) => match e {
+                FromByteSliceError::InvalidLength => Err(Felt252WrapperError::InvalidLength),
+                FromByteSliceError::OutOfRange => Err(Felt252WrapperError::OutOfRange),
+            },
+        }
+    }
+}
+
+/// u128 converter. Might be supported natively?
```
tdelabro: by `natively`, do you mean using the language keyword `as`?

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,373 @@
...
+            Felt252WrapperError::FromArrayError => Cow::Borrowed("input array invalid"),
+            Felt252WrapperError::InvalidCharacter => Cow::Borrowed("invalid character"),
+            Felt252WrapperError::OutOfRange => Cow::Borrowed("number out of range"),
+            Felt252WrapperError::InvalidLength => Cow::Borrowed("invalid length"),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl From<Felt252WrapperError> for std::string::String {
```
tdelabro, line 304: I think this could be implemented in `no_std` too.
It just requires the feature using `alloc::string::String` rather than the `std` path.
If you really don't wish to make it available by default, put it behind `#[cfg(any(feature = "std", feature = "alloc"))]`

tdelabro (changes requested):

crates/client/mapping-sync/src/sync_blocks.rs
ClementWalter: my comment was just about naming though

crates/client/rpc-core/src/utils.rs
```diff
@@ -56,7 +56,9 @@ pub fn to_invoke_tx(tx: BroadcastedInvokeTransaction) -> Result<InvokeTransactio
                 invoke_tx_v1.signature.iter().map(|x| (*x).into()).collect::<Vec<Felt252Wrapper>>(),
             )
             .map_err(|e| anyhow!("failed to convert signature: {:?}", e))?,
-            sender_address: invoke_tx_v1.sender_address.to_bytes_be().into(),
+
+            // Safe to unwrap, starknet-core already parsed the FieldElement with jsonrpsee.
```
ClementWalter, line 60: no more unwrap

crates/primitives/starknet/src/execution/felt252_wrapper.rs
glihm: It was more ignorance of my side. I read a bit about `no_std` and `alloc`, your comment is definitely the way to go. Thanks again @tdelabro.

glihm:
@tdelabro @EvolveArt @ClementWalter I have tried to make a new revision with all your comments. Let me know if there is something missing.

I am still trying to understand why the tests are failing. It looks like the extrinsic is well submitted, but not added to the block. Is there something to force the execution of transaction in the mempool that is not happening?

I am still discovering TS testing with mocha and polkadot.js, I'll continue to hard try on that to understand how to fix it. If someone has ideas or knowledge, very welcome!

Sorry for not being to solve this more quickly, I'll do my best.

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,404 @@
...
+/// Array of bytes from [`Felt252Wrapper`].
+impl From<Felt252Wrapper> for [u8; 32] {
+    fn from(felt: Felt252Wrapper) -> Self {
+        felt.0.to_bytes_be()
+    }
+}
+
+/// [`Felt252Wrapper`] from bytes.
+/// Overflow may occur and return [`Felt252WrapperError::OutOfRange`].
+impl TryFrom<&[u8; 32]> for Felt252Wrapper {
```
tdelabro, line 94: I would have expected `TryFrom<[u8; 32]>` to be implemented, rather than. `TryFrom<&[u8; 32]>`.

tdelabro: Maybe both can exist

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -0,0 +1,404 @@
...
+        StarkFelt::new(buf).unwrap()
+    }
+}
+
+/// SCALE trait.
+impl Encode for Felt252Wrapper {
+    fn encode_to<T: Output + ?Sized>(&self, dest: &mut T) {
+        dest.write(&self.0.to_bytes_be());
+    }
+}
```
tdelabro, line 215: https://docs.substrate.io/reference/scale-codec/

I'm not sure at all that this is a good implementation.
`to_bytes_be` will return [u8; 32], 256 bits, even if the value is 0. With scale codec zero should usually be encoded as `0x00`. 
The whole purpose of scale codec is to reduce the size of the data being stored/send

tdelabro (changes requested):

crates/primitives/starknet/src/execution/felt252_wrapper.rs
tdelabro: Also, scale codec is little-endian. When `to_bytes_be` is big-endian.
Scale is for: Simple Concatenated Aggregate Little-Endian.

Issue #412, opened 2023-05-19
infra: Create a docker compose for the Starknet stack.


abdelhamidbakhta:
## Description:

The goal of this task is to prepare a Docker Compose configuration to run an instance of Madara (the latest image pushed from the release) and the Madara App (a fork of PolkadotJS) as an alternative to a proper block explorer.

The Madara App should take the URL of the Madara WebSocket (WS) as an environment variable. If this functionality isn't yet implemented in the app, it might require adding this feature.

This configuration should live in the `infra/starknet-stack` directory. The intention is to expand this stack later with additional Starknet components like a data indexer (Checkpoint / Apibara), a block explorer, and a faucet (Starkcet).

## Additional Information:

- The Dockerfile used to automatically push Docker images on each release (through a Github action workflow) is located here: [Dockerfile](https://github.com/keep-starknet-strange/madara/blob/main/infra/docker/Dockerfile)
- For reference, the latest published image is `ghcr.io/keep-starknet-strange/madara:v0.1.0-alpha`. The Docker Compose configuration can use the `latest` tag for the Docker image.

## Tasks:

1. Create a Docker Compose configuration to run a Madara instance and the Madara App.
2. Ensure the Madara App can take the Madara WS URL as an environment variable. If not, add this feature.
3. Place the configuration in the `infra/starknet-stack` directory.
4. Test the Docker Compose setup to ensure both the Madara instance and Madara App are working as expected.

## Acceptance Criteria:

- A working Docker Compose configuration that starts a Madara instance and the Madara App.
- The Madara App should be able to accept the Madara WS URL as an environment variable.

If you need any clarification or run into any issues, please comment on this issue.

robinstraub:
Can you assign this issue to me ?

abdelhamidbakhta:
> Can you assign this issue to me ?

Sure thing, done. Thanks

PR #430, opened 2023-05-22
build: add a docker compose for the starknet stack

robinstraub:
# Pull Request type

- Build-related changes
- Other (please describe): infra, the closest conventional-commit tag is build IMO

Every piece of the starknet stack (madara, madara-app, etc...) needs to be run on its own.

Resolves: #412 

## What is the new behavior?

A new docker-compose.yml allows to start the services composing the starknet stack. This file serves as a foundation for adding later services. Current services are

- madara
- madara-app

No

## Other information

This configuration targets local setup and tries to maximize ease of use. As such : 
- madara is opened as much as possible (prometheus, rpc and websocket are exposed on the host)
- madara-app is exposed on port 8080 (to avoid conflicts with local HTTP servers)

### DIFF
--- a/infra/starknet-stack/docker-compose.yaml
+++ b/infra/starknet-stack/docker-compose.yaml
@@ -1,0 +1,19 @@
+version: "3.2"
+
+services:
+  madara:
+    image: ghcr.io/keep-starknet-strange/madara:latest
+    ports:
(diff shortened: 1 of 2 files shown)


abdelhamidbakhta (approved):
lgtm

PR #441, opened 2023-05-23
fix: build error from starknet-rs orphan commit

xJonathanLEI:
The branch `dev/jsonrpc_0_3_0` in `starknet-rs` has bee rebased and forced pushed, causing the previously pinned commit to be orphaned. New contributors who do not have the commit checked out in their Cargo cache would have trouble building the project. While the orphan commit has been made available again via a Git tag on the upstream, this project specifically locks to the `dev/jsonrpc_0_3_0` branch, causing dependency resolution to fail (as the commit is _not_ on that branch).

The solution here is to simply lock by commit, which is the best practice anyways.

# Pull Request type

- Build-related changes

Project cannot be built from a fresh Cargo installation (cache pruned).

## What is the new behavior?

It builds.

No

## Other information

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4493,7 +4493,7 @@ dependencies = [
  "sp-core 7.0.0 (git+https://github.com/paritytech/substrate?branch=polkadot-v0.9.42)",
  "sp-runtime",
  "starknet-core",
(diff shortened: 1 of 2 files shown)


abdelhamidbakhta (approved):
lgtm

PR #446, opened 2023-05-24
Add benchmark in ci

LucasLvy:
# Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-


crates/pallets/starknet/src/lib.rs
```diff
@@ -731,25 +731,25 @@ pub mod pallet {
 
             match call {
                 Call::invoke { transaction } => ValidTransaction::with_tag_prefix("starknet")
-                    .priority(T::UnsignedPriority::get())
+                    .priority(u64::MAX - transaction.nonce.as_u64())
```
abdelhamidbakhta, line 734: comment the rationale behind this. it's not trivial to understand the reasoning and the strategy here. please explain it very clearly. eventually add examples

scripts/bench.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
abdelhamidbakhta, line 1: I think the name of the script is not explicit. It's goal is to compute the difference between two benchmarks, not to run a benchmark, hence I think the name is confusing

benchmarking/scripts/metrics.js
```diff
@@ -11,14 +11,14 @@ async function main() {
   const blockNumber = blockHash.block.header.number;
 
   let totalExtrinsics = 0;
-
-  for (let i = blockNumber.toNumber() - 9; i <= blockNumber.toNumber(); i++) {
+  setTimeout(() => {}, 10000);
```
abdelhamidbakhta, line 14: explain the testing strategy in a human readable form, with a block of comments before the test that explains the global process

abdelhamidbakhta (changes requested):
left comments

abdelhamidbakhta (approved):
lgtm

PR #447, opened 2023-05-24
[RPC] pendingTransactions

ClementWalter:
# Pull Request type

- Feature

Resolves: #204 

No

## Other information

Rebase from https://github.com/keep-starknet-strange/madara/pull/293 ; I cannot push to the original branch on finiam repo


glihm:
Could you please check that the pending transaction vector is also empty until the moment of finalizing the block?

It does not block this PR, just that it's a behavior that will cause this endpoint to almost always return an empty vector.

PR #458, opened 2023-05-25
Import starknet ff serde in std

tdelabro:
# Pull Request type

- Bugfix

- in primitives, `starknet-ff` was imported without the `serde` feature. Making it impossible to compile with `std` feature because the serde derivation on `Felt252Wrapper` failed.
- `one`, `two` and `tree` were functions

## What is the new behavior?

- `starknet-ff` is now imported with `serde` feature` when building `primitive` with `std` feature
- `ONE` `TWO` `THREE` are now constants and `MAX` is introduced

no

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -787,7 +787,7 @@ impl<T: Config> Pallet<T> {
     #[inline(always)]
     pub fn parent_block_hash(current_block_number: &U256) -> Felt252Wrapper {
         if current_block_number == &U256::zero() {
-            Felt252Wrapper::zero()
+            Felt252Wrapper::ZERO
         } else {
             Self::block_hash(current_block_number - 1)
         }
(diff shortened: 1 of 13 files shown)


crates/primitives/starknet/Cargo.toml
```diff
@@ -25,7 +25,9 @@ starknet-ff = { workspace = true, default-features = false, features = [
 	"alloc",
 ] }
 poseidon_hash = { workspace = true, default-features = false }
-blockifier = { workspace = true, default-features = false, features = ["testing"]}
+blockifier = { workspace = true, default-features = false, features = [
+	"testing",
```
ClementWalter, line 29: it looks like we don't all have the same formatter, not sure how to standardize this though, may investigate

LucasLvy (approved):
lgtm

ClementWalter (approved):

crates/primitives/starknet/Cargo.toml
tdelabro: Better TOML
Id: bungcip.better-toml
Description: Better TOML Language support
Version: 0.3.2
Publisher: bungcip
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=bungcip.better-toml

tdelabro: I have this VS code extension enabled. It's probably the one messing with TOMLs

PR #474, opened 2023-05-27
feat: add `Felt252Wrapper` for `nonce` and `max_fees`

bajpai244:
# Pull Request type

- Feature

The current transaction types use `U256` for `nonce` and `max_fees`.

Resolves: #456 

## What is the new behavior?

This PR makes the type of `nonce` and `max_fees` -> `Felt252Wrapper`.

### DIFF
--- a/tests/util/starknet.ts
+++ b/tests/util/starknet.ts
@@ -2,8 +2,9 @@ import "@keep-starknet-strange/madara-api-augment";
 import { type ApiPromise } from "@polkadot/api";
 import { type ApiTypes, type SubmittableExtrinsic } from "@polkadot/api/types";
 import { type ISubmittableResult } from "@polkadot/types/types";
-import { stringify, u8aWrapBytes } from "@polkadot/util";
+import { stringify, u8aWrapBytes, numberToU8a } from "@polkadot/util";
 import erc20Json from "../contracts/compiled/erc20.json";
+import { numberToU832Bytes } from "./utils";
 export async function sendTransactionNoValidation(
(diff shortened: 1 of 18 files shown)


tests/util/starknet.ts
```diff
@@ -248,7 +249,7 @@ export function transfer(
     version: 1, // version of the transaction
     signature: [], // leave empty for now, will be filled in when signing the transaction
     sender_address: contractAddress, // address of the sender contract
-    nonce: nonce || 0, // nonce of the transaction
+    nonce: numberToU832Bytes(nonce ? nonce : 0), // nonce of the transaction
```
bajpai244, line 252: If you are wondering about this function, when running the typescript tests for the extrinsic, after changing nonce to Felt252Wrapper, it was expecting it to be in a 32 bytes format, hence this utility was added which converts the nonce to a 32 bytes uint format.

tests/util/utils.ts
```diff
@@ -24,6 +25,11 @@ export function toBN(value: BigNumberish) {
   return number.toBN(value);
 }
 
+// Convert a BigNumberish to a 32 byte uint array
+export function numberToU832Bytes(value: number) {
```
bajpai244, line 29: This [comment](https://github.com/keep-starknet-strange/madara/pull/474/files#r1207781425) explains why this utility was added.

abdelhamidbakhta (approved):
Lgtm

bajpai244:
Lgtm ser 🚀

Issue #463, opened 2023-05-26
dev: unify all error enum declaration using thiserror_no_std in runtime context and this_error in client


Eikix:
To declare our error system, we use the thiserror crate. It has a no_std version for the runtime.

Some errors are declared as simple enums. We need to convert these to thiserror_no_std / this_error depending on the context.

1. Find errors declared as simple enums.
2. Find the proper way to convert them to this_error enums, respecting the structure and intent of the codebase.
3. When you come accross a pallet error declaration, ignore it. It is a substrate specific error declaration (dispatch error), it should not be changed.
4. Convert all "raw" enum errors to this_error enum errors

Examples of how it should be ✅:
https://github.com/keep-starknet-strange/madara/blob/main/crates/primitives/starknet/src/execution/contract_class_wrapper.rs#L61

Examples of how it should not be ❌:
https://github.com/keep-starknet-strange/madara/blob/ae4771234a1c37f03fc000ac57e9094d3842f139/crates/primitives/starknet/src/starknet_serde/mod.rs#L51

https://github.com/keep-starknet-strange/madara/blob/ae4771234a1c37f03fc000ac57e9094d3842f139/crates/primitives/starknet/src/transaction/types.rs#L30

Note from Abdel:
> I am wondering if we can't derive [here](https://github.com/keep-starknet-strange/madara/blob/ae4771234a1c37f03fc000ac57e9094d3842f139/crates/primitives/starknet/src/starknet_serde/mod.rs#L51) because some of those errors that have Felt252WrapperError inside. But yeah if we can derive them that's better. Worth trying

chirag-bgh:
I can take this.

rkdud007:
I wanna handle this one!

PR #477, opened 2023-05-28
unify all error enum declaration

rkdud007:
# Pull Request type

dev: unify all error enum declaration using thiserror_no_std in runtime context and this_error in client #463

Resolves: #463 

## What is the new behavior?

- Unified all error declaration using `use thiserror_no_std::Error;`.

- but the Error declared in /pallets crates, cannot unified because it depend on external error declaration which doesn't have Error derive

NO


abdelhamidbakhta (approved):
lgtm

PR #479, opened 2023-05-29
bug: Corrections on docker_run.sh and the documentation relative to it

d-roak:
This PR introduces some corrections to the `docker_run.sh` file and the documentation. It also tries to clarify the usage of the script.

# Pull Request type
Bug

- Bugfix

- It was not possible to run the `infra/docker/docker_run.sh` due to permissions
- The `infra/docker/docker_run.sh` would fail to run if we were not in the right folder
- The `Getting Started` documentation had wrong paths for running some commands
- The `Getting Started` documentation was unclear on running the `docker_run.sh`. From the text, it seemed that you should change the content of the `docker-compose.yml`, replacing the content of the command. In reality, the `docker_run.sh` is a wrapper for the `docker-compose.yml` and should be run ad-hoc

Resolves: #NA

## What is the new behavior?
- It is now possible to run the `infra/docker/docker_run.sh` as soon as you clone the repo
- We can run the `infra/docker/docker_run.sh` from any folder
- The `Getting Started` documentation was corrected. Please verify if you agree with the sentence that was "clarified"

No

## Other information
None

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -204,17 +204,15 @@ docker-compose -f infra/docker/docker-compose.yml up -d
 ```
 
 This command will firstly compile your code, and then start a local development
-network. You can also replace the default command
-(`cargo build --release && ./target/release/madara --dev --ws-external`) by
-appending your own. A few useful ones are as follow.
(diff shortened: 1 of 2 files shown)


abdelhamidbakhta (approved):
lgtm

PR #491, opened 2023-05-29
Fix CI benchmark

ClementWalter:
# Pull Request type

- Bugfix
- Build-related changes

The Benchmark Action fails not because of the performance checks but because of a file not found error.

## What is the new behavior?

Removed the loading of a file in the shared constant file but put it right in the test.

No

## Other information

The typescript part of the project could be refactored in a monorepo fashion.


abdelhamidbakhta (approved):
Lgtm

PR #494, opened 2023-05-30
feat: make chain_id configurable

chirag-bgh:
# Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #364 

## What is the new behavior?

-
-
-


crates/primitives/starknet/src/crypto/commitment/mod.rs
```diff
@@ -202,8 +202,7 @@ where
     let nonce = FieldElement::from_bytes_be(&nonce.into()).unwrap();
     let version = FieldElement::from_byte_slice_be(&version.to_be_bytes()).unwrap();
     let tx_prefix = FieldElement::from_byte_slice_be(tx_prefix).unwrap();
-    // TODO: make it configurable
-    // FIXME: https://github.com/keep-starknet-strange/madara/issues/364
+
```
LucasLvy, line 205: take the chain id as argument so you don't have it hardcoded here

LucasLvy (changes requested):
Small change otherwise looks good

crates/primitives/starknet/src/crypto/commitment/mod.rs
chirag-bgh: Then we might have to add `chain_id`  as input parameter here as well
https://github.com/keep-starknet-strange/madara/blob/151dffb9bb120e45163bf7ff3c5e366ce4b5f78a/crates/primitives/starknet/src/transaction/types.rs#L340

LucasLvy (changes requested):
Change the chain id to be a`Felt252Wrapper` in the substrate storage. Create a getter 
like that (unsafe is fine here as it has to be an utf-8 string in the genesis config (if you think it's not then don't use the unsafe version):
```rust
    #[inline(always)]
    pub fn chain_id_str() -> String {
        unsafe { from_utf8_unchecked(&Self::chain_id().0.to_bytes_be()).to_string() }
    }
    ```
and the genesis config should take the chain id as a string `SN_GOERLI` for example and then convert it into a felt to save it in the storage. 
Let me know if that's unclear

LucasLvy (approved):
lgtm

Issue #498, opened 2023-05-30
bug: Add a timeout for Benchmarking


d-roak:
# Bug Report
**Madara version:** c49f2c8

**Current behavior:**
The Benchmark workflow stays in an infinite loop when trying to connect to a failed RPC endpoint
E.g. [Benchmark#195](https://github.com/keep-starknet-strange/madara/actions/runs/5124101836)

**Expected behavior:**
After some time of failing to connect to the RPC, fail gracefully. Like done in the tests workflow

**Steps to reproduce:**
Change this line to a faulty command
https://github.com/keep-starknet-strange/madara/blob/c49f2c8bc7c3668ec3ccfac1db454b7c5cbff62a/.github/workflows/benchmark.yml#L50

apoorveth:
I can try this one!

abdelhamidbakhta:
> I can try this one!

Great!

PR #511, opened 2023-06-01
bugfix: fixed the endless loop in benchmarking if node start fails

apoorveth:
# Pull Request type

- Bugfix

If the benchmarking tests fail to start a madara node, the artillery tests keep retrying the connection and never end. This causes the benchmarking test to keep running and never fail.

Resolves: #498 

## What is the new behavior?

- Instead of using `cargo run` in the `Run benchmark` step of the workflow, we can now use `npm run test:ci` as it's more standard and any changes in CI testing will automatically reflect in benchmarking.
- `npm run test:ci` command uses the `concurrently` package to start the node (1) and run the tests (2). The command has now been configured to kill other commands if any one of the command fails (-k option). 
    - If the node fails, the benchmarking fails and exits with code 1. 
    - If the node runs, the tests are executed. After test execution, the node is killed and the entire command exits with the status of test execution (--success 'command-1')

No

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -47,8 +47,7 @@ jobs:
       - name: Run benchmark
         run: |
           cd benchmarking
-          cargo run --release -- --dev  --pool-limit=100000 --pool-kbytes=500000 --rpc-methods=unsafe --rpc-cors=all --in-peers=0 --out-peers=1 --no-telemetry &
-          npm run test
+          npm run test:ci
       - name: Compare result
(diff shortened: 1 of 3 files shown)


abdelhamidbakhta (approved):
lgtm

d-roak:
Sorry bumping a merged issue. Why the extra args in `scripts/run-node.sh`?

apoorveth:
Hey, these were being used already in benchmarking.yml and run-node.sh wasn't being used anywhere else, so I changed `run_node.sh` to be compatible with the already being used arguments

PR #513, opened 2023-06-01
Add host spec in benchmark

ClementWalter:
# Pull Request type

- Build-related changes

The spec of the host machines are not reported in the benchmark

## What is the new behavior?

Spec are added in the tooltip for each data point.

No

## Other information

The GH action actually dumps the data in the follows [data file](../blob/gh-pages/dev/bench/data.js)
When needed, we could update the default FE [index.html](../blob/gh-pages/dev/bench/index.html) to do more data viz.


abdelhamidbakhta (approved):
Lgtm

Issue #529, opened 2023-06-02
dev: In the client, read the starkware block form the Header's Digest, not from the cold storage


tdelabro:
# Current state

Currently, in the client (mostly RPCs), in order to retrieve a past Starknet block, we read the Substrate storage of the corresponding Substrate block.

It is done in `crates/client/storage/src/overrides/schema_v1_override.rs`
```rust
 fn current_block(&self, block_hash: B::Hash) -> Option<StarknetBlock> {
        self.query_storage::<StarknetBlock>(
            block_hash,
            &StorageKey(storage_prefix_build(PALLET_STARKNET, STARKNET_CURRENT_BLOCK)),
        )
        .map(Into::into)
    }
```

# Problem

This is a waste of time because the Starknet block is already present in the Substrate block header.
We should rather get it from there. 

# How to improve?

The [sc_client_api](https://paritytech.github.io/substrate/master/sc_client_api/index.html)::[HeaderBackend](https://paritytech.github.io/substrate/master/sc_client_api/trait.HeaderBackend.html#) traits provide a really handy `header` method:
```rust
fn header(
        &self,
        hash: <Block as Block>::Hash
    ) -> Result<Option<<Block as Block>::Header>, Error>;
```

So instead of using our `client: Arc<C>` to read the storage, let use it to get the header. Then, we just have to get this header Digest (the trait `Header` provide with `digest` method), then, just call the `fn find_starknet_block(digest: &Digest) -> Result<StarknetBlock, FindLogError>` defined in `crates/primitives/digest-log/src/lib.rs`.

It will be part of your mission to provide your fellow contributors with a nice interface to do all this work. Come up with your best implem :)

chirag-bgh:
I can take this.

PR #532, opened 2023-06-02
perf: read block from header's digest

chirag-bgh:
# Pull Request type

- Feature/Performance

Resolves: #529 

## What is the new behavior?

-
-
-

No


crates/client/storage/src/overrides/schema_v1_override.rs
```diff
@@ -58,12 +63,11 @@ where
     C: HeaderBackend<B> + StorageProvider<B, BE> + 'static,
     BE: Backend<B> + 'static,
 {
-    fn current_block(&self, block_hash: B::Hash) -> Option<StarknetBlock> {
-        self.query_storage::<StarknetBlock>(
-            block_hash,
-            &StorageKey(storage_prefix_build(PALLET_STARKNET, STARKNET_CURRENT_BLOCK)),
-        )
-        .map(Into::into)
+    fn current_block(&self, block_hash: <B as BlockT>::Hash) -> Option<StarknetBlock> {
```
tdelabro, line 66: It has no reason to be in the StorageOverride trait.
We are not reading the storage, we are not using override.
In think you can just a convenient, well named, wrapper function in a `utils.rs` file in rpc

tdelabro (changes requested):

tdelabro (changes requested):

crates/client/rpc-core/src/utils.rs
```diff
@@ -211,7 +210,7 @@ fn _to_btree_map_entrypoints(
 }
 
 /// Returns the current Starknet block from the block header's digest
-pub fn get_block_by_block_hash<B, C, BE>(client: Arc<C>, block_hash: <B as BlockT>::Hash) -> Option<StarknetBlock>
+pub fn get_block_by_block_hash<B, C, BE>(client: &C, block_hash: <B as BlockT>::Hash) -> Option<StarknetBlock>
```
tdelabro, line 213: I think you can get away with much less constraints:

```rust

pub fn get_block_by_block_hash<B, C, BE>(client: &C, block_hash: <B as BlockT>::Hash) -> Option<StarknetBlock>
where
    B: BlockT,
    C: HeaderBackend<B>,
{
    let header = client.header(block_hash).ok().flatten()?;
    let digest = header.digest();
    let block = find_starknet_block(digest).ok()?;
    Some(block)
}
```

It will also lift the need for the `'static` everywhere

tdelabro (changes requested):

tdelabro (approved):

crates/client/rpc-core/Cargo.toml
```diff
@@ -26,8 +26,13 @@ serde = { workspace = true, default-features = true }
 serde_json = { workspace = true }
 sp-blockchain = { workspace = true, default-features = true }
 sp-runtime = { workspace = true, default-features = true }
+mp-digest-log = { workspace = true }
+sp-storage = { workspace = true, default-features = true }
```
tdelabro, line 30: I think this one is not used anymore

crates/client/rpc-core/Cargo.toml
```diff
@@ -26,8 +26,13 @@ serde = { workspace = true, default-features = true }
...
 starknet_api = { workspace = true, default-features = false }
 frame-support = { workspace = true }
 sp-core = { workspace = true }
 starknet-core = { workspace = true }
 cairo-vm = { workspace = true, default-features = false }
+sc-client-api = { workspace = true, default-features = true }
```
tdelabro, line 36: This one neither

crates/client/storage/Cargo.toml
```diff
@@ -17,6 +17,8 @@ repository = "https://github.com/keep-starknet-strange/madara"
 frame-support = { workspace = true }
 frame-system = { workspace = true }
 mp-starknet = { workspace = true, default-features = true }
+mp-digest-log = { workspace = true }
+mc-rpc-core = { workspace = true }
```
tdelabro, line 21: Same for those two

tdelabro (changes requested):
Cargo.toml dependencies need a cleanup

tdelabro (approved):

Issue #537, opened 2023-06-05
infra: Add automation code for cluster instantiation


d-roak:
# Feature Request

**Describe the Feature Request**
Add automation code to deploy Madara network clusters.

**Describe Preferred Solution**
Use the chainspec file name as argument to create a testnet of Madara instances. We will use Terraform for the infrastructure deployment and Ansible for the configuration, adapting [LambdaClass substrate cluster instantiation](https://github.com/lambdaclass/substrate_cluster).
We also should support different [providers](https://registry.terraform.io/browse/providers?product_intent=terraform) in order to allow users to use their favorite compute provider.

PR #548, opened 2023-06-05
infra: Terraform cluster instantiation

d-roak:
Added the scripts for AWS cluster installation with Terraform and the Ansible for creating and running a madara service on the nodes.

# Pull Request type
- Feature

Resolves: #537 (partially, we can add gcp and azure as other providers, and we can add the chainspec as an input)

No


abdelhamidbakhta (approved):
lgtm

PR #549, opened 2023-06-06
fix: Refactor docker for multi arch

d-roak:
> Props to @ClementWalter for helping test this

- Changed Dockerfile to remove `paritytech/ci-linux` dependency (only worked for amd64 systems)
- Reduced docker image size (currently around 150mb)
- Possible to build for amd64 and aarch64 systems, Mac users can run a container without virtualization now (better performance)
- Changed `docker-run.sh` to use Dockerfile instead docker-compose

# Pull Request type
- Build-related changes

No


infra/docker/Dockerfile
```diff
@@ -1,37 +1,25 @@
...
-  rm -rf /usr/bin /usr/sbin && \
-  # check if executable works in this container
-  /usr/local/bin/madara --version
-
-USER node-dev
-
-EXPOSE 30333 9933 9944 9615
-VOLUME ["/chain-data"]
-
 COPY ./infra/chain-specs /chain-specs
```
tarrencev, line 12: can we keep this in the final image?

d-roak: what? the user? this is the builder image, it's not the execution env

tarrencev: the chain-specs, so they can be easily used inside the execution env without having to mount an external volume

infra/docker/Dockerfile
```diff
@@ -19,7 +18,9 @@ LABEL description="Madara, a blazing fast Starknet sequencer" \
   documentation="https://docs.madara.wtf/"
 
 COPY --from=builder /madara/target/release/madara /
+COPY --from=builder /madara/infra/chain-specs /chain-specs
 # 9444 JSON-RPC server
 # 9615 Prometheus exporter
+# 30333 P2P communication
 EXPOSE 9933 9944 9615 30333
```
tarrencev: @abdelhamidbakhta mentioned rpc and ws are both on 9933 now. so we can close 9944 and update the comment on line 22

abdelhamidbakhta (approved):
Lgtm

Issue #551, opened 2023-06-06
bug: Madara RPC doesn't handle "pending" block id


ClementWalter:
# Bug Report

**Current behavior:**

When sending a request with `block_id: "pending"`, the RPC returns "block not found".

**Expected behavior:**

Working

**Steps to reproduce:**

```
curl 0.0.0.0:9933 -H 'content-type: application/json'  --data-raw '{"method":"starknet_getNonce","jsonrpc":"2.0","params":{"contract_address": "0x2", "block_id": "pending"},"id":0}' --compressed
```

**Related code:**

[substrate_block_hash_from_starknet_block](../blob/974d7806ddb7765533c8980bc2ee0570548f2f51/crates/client/rpc/src/lib.rs#L126)

**Other information:**

We would probably just return the same as the "latest" because this "pending" state is not expected to last long enough on madara.

ptisserand:
Can I take this one?
To be sure that I have understood, when ending it should also returned the hash?

PR #559, opened 2023-06-06
fix: support "pending" block id (#551)

ptisserand:

Resolves: #551

## What is the new behavior?

When sending a request with block_id: `pending`, the RPC now returns the same as the block_id: `latest`.

No


tests/tests/test-rpc/test-starknet-rpc.ts
```diff
@@ -1146,4 +1146,22 @@ describeDevMadara("Starknet RPC", (context) => {
       });
     });
   });
+
+  describe("Fix #551: Madara RPC doesn't handle 'pending' block id", async () => {
+    it("should support 'pending' block id", async function () {
```
ClementWalter, line 1151: could you maybe put this then in the used rpc method?

```
describe("getNonce", async () => {
    it("shoule return same nonce for pending and latest", async () => {
        /// ...
    })
})

ptisserand: For me this test is not really related to `getNonce` that why I put in another group.
But for sure I can move it in `getNonce` group if it makes more sense.

abdelhamidbakhta (approved):
lgtm

PR #560, opened 2023-06-06
Update nix shell

fracek:
# Pull Request type

- Build-related changes

Currently the nix build environment relies on an outdated `shell.nix` script that contains a version of the rust compiler that's too old to compile the project.

## What is the new behavior?

I updated the project to use a `flake.nix` that provides a shell:

 * rust version is picked up from `rust-toolchain.toml`, so it will be always in sync
 * dependencies are pinned in the `flake.lock` file

No


abdelhamidbakhta (approved):
Lgtm

PR #561, opened 2023-06-07
feat: remove reductant event storage

chirag-bgh:
# Pull Request type

- Feature

Resolves: #556


abdelhamidbakhta:
Can you check that it does not decrease performances please ? Can you measure before / after this change on your local machine and publish results here ?

chirag-bgh:
> Can you check that it does not decrease performances please ? Can you measure before / after this change on your local machine and publish results here ?

@abdelhamidbakhta It seems this change decrease the avg TPS: 

Before at commit d7f15ee:
image


After: 
image

Issue #567, opened 2023-06-07
feat: add ERC721 as part of the genesis state


bajpai244:
# Feature Request

#502 requests for NFT benchmarking to be added, till the issue around `declare` and nonce race condition in benchmarking is fixed, it is better to make ERC721 part of the genesis state.

When there is a reliable way to benchmark, the deployed ERC721 can be used out of the box to benchmark, as its tested version would already exist on the sequencer.

Hence, the issue requests for a feature to add a Cairo0 ERC721 to be part of the genesis state, along with tests for its `mint` functionality, which will make it ready to use when benchmarking is solved. 

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes

bajpai244:
@abdelhamidbakhta as discussed on the chat, you can assign this to me.

Issue #530, opened 2023-06-02
feat: add UDC to the genesis state of the chain 


bajpai244:
# Add UDC to the genesis state

[UDC](https://medium.com/starknet-edu/deploying-to-starknet-with-the-universal-deployer-contract-c6de07092bfb) is via how Starknet contracts are now supposed to be deployed, having it deployed in the genesis state of the chain can be very helpful.

**Describe Preferred Solution**

Deploy it as part of the genesis state of the chain, just like how ERC20 and wallet contracts are already deployed as part of the genesis state.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes

bajpai244:
@abdelhamidbakhta as discussed in the chat, I would be interested in taking it :)

tdelabro:
@bajpai244 go ahead!

xJonathanLEI:
Please add it to this address: `0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf`. It's the expected address I believe.

bajpai244:
@xJonathanLEI thanks for the heads up!

bajpai244:
Update: #599 was closed by mistake when I synced my fork, I will open another PR that fixes this in ~45 mins!

PR #568, opened 2023-06-07
feat: add ERC721 to genesis state

bajpai244:
# Pull Request type

- Feature

The current genesis state doesn't have ERC721 as part of the genesis state, the PR aims to make support for it along with the relevant test for its `mint` functionality.

Resolves: #567 

## What is the new behavior?

The ERC721 implementation will be added, along with tests.

This will also ensure that when the UDC is added to the genesis state #530, even if `declare` doesn't work, users can deploy ERC721 to Madara. 

NO.


bajpai244:
The deployed contract is from [openzepplin](https://github.com/OpenZeppelin/cairo-contracts/tree/main/src/openzeppelin/token/erc721), the MintableBurnable version is deployed, [here](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/token/erc721/presets/ERC721MintableBurnable.cairo) is the link for the same.

abdelhamidbakhta (approved):
lgtm

bajpai244:
lgtm 🚀

Issue #517, opened 2023-06-01
dev: unit testing for `get_events` rpc logic


tdelabro:
There is no unit testing for the logic of the `get_events` rpc.

What needs to be testing:
- filtering the events of a single block
- filtering events of multiple blocks
- calls to the RPC (already in place)

Doing so simply splits the code into multiple smaller testable functions.
We also have to find a way to either mock or perform dependency injection in order to test those functions without the whole client running.

apoorveth:
Can I try this one?

tdelabro:
Sure. It's all yours, contact either me @timothee_delabrouille or Greg @greged93 (who developed the RPC) on telegram if you need help understanding the code or on the direction to take :)

PR #580, opened 2023-06-08
Get events unit tests

apoorvsadana:
# Pull Request type

- Testing

We only have integration tests for get_events. So we need to run the client, add blocks, call get_events and check if it's working as expected.

Resolves: #517 

## What is the new behavior?

Unit tests have been added and the test structure has been setup so that we can now test rpc methods and internal functions without having to run the client.

No


crates/client/rpc/src/lib.rs
```diff
@@ -150,7 +153,85 @@ where
...
+    /// * `address` - Address to use to filter the events
+    /// * `keys` - Keys to use to filter the events. An event is filtered if any key is present
+    /// * `max_results` - Optional, indicated the max events that need to be filtered
+    ///
+    /// # Returns
+    ///
+    /// * `(block_events: Vec<EventWrapper>, continuation_token: usize)` - A tuple of the filtered
+    ///   events and the first index which still hasn't been processed block_id and an instance of
+    ///   Block
+    pub fn filter_events_by_params(
```
tdelabro, line 202: This does not need to be part of a `impl` block.
It does not use `self` at all. It can just be a regular function.
This way you can test it without building the `Starknet` type.

And you can test it here, in `client/rpc`, rather than in `node`

crates/client/rpc/src/lib.rs
```diff
@@ -207,36 +264,42 @@ where
                 continue;
             }
 
-            let block_events = block_events.into_iter().skip(continuation_token);
+            let block_events = block_events.into_iter().skip(continuation_token).collect();
```
tdelabro, line 267: You are collecting twice, which is a waste we tried to avoid when we originally designed this method

crates/node/Cargo.toml
```diff
@@ -83,6 +84,10 @@ blockifier = { workspace = true }
 hex = { workspace = true }
 starknet-core = { workspace = true }
 
+# Testing
+tokio = "1.28.2"
+serial_test = "2.0.0" # to run rpc tests sequentially, else all tests try to call rpc methods at the same time and we get lock errors
+
```
tdelabro, line 90: Testing import should go into `[dev-dependencies]`

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
apoorvsadana: moved it outside the impl block and move the test cases within rpc

crates/client/rpc/src/lib.rs
apoorvsadana: removed the collect here and changed the argument type of the filter function

crates/client/rpc/src/lib.rs
```diff
@@ -937,3 +962,47 @@ where
...
+/// * `events` - A vector of all events
+/// * `address` - Address to use to filter the events
+/// * `keys` - Keys to use to filter the events. An event is filtered if any key is present
+/// * `max_results` - Optional, indicated the max events that need to be filtered
+///
+/// # Returns
+///
+/// * `(block_events: Vec<EventWrapper>, continuation_token: usize)` - A tuple of the filtered
+///   events and the first index which still hasn't been processed block_id and an instance of Block
+pub fn filter_events_by_params(
```
tdelabro, line 979: move it in a different files `events.rs`
Along with the `get_block_events` method and basically anything event-related you can find in the file.

It will help readablility and code maintaining

crates/client/rpc/src/lib.rs
```diff
@@ -937,3 +962,47 @@ where
...
+    keys: Vec<Vec<FieldElement>>,
+    max_results: Option<usize>,
+) -> (Vec<EventWrapper>, usize) {
+    let mut filtered_events = vec![];
+    let mut index = 0;
+
+    // Iterate on block events.
+    for event in events {
+        index += 1;
+        let match_from_address = address.map_or(true, |add| add == event.from_address);
```
tdelabro, line 991: `add` is a bad abreviation for address. It sound like addition
addr is better

crates/client/rpc/src/lib.rs
```diff
@@ -150,7 +158,39 @@ where
...
+        let block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).ok_or_else(|| {
+            error!("Failed to retrieve block");
+            StarknetRpcApiError::BlockNotFound
+        })?;
+        let block_events = self
+            .overrides
+            .for_block_hash(self.client.as_ref(), substrate_block_hash)
+            .events(substrate_block_hash)
+            .unwrap_or_else(|| {
+                info!("No events found in block {}", block_id);
```
tdelabro, line 187: info! is probably too strong.
dbg! should be enough

tdelabro (changes requested):

crates/client/rpc/Cargo.toml
```diff
@@ -52,3 +52,7 @@ jsonrpsee = { workspace = true, default-features = true, features = [
 log = { workspace = true, default-features = true }
 serde_json = { workspace = true, default-features = true }
 thiserror = { workspace = true }
+
+[dev-dependencies]
+rstest = { workspace = true, default-features = true }
```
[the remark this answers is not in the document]
apoorvsadana: removed default-features from here

crates/client/rpc/src/lib.rs
apoorvsadana: moved `filter_events_by_params` to a new file, however `get_block_events` depends on `self`, so should that stay inside the implementation

crates/client/rpc/src/lib.rs
apoorvsadana: fixed

crates/client/rpc/src/tests.rs
```diff
@@ -0,0 +1,134 @@
...
+	_name: &'a str,
+	events:Vec<EventWrapper>,
+	filter_keys: Vec<Vec<FieldElement>>,
+	filter_address: Option<Felt252Wrapper>,
+	max_results: Option<usize>,
+	expected_events: Vec<EventWrapper>,
+	expected_continuation_token: usize,
+}
+
+#[fixture]
```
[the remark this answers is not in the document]
apoorvsadana: added `#[once]`

crates/client/rpc/src/lib.rs
apoorvsadana: fixed

crates/client/rpc/src/lib.rs
tdelabro: you can have multiple `impl MyType {}` blocks

apoorvsadana: moved to `events.rs`

abdelhamidbakhta:
@tdelabro can you do a final review please ? @apoorvsadana applied your suggestions

apoorvsadana:
there was a conflict with `.gitignore` after the recent addition of `.vscode`, just resolved that

crates/client/rpc/src/events.rs
```diff
@@ -1,10 +1,66 @@
...
+use crate::Starknet;
+
+impl<B, BE, C, P, H> Starknet<B, BE, C, P, H>
+where
+    B: BlockT,
+    C: HeaderBackend<B> + StorageProvider<B, BE> + 'static,
+    C: ProvideRuntimeApi<B>,
+    C::Api: StarknetRuntimeApi<B> + ConvertTransactionRuntimeApi<B>,
+    BE: Backend<B>,
+    H: HasherT + ThreadSafeCopy,
```
tdelabro, line 29: Are you sure that everything here is absolutely necessary?
Did you try removing some of those

crates/client/rpc/src/lib.rs
```diff
@@ -157,38 +154,6 @@ where
...
-            .overrides
-            .for_block_hash(self.client.as_ref(), substrate_block_hash)
-            .events(substrate_block_hash)
-            .unwrap_or_else(|| {
-                dbg!("No events found in block {}", block_id);
-                Vec::new()
-            });
-        Ok((block_events, block))
-    }
-
```
tdelabro: `filter_events` can also be moved to the events.rs file

crates/client/rpc/src/tests.rs
```diff
@@ -0,0 +1,125 @@
+use mp_starknet::execution::types::{ContractAddressWrapper, Felt252Wrapper};
```
tdelabro, line 1: the file hierary could be:
src/
   events/
     mod.rs
     tests.rs
     
     
Because all the tests in this file are event related

crates/client/rpc/src/tests.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[rstest]
+#[case::filter_keys(build_test_case()[0].clone())]
+#[case::filter_address(build_test_case()[1].clone())]
+#[case::filters_keys_and_address(build_test_case()[2].clone())]
+#[case::filter_max_results_met(build_test_case()[3].clone())]
+#[case::filter_max_results_not_met(build_test_case()[4].clone())]
+#[case::filter_keys_less_than_actual(build_test_case()[5].clone())]
+#[case::filter_keys_more_than_actual(build_test_case()[6].clone())]
```
tdelabro, line 105: There is no other edge case you could think of?
Better too much that not enough

tdelabro (changes requested):

crates/client/rpc/src/events.rs
apoorvsadana: Yes, I just checked again as well, everything is necessary

crates/client/rpc/src/tests.rs
apoorvsadana: makes sense! fixed.

crates/client/rpc/src/lib.rs
apoorvsadana: moved

crates/client/rpc/src/tests.rs
apoorvsadana: I thought a little more, I have added two new test cases

- Filter events when there are no filters (return all events)
- Filter events when there are no events (return empty vector)

Apart from these, there are few test cases which I added before, like

- filter_address_chunk_size_met
- filter_address_chunk_size_not_met
- filter_keys_chuck_size_met
- filter_keys_chunk_size_not_met

These felt redundant and I thought it made it challenging to filter the useful cases, so I removed them. However, if you feel we should keep these to protect code changes later, I can add them back.

PR #587, opened 2023-06-08
Feat: short cmd for running madara on the testnet

d-roak:
Added a short command for running a full node with `madara --testnet sharingan`, in the future we can add new testnets by adding the chain-spec in the chain-specs folder, and adding the name of the testnet in the enum.

# Pull Request type
- Feature

Changed the location of `chain-specs` folder


abdelhamidbakhta (approved):
lgtm

PR #599, opened 2023-06-09
feat: add UDC to genesis state 🥷 

bajpai244:
# Pull Request type

- Feature

Madara doesn't have UDC as part of the genesis state, but it is an important contract that needs to be present.

Resolves: #530 

## What is the new behavior?

UDC is deployed, the address of the deployed UDC is the same as on the other networks, which is **"0x041a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf"**, hence it should make it compatible with existing tooling.

NO

## Other information

The contract has been taken from [OpenZepplin.](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/utils/presets/UniversalDeployer.cairo)

The test includes deploying an ERC-20 token, which is already declared as part of the genesis state of the sequencer.

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -71,7 +71,7 @@ use blockifier::block_context::BlockContext;
 use blockifier::execution::entry_point::{CallInfo, ExecutionResources};
 use blockifier_state_adapter::BlockifierStateAdapter;
 use frame_support::pallet_prelude::*;
-use frame_support::traits::Time;
+use frame_support::traits::UnixTime;
 use frame_system::pallet_prelude::*;
 use mp_digest_log::MADARA_ENGINE_ID;
 use mp_starknet::block::{Block as StarknetBlock, Header as StarknetHeader, MaxTransactions};
@@ -139,7 +139,7 @@ pub mod pallet {
         /// The hashing function to use.
         type SystemHash: HasherT + DefaultHasher + CryptoHasherT;
         /// The time idk what.
-        type TimestampProvider: Time;
+        type TimestampProvider: UnixTime;
         /// A configuration for base priority of unsigned transactions.
         ///
         /// This is exposed so that it can be tuned for particular runtime, when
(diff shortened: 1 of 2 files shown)


tests/tests/test-starknet/test-starknet-extrinsics.ts
```diff
@@ -99,4 +105,26 @@ describeDevMadara("Pallet Starknet - Extrinsics", (context) => {
...
+        context.polkadotApi,
+        CONTRACT_ADDRESS,
+        TOKEN_CLASS_HASH,
+        numberToHex(1, 256),
+        false,
+        3
+      )
+    );
+
+    expect(
```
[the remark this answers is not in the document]
bajpai244: I have followed the approach used in earlier tests, where extrinsic success was used as a metric.

For example, the [deploy test here](https://github.com/keep-starknet-strange/madara/blob/d862d1af49b180a4c596348500ce21832a191353/tests/tests/test-starknet/test-starknet-extrinsics.ts#L44) relied on the same, so I also followed the same test for the time being.

I also think extrinsic testing needs improvement, I guess we need a separate issue and PR addressing all extrinsic tests. For example, in the ERC-20 transfer, we should also check for the balance change.

tests/tests/test-starknet/test-starknet-extrinsics.ts
ClementWalter, line 1: no integration test?

bajpai244: can you elaborate a bit more in this context on what would an integration test look like?

ClementWalter: a test using starknet js, in https://github.com/keep-starknet-strange/madara/blob/d862d1af49b180a4c596348500ce21832a191353/tests/tests/test-rpc/test-starknet-rpc.ts#L495 `it("should deploy contract")` or something like this

bajpai244: Make sense!

bajpai244:
I forked my sync, forgetting about this, and it closed this PR.

I am making another PR within ~45 mins with the integration test that can be merged!

Issue #605, opened 2023-06-09
dev: Declare all dependencies on workspace level when used in more than one crate


abdelhamidbakhta:
# Description

Go over the dependencies of each create of the project, and try to declare dependencies at the workspace level when they are used in more than one crate. It will be easier to upgrade versions, or being able to switch dependencies.

makluganteng:
Can i try this ?

abdelhamidbakhta:
> Can i try this ?

Sure, assigned to you

PR #608, opened 2023-06-09
Moving the dependency to workspace level

makluganteng:
# Pull Request type

- Refactoring

Resolves: #605 

## What is the new behavior?

- Moving the dependency that is called and not declared in the workspace level

No


abdelhamidbakhta (approved):
Lgtm

Issue #564, opened 2023-06-07
bug: block timestamp should be in seconds not milliseconds


xJonathanLEI:
# Bug Report

**Madara version:**

[v0.1.0-testnet-sharingan-alpha.4](https://github.com/keep-starknet-strange/madara/releases/tag/v0.1.0-testnet-sharingan-alpha.4)

**Current behavior:**

Block timestamps are be in UNIX milliseconds.

**Expected behavior:**

Block timestamps should be in UNIX seconds.

**Steps to reproduce:**

```console
$ starkli block-time --rpc "https://sharingan.starknet.rs/" --unix
1686124104000
```

**Related code:**

N/A

**Other information:**

N/A

abdelhamidbakhta:
Oh btw this issue is already assigned to @pierrot498 . But please you can take another one

lambda-0x:
@pierrot498 are you still working on this? if not i can take this up.

lambda-0x:
@pierrot498 are you still working on this? if not I can take this up.

PR #610, opened 2023-06-10
fix: use `UnixTime` instead of `Time`

lambda-0x:
# Pull Request type

- Bugfix

Resolves: #564

## What is the new behavior?

- Change the `TimestampProvider` to use `UnixTime`
- Update test to expect new behaviour

Maybe, @abdelhamidbakhta wdyt?


dbejarano820 (approved):

abdelhamidbakhta (approved):
Lgtm

Issue #329, opened 2023-05-03
Use real gas price


LucasLvy:
# Description

For now we use a fixed fake gas price (10). We should make it dynamic based on the L1 gas price.

# Acceptance criteria

* gas price is correct and dynamic

dbejarano820:
If no one is taking this one on, I'd love to give it a shot!

chee-chyuan:
> If no one is taking this one on, I'd love to give it a shot!

sorry im supposed to take this. stuck at something atm

abdelhamidbakhta:
> If no one is taking this one on, I'd love to give it a shot!

Assigned to you

edisontim:
If this is stale, I would love to be able to work on this one!

edisontim:
> Hey @edisontim, I'm currently working on it! I'd love to get your take on my approach once I open a PR

Ok! I'll try to keep an eye out for when this is being reviewed

dbejarano820:
Still WIP

chirag-bgh:
@LucasLvy I would like to take this on.

abdelhamidbakhta:
> @LucasLvy I would like to take this on.

Assigned to you

apoorvsadana:
@tdelabro I guess we should open this as well?

EvolveArt:
> What is the archi there:
> 
> 1. query the current gas price to some oracle
> 2. insert it as the beginning of the block as an inherent block value
> 3. use it in the runtime
> 
> Works for a centralized setup.
> 
> @apoorvsadana @EvolveArt wdyt?


Remind the reason why we can't get this from L1 directly querying some full node ?
Althoug I agree that this solution would be fine for now it's far from ideal.

apoorvsadana:
Can we ask Starkware what function are they using on top of the L1 gas price for fee estimation? There is very old [post](https://community.starknet.io/t/fees-in-starknet-alpha/286) on the community forum which talks about using a markup.

> The fee calculated may include a markup to compensate for L1 gas cost fluctuations.

But its almost 2 years since then, maybe they have a mature solution or they know what constants work.

tdelabro:
superseeded by #1573

PR #611, opened 2023-06-11
feat: use dynamic gas price based on L1 gas price

chirag-bgh:
# Pull Request type

- Feature

Resolves: #329 

## What is the new behavior?

- Fetches L1 gas price using `eth_gasPrice` JSON-RPC method

No


crates/pallets/starknet/src/lib.rs
```diff
@@ -715,6 +747,17 @@ pub mod pallet {
             // otherwise we have a nonce error and everything fails.
             // Once we have a real fee market this is where we'll chose the most profitable transaction.
 
+            // Firstly let's check that we call the right function.
+            if let Call::submit_gas_price_unsigned { gas_price: new_gas_price } = call {
+                // We can only accept transactions with a priority lower than `UnsignedPriority`.
+                return ValidTransaction::with_tag_prefix("starknet")
+                    .priority(u64::MAX)
+                    .and_provides(new_gas_price)
```
LucasLvy, line 755: and_provies adds a tag to the transaction don't set that

crates/pallets/starknet/src/lib.rs
```diff
@@ -815,8 +879,12 @@ impl<T: Config> Pallet<T> {
         let provided_seq_addr = Self::sequencer_address();
         let sequencer_address = ContractAddress(starknet_api::api_core::PatriciaKey(StarkFelt(provided_seq_addr)));
         let vm_resource_fee_cost = HashMap::default();
-        // FIXME: https://github.com/keep-starknet-strange/madara/issues/329
-        let gas_price = 10;
+
+        let gas_price = match Self::gas_price_l1() {
+            0 => 10, // default gas price
```
LucasLvy, line 884: 10 was just a random value you can use unwrap_or_default here imo

crates/pallets/starknet/src/offchain_worker/mod.rs
```diff
@@ -16,20 +16,24 @@ use crate::{Config, Pallet, ETHEREUM_EXECUTION_RPC};
...
-        // Get the last known block from storage.
-        let last_known_eth_block = Self::last_known_eth_block().ok_or(OffchainWorkerError::NoLastKnownEthBlock)?;
         // Query L1 for the last finalized block.
         let raw_body = query_eth(LAST_FINALIZED_BLOCK_QUERY)?;
         let last_finalized_block: u64 = from_slice::<EthGetBlockByNumberResponse>(&raw_body)
             .map_err(|_| OffchainWorkerError::SerdeError)?
             .try_into()?;
+
+        // Get the last known block from storage.
+        let last_known_eth_block = Self::last_known_eth_block().ok_or(OffchainWorkerError::NoLastKnownEthBlock)?;
```
LucasLvy, line 35: why change this ?

LucasLvy (changes requested):

crates/pallets/starknet/src/offchain_worker/mod.rs
chirag-bgh: I just changed this to query first because this always returned `NoLastKnownEthBlock` though the rpc_url was wrong.

crates/pallets/starknet/src/lib.rs
chirag-bgh: OKay

crates/pallets/starknet/src/lib.rs
chirag-bgh: But won't that set the value to be 0 always? Or there's way to set custom default value?

LucasLvy: nah it would set it to 0 but that's not a problem for now

chirag-bgh: so should change the value in tests accordingly?

LucasLvy: yes plz

abdelhamidbakhta:
closing as there is no sign of activity, do you still want to tackle it ?

PR #318, opened 2023-05-01
chore: use starknet-rs from crates.io

xJonathanLEI:
# Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [x] Build-related changes
- [ ] Documentation content changes
- [ ] Testing
- [ ] Other (please describe):

# What is the current behavior?

Installs `starknet-rs` crates from GitHub.

Issue Number: N/A

# What is the new behavior?

Installs `starknet-rs` crates from crates.io instead.

# Does this introduce a breaking change?

- [ ] Yes
- [x] No

# Other information

Because it's generally a better practice to install from crates.io :)

PR #618, opened 2023-06-12
chore: use starknet-rs crates from crates.io

xJonathanLEI:
# Pull Request type

- Build-related changes

Using [`starknet-rs`](https://github.com/xJonathanLEI/starknet-rs) crates from GitHub directly due to staled versions on crates.io.

A previous merged attempt (#318) tried to do the exact same thing, but was later reverted due to the need for using crates not being maintained on crates.io.

## What is the new behavior?

All [`starknet-rs`](https://github.com/xJonathanLEI/starknet-rs) crates are downloaded from crates.io.

We can do this now due to support for jsonrpc spec 0.3.0 being merged (https://github.com/xJonathanLEI/starknet-rs/pull/396) into `master` and released to crates.io.

No

## Other information

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4497,7 +4497,7 @@ dependencies = [
  "sp-core",
  "sp-runtime",
  "starknet-core",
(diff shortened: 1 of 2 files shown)


abdelhamidbakhta (approved):
lgtm

PR #619, opened 2023-06-13
fix: revert back to using Time which modified block_timestamp function

lambda-0x:
# Pull Request type

- Bugfix

Resolves: #564 
Previous PR: #610 

## What is the new behavior?

- `UnixTime` [logs an error](https://github.com/paritytech/substrate/blob/85415fb3a452dba12ff564e6b093048eed4c5aad/frame/timestamp/src/lib.rs#L299-L314) if its called in genesis block 
- So instead of that just use `Time` with modified `block_timestamp` function

No


abdelhamidbakhta (approved):
lgtm

Issue #617, opened 2023-06-12
dev: improve extrinsic testing


bajpai244:
The current extrinsic testing in [test-starknet-extrinsics](https://github.com/keep-starknet-strange/madara/blob/main/tests/tests/test-starknet/test-starknet-extrinsics.ts) needs some improvement.

If we look closely at these tests, they are just testing for events being emitted marking success, a better way to test should be to test state transitions that occur as part of calls to these extrinsic.

For example:
- In case of deployment: check that the class hash is the one for which the contract has been deployed at that address
- Token transfer should check for changes in the balance
- Mint should check for NFT with the given token id being available and owned by the specific 
- UDC deployment should check the contract was deployed successfully

~~+ Make class hash declaration test work!~~ 

Right now the test is skipped, but it should also be working

bajpai244:
@abdelhamidbakhta I would love to take this!

abdelhamidbakhta:
> @abdelhamidbakhta I would love to take this!

Assigned to you my friend

bajpai244:
NOTE: 

Updating the issue here, as mentioned in PR #631, we should skip bringing fixing the `declare` test into the scope of this issue.

It should be its own issue, as from my previous knowledge it requires a decent scope of work.

Issue #630, opened 2023-06-14
bug: `unique` wrong value being assigned for a given value in `deployTokenContractUDC`  


bajpai244:
# Bug Report

**Madara version:**

d1c45e178c05e1f718fcd4c783c735d9c12f0f02

**Current behavior:**

The `unique` parameter in [deployTokenContractUDC](https://github.com/keep-starknet-strange/madara/blob/d1c45e178c05e1f718fcd4c783c735d9c12f0f02/tests/util/starknet.ts#L349), for now, is assigned to the wrong value given the boolean, this went uncaught because of extrinsic testing not being very tight.

**Expected behavior:**

The right value should be assigned for a given boolean value for `unique`.

bajpai244:
My PR for #617 will resolve this as well.

bajpai244:
@abdelhamidbakhta you can close this as well, it was also resolved in #631

PR #631, opened 2023-06-14
chore: enhance extrinsic testing 🚀 

bajpai244:
# Pull Request type

- Testing

The current extrinsic testing just depends upon event emission, the testing should be more dependent on the state changes that calls to these extrinsic calls should result in { have been elaborated in #617  }.

The current implementation of[ deployTokenContractUDC](https://github.com/keep-starknet-strange/madara/blob/d1c45e178c05e1f718fcd4c783c735d9c12f0f02/tests/util/starknet.ts#L349) also has a bug, where the wrong value for `unique` is provided { have been described in #630  }

Resolves: #617 , #630 

## What is the new behavior?

Extrinsic testing involves checking for real state transitions that should result from calls to these extrinsic.

Example: 
- check for a change in equivalent balance in case of ERC20/ERC721 transfer/mint
- check for contract deployment in case of deployment { by checking class hash at address, and checking ERC20 balance of the initial amount of recipient, which is assigned in constructor }

#630 is resolved, and correct value for `unique` is assigned.

NO

### DIFF
--- a/tests/tests/test-starknet/test-starknet-extrinsics.ts
+++ b/tests/tests/test-starknet/test-starknet-extrinsics.ts
@@ -2,7 +2,7 @@ import "@keep-starknet-strange/madara-api-augment";
 
 import { expect } from "chai";
 
-import { numberToHex } from "@polkadot/util";
+import { hexFixLength, numberToHex } from "@polkadot/util";
 import { jumpBlocks } from "../../util/block";
 import { describeDevMadara } from "../../util/setup-dev-tests";
 import {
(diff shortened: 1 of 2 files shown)


bajpai244:
NOTE:

For now, to limit the scope of this PR, I haven't touched the `declare` test, it remains skipped and I have marked it `TODO`.

- I think fixing `declare` should be a different issue, and should be addressed in a different PR.
- While navigating the codebase inconsistency in using constants, where in some places rather than using the constant, value was hardcoded again. This also needs to be addressed in a different issue and a different PR!

I will make issues for them if everyone agrees.

cc: @abdelhamidbakhta

abdelhamidbakhta (approved):
Lgtm

PR #634, opened 2023-06-14
fix: block hash

EvolveArt:
Fixes block hash computation according to [starknet docs](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/header/#block_hash)


abdelhamidbakhta (approved):
lgtm

Issue #607, opened 2023-06-09
dev: Use taplo to format TOML files of the project


abdelhamidbakhta:
# Description

Use [taplo](https://github.com/tamasfe/taplo) to format TOML files of the project.
It's painful for code reviews when we don't all use same formatting rules.
- Add instructions on how to run taplo locally
- Add a formatting rules file for universal formatting
- Add a workflow in the CI to check the formal of TOML files and fail if not valid

lambda-0x:
I can take a stab at this

ApereLieZ:
There are a few questions
Is the task open?
And can tell you more about how it should be implemented: As a script that the developer can run, or as part of the build process?

abdelhamidbakhta:
> There are a few questions Is the task open? And can tell you more about how it should be implemented: As a script that the developer can run, or as part of the build process?

Hey, yes task is open.
As part of the CI workflow, and eventually adds cargo husky hook to do it + add commands in the readme to run manually

ApereLieZ:
Can I take this?

abdelhamidbakhta:
> Can I take this?

please go, assigned to you

PR #642, opened 2023-06-14
dev: Use taplo to format TOML files of the project

ApereLieZ:
# Pull Request type

Resolves: #607

## What is the new behavior?

- Add instructions on how to run taplo locally
- Add a formatting rules file for universal formatting
- Add a workflow in the CI to check the formal of TOML files and fail if not valid
- add husky hooks with TOML file validation (need to run hooks once)

Yes
You must check the TOML files before pushing, otherwise ci/cd will fail.
To prevent this behavior, enable husky hooks
 
```bash
cargo test 
```


EvolveArt (changes requested):
As said in madara channel :
> couldn't we just use the taplo rust crate. Not a huge fan of having to add npm dependencies at the root of the project..

We should rather use [cargo-husky](https://github.com/rhysd/cargo-husky) there imo

.github/workflows/test.yml
```diff
@@ -78,3 +78,10 @@ jobs:
         run: |-
           cd tests
           npm run test
+  toml-lint:
```
EvolveArt, line 81: add spacing here

EvolveArt:
@ApereLieZ why does it still use `npx @taplo/cli` in the CI ? I don't get why we cannot just use the [taplo](https://crates.io/crates/taplo/0.12.0) crate ?

ApereLieZ:
> @ApereLieZ why does it still use `npx @taplo/cli` in the CI ? I don't get why we cannot just use the [taplo](https://crates.io/crates/taplo/0.12.0) crate ?

From my measurements in terms of runtime, using npx is noticeably faster than using crate

EvolveArt:
> > @ApereLieZ why does it still use `npx @taplo/cli` in the CI ? I don't get why we cannot just use the [taplo](https://crates.io/crates/taplo/0.12.0) crate ?
> 
> @EvolveArt From my measurements in terms of runtime, using npx is noticeably faster than using crate

But then no need to `cargo add taplo-cli` in this case ?

EvolveArt (approved):
lgtm

ApereLieZ:
@EvolveArt 
Done

Issue #552, opened 2023-06-06
dev: replace util functions by `From` impl 🔨


EvolveArt:
Investigate if the [following](https://github.com/keep-starknet-strange/madara/blob/main/crates/client/rpc-core/src/utils.rs#L56) functions could be replaced by a `From` implementation.

**Acceptance Criterion**

The following functions are removed
- `to_tx`
- `to_declare_tx`
- `to_deploy_account_tx`
- `to_invoke_tx`

pierrot498:
Can I take this?

LucasLvy:
once #564 is merged

dbejarano820:
Let me at it!

dbejarano820:
@abdelhamidbakhta

lambda-0x:
@greged93 implementing this already resolves #625, #626, #627, #628, and #629 so maybe we can close them?

cc: @abdelhamidbakhta

abdelhamidbakhta:
> @greged93 implementing this already resolves #625, #626, #627, #628, and #629 so maybe we can close them?
> 
> cc: @abdelhamidbakhta

good point, @greged93 what do you think ?

dbejarano820:
I agree, we still need to add testing. The PR is passing all checks now, just waiting on Lucas to give it a look again

lambda-0x:
Okay I will work on adding test after 597 is merged.

PR #644, opened 2023-06-14
chore: remove duplication

lambda-0x:
# Pull Request type

Resolves: #625 

No 

## Other information

Much of this was already simplified after #552 refactor, so this is the only thing i could think of to simplify.


abdelhamidbakhta (dismissed):
Lgtm

lambda-0x:
@abdelhamidbakhta looks like a intermittent ci failure, and its not related to the changes. Can you try rerunning the job?

Issue #579, opened 2023-06-08
feat: Madara faucet


abdelhamidbakhta:
# Feature Request
 
 @zizou0x initiated the work in this [PR](https://github.com/keep-starknet-strange/madara/pull/576) to have a faucet for Madara based Appchains. 
 
 The 2 components (frontend and backend) of the faucet are defined [here](https://github.com/keep-starknet-strange/madara/blob/main/infra/starknet-stack/docker-compose.yml#L25-L43) in a docker compose file.
 
 However it is not working. 
 
 The goal of this issue it to make the faucet work with Madara so that we can start using it on testnets and for different use cases.
 
 # Acceptance criteria
 
 - Have a working faucet integrated to a Madara chain (can be a local single node chain to start with)
 - Write useful documentation about how to use the faucet and how to deploy it

zizou0x:
I noticed 2 issues: 

1. The transaction is sent from the container and Madara's RPC only allows request from localhost.
2. Even when trying to send the transaction from localhost it fails with : 
```
Failed to parse JSON-RPC params as object: Invalid params in the call: unknown variant `0x100000000000000000000000000000001`, expected `0x0` or `0x1` at line 1 column 603
```

Please reach me out on Telegram (@zizou_0x) if you need anything.

Issue #646, opened 2023-06-14
feat: support for estimateFee version


apoorvsadana:
# Feature Request

Add support for `0x100000000000000000000000000000001` version in estimateFee rpc.

Currently, the testnet and mainnet support 0x100000000000000000000000000000001 version for the estimateFee rpc call. [Starknetjs](https://github.com/starkware-libs/cairo-lang/blob/00ee90123c0bbaffe13dd05a79d2d57b0cb747be/src/starkware/starknet/definitions/constants.py#LL64C9-L64C9) also uses this version by default when calling estimateFee within the account.execute method.

**Describe Preferred Solution**

Madara currently relies on the `starknet-core` crate for this. I have created an issue over there which can be tracked [here](https://github.com/xJonathanLEI/starknet-rs/issues/402). Once this is fixed, we can directly integrate their solution with Madara.

**Describe Alternatives**

An alternate method could be to create a custom wrapper on top of `BroadcastedTransaction` to handle a custom fee version. However, that will only be a temporary solution and ideally we should have the issue fixed in the core library

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes

xJonathanLEI:
See: https://github.com/xJonathanLEI/starknet-rs/issues/402#issuecomment-1625849247.

Please test with the branch and let me know if it works. Then I can merge and release a new (breaking) version for `starknet-core`.

PR #647, opened 2023-06-14
bugfix: starkcet faucet not working

apoorvsadana:
# Pull Request type

- Bugfix

The Starkcet faucet fails to interact with Madara and collect faucet funds.

Resolves: #579 

## What is the new behavior?

- Madara doesn't support `0x100000000000000000000000000000001` version for estimateFee. The issue can be tracker [here](https://github.com/keep-starknet-strange/madara/issues/646). So we need to manually change the version to `0x1` when estimating fees.
- Docs have been added giving instructions on how to use the faucet 

No

### DIFF
--- a/docs/faucet-setup.md
+++ b/docs/faucet-setup.md
@@ -1,0 +1,136 @@
+# Madara Faucet
+
+You can have a custom faucet for your app chains!
+
+<figure class="video_container">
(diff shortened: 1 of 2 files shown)


abdelhamidbakhta (approved):
lgtm

Issue #250, opened 2023-04-18
dev: investigate best values for bounded types 🔨


EvolveArt:
We use the `BoundedVec` type in many places which requires some bound values.

For now we mostly use `u32::MAX = 4294967295` but probably it shouldn't be the case for everything. 

e.g [1](https://github.com/EvolveArt/kaioshin/commit/9ce2dab5825f67d8b8596640d9656f204b818c31#diff-e6236d7c38d69c88218836096a3fa2873f71671d87b453a9090c716b1e6fc6aeR14) ,[2](https://github.com/keep-starknet-strange/madara/commit/47d4bcf6b25997f891bb146907276325e625c95b#diff-d1e7254b6a1f394f0f8f9fee94b5cfad347b50e492ef7a6d92e7238220852faeR7)

tdelabro:
we can disable the requirement for values in storage to have a bounded size at any moment. For single storages or for the whole pallet. It's up to us

tdelabro:
I will disable the requirement for bounder types in storage in an incoming PR

PR #657, opened 2023-06-15
dev(test): add tests for TryFrom implementation for DeployAccountTransaction

lambda-0x:
# Pull Request type

- Testing

Resolves: #627 

No

## Other information

- added tests but cannot yet run all, because bounded types are very large and when trying to run them allocation for that large chunk fails
    - ref: #250

### DIFF
--- a/crates/primitives/starknet/src/tests/transaction.rs
+++ b/crates/primitives/starknet/src/tests/transaction.rs
@@ -379,7 +379,9 @@ fn test_event_wrapper_builder_with_event_content() {
 
 #[test]
 fn test_try_into_deploy_account_transaction() {
-    fn helper(
+    // This helper methods either returns result of `TryInto::try_into()` and expected result or the
+    // error incase `TryInto::try_into()` fails
+    fn get_try_into_and_expected_value(
         array_size: usize,
         calldata_size: usize,
     ) -> Result<(DeployAccountTransaction, DeployAccountTransaction), BroadcastedTransactionConversionErrorWrapper>
@@ -402,7 +404,7 @@ fn test_try_into_deploy_account_transaction() {
         let expected_signature = bounded_vec![Felt252Wrapper::default(); array_size];
         let expected_calldata = bounded_vec![Felt252Wrapper::default(); calldata_size];
 
-        let expect_output = DeployAccountTransaction {
+        let expected_output = DeployAccountTransaction {
             version: 1_u8,
             calldata: expected_calldata,
             signature: expected_signature,
@@ -412,13 +414,13 @@ fn test_try_into_deploy_account_transaction() {
             max_fee: Felt252Wrapper::from(FieldElement::default()),
         };
 
-        Ok((output, expect_output))
+        Ok((output, expected_output))
     }
 
-    let zero_len = helper(0, 0).expect("shouldn't fail");
+    let zero_len = get_try_into_and_expected_value(0, 0).expect("failed to bound signature or calldata");
     pretty_assertions::assert_eq!(zero_len.0, zero_len.1);
 
-    let one_len = helper(1, 1).expect("shouldn't fail");
+    let one_len = get_try_into_and_expected_value(1, 1).expect("failed to bound signature or calldata");
     pretty_assertions::assert_eq!(one_len.0, one_len.1);
 
     // Cannot run this tests because currently `MaxArraySize` and `MaxCalldataSize` is very large
--- a/crates/primitives/starknet/src/tests/transaction.rs
+++ b/crates/primitives/starknet/src/tests/transaction.rs
@@ -380,7 +380,7 @@ fn test_event_wrapper_builder_with_event_content() {
 #[test]
 fn test_try_into_deploy_account_transaction() {
     // This helper methods either returns result of `TryInto::try_into()` and expected result or the
-    // error incase `TryInto::try_into()` fails
+    // error in case `TryInto::try_into()` fails
     fn get_try_into_and_expected_value(
         array_size: usize,
         calldata_size: usize,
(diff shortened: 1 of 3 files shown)


crates/primitives/starknet/src/tests/transaction.rs
```diff
@@ -373,3 +376,72 @@ fn test_event_wrapper_builder_with_event_content() {
 
     pretty_assertions::assert_eq!(event_wrapper, expected_event);
 }
+
+#[test]
+fn test_try_into_deploy_account_transaction() {
+    fn helper(
```
EvolveArt, line 382: define `helper` out of the test please

EvolveArt: and give it an explicit name

crates/primitives/starknet/src/tests/transaction.rs
```diff
@@ -373,3 +376,72 @@ fn test_event_wrapper_builder_with_event_content() {
...
+            version: 1_u8,
+            calldata: expected_calldata,
+            signature: expected_signature,
+            nonce: Felt252Wrapper::from(FieldElement::default()),
+            salt: Felt252Wrapper::from(FieldElement::default()),
+            account_class_hash: Felt252Wrapper::from(FieldElement::default()),
+            max_fee: Felt252Wrapper::from(FieldElement::default()),
+        };
+
+        Ok((output, expect_output))
```
EvolveArt, line 415: `expected_output`

crates/primitives/starknet/src/tests/transaction.rs
```diff
@@ -373,3 +376,72 @@ fn test_event_wrapper_builder_with_event_content() {
...
+        let expected_signature = bounded_vec![Felt252Wrapper::default(); array_size];
+        let expected_calldata = bounded_vec![Felt252Wrapper::default(); calldata_size];
+
+        let expect_output = DeployAccountTransaction {
+            version: 1_u8,
+            calldata: expected_calldata,
+            signature: expected_signature,
+            nonce: Felt252Wrapper::from(FieldElement::default()),
+            salt: Felt252Wrapper::from(FieldElement::default()),
+            account_class_hash: Felt252Wrapper::from(FieldElement::default()),
```
EvolveArt, line 411: can't we just do `FieldElement::default().into()` ?

crates/primitives/starknet/src/tests/transaction.rs
```diff
@@ -373,3 +376,72 @@ fn test_event_wrapper_builder_with_event_content() {
...
+    }
+
+    let zero_len = helper(0, 0).expect("shouldn't fail");
+    pretty_assertions::assert_eq!(zero_len.0, zero_len.1);
+
+    let one_len = helper(1, 1).expect("shouldn't fail");
+    pretty_assertions::assert_eq!(one_len.0, one_len.1);
+
+    // Cannot run this tests because currently `MaxArraySize` and `MaxCalldataSize` is very large
+    // so its not possible to make allocation for such large `Vec` and `BoundedVec`
```
EvolveArt, line 425: You can mock the max values to do these tests.

crates/primitives/starknet/src/tests/transaction.rs
```diff
@@ -373,3 +376,72 @@ fn test_event_wrapper_builder_with_event_content() {
...
+            nonce: Felt252Wrapper::from(FieldElement::default()),
+            salt: Felt252Wrapper::from(FieldElement::default()),
+            account_class_hash: Felt252Wrapper::from(FieldElement::default()),
+            max_fee: Felt252Wrapper::from(FieldElement::default()),
+        };
+
+        Ok((output, expect_output))
+    }
+
+    let zero_len = helper(0, 0).expect("shouldn't fail");
```
EvolveArt, line 418: more explicit error please

EvolveArt (changes requested):
left few comments

crates/primitives/starknet/src/tests/transaction.rs
lambda-0x: that works too, i just like to be more explicit about the type being used. i can change it if you want.

crates/primitives/starknet/src/tests/transaction.rs
lambda-0x: can you explain how can I do that, because this max values are set inside [`DeployAccountTransaction`](https://github.com/keep-starknet-strange/madara/blob/f887287a1ae8b3fe1be00ef39358bab3040f42f0/crates/primitives/starknet/src/transaction/types.rs#L352-L367) and its not generic over that value.

crates/primitives/starknet/src/tests/transaction.rs
lambda-0x: this helper is very specific to this test and I don't think it will be helpful anywhere else. will give it more explicit name though.

crates/primitives/starknet/src/tests/transaction.rs
EvolveArt: yes please

crates/primitives/starknet/src/tests/transaction.rs
EvolveArt: we never know, I prefer to keep it outside even in terms of readability :)

crates/primitives/starknet/src/tests/transaction.rs
EvolveArt: You can find the test I mentioned on TG [here](https://github.com/keep-starknet-strange/madara/blob/main/crates/primitives/starknet/src/tests/transaction.rs#L269).
Indeed here it works because `MaxArraySize` is 10001 and wouldn't work with `u32::MAX`. So we need to find some workaround to mock these `Max` values within tests.

lambda-0x: will use [this](https://doc.rust-lang.org/reference/conditional-compilation.html#test) to change max values while running tests

EvolveArt (approved):
lgtm

PR #166, opened 2023-04-05
feat: global state root computation

ben2077:
…e root, and global state root.

# Pull Request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Testing
- [ ] Other (please describe):

# What is the current behavior?

Issue Number: N/A

# What is the new behavior?

-
-
-

# Does this introduce a breaking change?

- [ ] Yes
- [ ] No

# Other information

PR #667, opened 2023-06-16
feat: global state root 🔥

EvolveArt:
# Pull Request type

- Bugfix
X Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #166 

## What is the new behavior?

- Adds `SCALE` encoding for the `StateCommitmentTree` structure
- Adds `State` struct in runtime which contains the two `StateCommitmentTree` : classes/contracts trees.
- Upgrades a few dependencies to latest version `scale-info`, `scale-coded` and `bitvec`
- Modify `BlockifierStateAdapter` such that it updates the state trees accordingly.
- Implements `IntermediateStateRoot` to compute the global state root from the intermediate trees.

Yes

## Other information

- A few things still need to be investigated : 
1. Is it fine to store these trees within the runtime ? Discussed with Lucas and spent a few hours thinking about # approaches but it seems unavoidable.
2. When should the tree updates happen ? Incrementally like in this implementation or all at the same time, If we take into account the Storage Overlay (caching layer of substrate) then reading through all the state and building the trees shouldn't be that expensive ? 
3. How can we `SCALE` encode `RefCell` ? Does it even make sense or should we replace by `Box` ?
4. `BitVec` has no `MaxEncodedLen` which makes sense but prevents from storing it within the runtime, so we can either create some `BoundedBitVec` wrapper or switch to `BoundedVec<bool>` for instance.

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -1162,6 +1162,7 @@ impl<T: Config> Pallet<T> {
 
         pending_storage_changes.for_each(|(contract_address, storage_diffs)| {
             // Retrieve state trie for this contract.
+            // TODO: Investigate what to do in case of failure of the state root computation
             let mut state_tree = ContractTries::<T>::get(contract_address).unwrap_or_default();
             // For each smart contract, iterate through storage diffs and update the state trie.
             storage_diffs.into_iter().for_each(|(storage_key, storage_value)| {


crates/pallets/starknet/src/lib.rs
```diff
@@ -207,6 +208,13 @@ pub mod pallet {
     pub(super) type Pending<T: Config> =
         StorageValue<_, BoundedVec<(Transaction, TransactionReceiptWrapper), MaxTransactions>, ValueQuery>;
 
+    /// The Starknet pallet storage items.
+    /// STORAGE
+    /// Current building block's transactions.
+    #[pallet::storage]
+    #[pallet::getter(fn state)]
+    pub(super) type State<T: Config> = StorageValue<_, StateCommitments, ValueQuery>;
```
tdelabro, line 216: Not sure we should store anything in the storage indeed. 
It will be present in the block header digest. I see no reason to double store it.

If we compute it as we go along we will have to store it for sure, but we can kill the storage in on_finalize.

If we compute it in one go we can do everything in memory.

tdelabro:
Does computing it in a single go produces the exact same result that tax by tx?

I can't think of any drawbacks and see a lot of pros as it allows us to skip storing all the intermediate states.

tdelabro:
Ideally (in a few months) this should be computed outside of the runtime, and maybe for block b-1.

So I would advise for the solution that is as decoupled as possible from substrate logic.

tdelabro:
You can disable the scale max length constraint. Either for one single storage or for the whole pallet.

It should not cause any trouble, especially if we kill the storage at the end of the block execution

tdelabro:
The problem was we were reading everything every tx. Now we read only what we need, but at each tx.

If we do this at the end in onfinalize rather than along the execution, it will be exactly the same amount of computing. Won't it ?
And we will spare the burden of storing/loading/making compatible the each transitive state.

EvolveArt:
Following some discussion with @tdelabro and Mikro from Equilibrium who wrote the initial MPT implementation, I decided to replace `RefCell` by `Box` within the MPT implementation. Thus, it is `SCALE` encodable!

Second thing, I had to do a PR to [scale-info](https://github.com/paritytech/scale-info) repo in order to add a `TypeInfo` implementation for `Rc`, while the [PR](https://github.com/paritytech/scale-info/pull/185) is not merged we can use my fork.

With these two things in place the trees are now persisted through time and we can update them iteratively, this sounds like the best approach until we adopt a client-first approach in the next few months.

EvolveArt:
Relaying this to serve as a reminder

> Computing the root state in the client will only lead to perf gain if it is done for the block b-1, where it can be paralleled. If done for block b it provides no true gain and may even be slower.
Matthias will now do little (big) tweak to the way it's tree is represented in memory so it doesn't use Rc at all. As soon as it is done we will have a state root computation done in the runtime. 
It will be a great starting point to explore alternative ways to do it and benchmark to see if they truly bring some perf improvements.

crates/client/block-proposer/src/lib.rs
```diff
@@ -217,11 +217,11 @@ where
...
             None,
             Box::pin(async move {
-                // Leave some time for evaluation and block finalization (20%)
-                // and some time for block production (80%).
+                // Leave some time for evaluation and block finalization (33%)
+                // and some time for block production (66%).
                 // We need to benchmark and tune this value.
                 // Open question: should we make this configurable?
-                let deadline = (self.now)() + max_duration - max_duration / 5;
+                let deadline = (self.now)() + max_duration - max_duration / 3;
```
abdelhamidbakhta, line 224: get back to 20% as discussed please

abdelhamidbakhta (dismissed):
left comment

crates/client/block-proposer/src/lib.rs
EvolveArt: done

crates/pallets/starknet/src/blockifier_state_adapter.rs
```diff
@@ -92,17 +94,53 @@ impl<T: Config> State for BlockifierStateAdapter<T> {
...
+        if T::EnableStateRoot::get() {
+            // Update contracts trie
+            let mut tree = crate::State::<T>::get().storage_commitment;
+            let nonce = Pallet::<T>::nonce(contract_address);
+            let class_hash = Pallet::<T>::contract_class_hash_by_address(contract_address).unwrap_or_default();
+
+            // Update storage trie
+            let mut state_tree = crate::State::<T>::get().state_commitment;
+            state_tree.set(key, value);
+            let state_root = state_tree.commit();
```
LucasLvy, line 113: We should only commit in the finalize block. Here everytime we write in the storage or do something we commit which is extremely long. same for every function in the blockifier adapter

LucasLvy (changes requested):

crates/pallets/starknet/src/blockifier_state_adapter.rs
tdelabro: But do we really need to update the leaf at each transaction?
Can't we just push values in the tree at each tx and only commit at the very end?

EvolveArt: That's what we do, apart from the state root which is necessary to compute the leaf hash for contracts tree

LucasLvy: hmm no we update every leaf of the contract tree during the block execution and then when we close a block we recompute the whole starknet tree to update the state root. We should compute the tree only once per block

tdelabro (changes requested):

crates/pallets/starknet/src/state_root.rs
```diff
@@ -2,20 +2,60 @@
...
-    /// TODO: Implement this function.
     /// # Returns
-    /// * `U256` - The intermediate state root.
+    /// * `Felt252Wrapper` - The intermediate state root.
     fn get() -> Felt252Wrapper {
-        FieldElement::ONE.into()
+        // If state root is disabled, return one.
+        if !T::EnableStateRoot::get() {
+            return Felt252Wrapper::ONE;
+        }
```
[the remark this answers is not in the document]
EvolveArt: it is, should we just put 0 ? ideally we'd even want to remove it from the header

crates/pallets/starknet/src/state_root.rs
```diff
@@ -2,20 +2,60 @@
...
 impl<T: Config> Get<Felt252Wrapper> for IntermediateStateRoot<T> {
     /// Compute the state root of Starknet and return it.
-    /// For now, we just return a dummy state root.
-    /// TODO: Implement this function.
     /// # Returns
-    /// * `U256` - The intermediate state root.
+    /// * `Felt252Wrapper` - The intermediate state root.
     fn get() -> Felt252Wrapper {
-        FieldElement::ONE.into()
+        // If state root is disabled, return one.
```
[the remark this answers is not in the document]
EvolveArt: We need to access the runtime storage to compute the root, so should I just add it as an internal function in the pallet directly ?

My understanding was that this approach helped with readbility/maintainability and it works so not sure about the downsides ?

crates/pallets/starknet/src/tests/mock.rs
```diff
@@ -76,6 +76,7 @@ parameter_types! {
     pub const TransactionLongevity: u64 = u64::MAX;
     pub const InvokeTxMaxNSteps: u32 = 1_000_000;
     pub const ValidateMaxNSteps: u32 = 1_000_000;
+    pub const EnableStateRoot: bool = false;
```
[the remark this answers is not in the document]
EvolveArt: You mean the part where it's enabled ?

Working on it, did some tests manually and it works. Still need a few things to complete it and have a full-blown test suite. 

Next PRs will be the following : 

- Multiple runtime implementations' testing suite
- Optimize and add test coverage to runtime MPT implementation
- Fix/Test Poseidon for all trees/hashes
- Add starknet_getProof

crates/pallets/starknet/src/state_root.rs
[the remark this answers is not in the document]
tdelabro: Downside are:
- premature abstraction. There is no alternative implementation of the computation of the root. So no need to make it configurable.
- added complexity. You use one extra feature of substrate when not needed
- bad for readability. When you are in the pallet and see `StateRoot::Get`, you have to go the runtime to know what is the code being actually executed
- hidden side effetcs. The method is named `get` but it actually write to numerous storages of the pallet.

Just write it as a function called `compute_and_store_state_root` or something like this.

crates/pallets/starknet/src/tests/mock.rs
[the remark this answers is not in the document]
tdelabro: Okay. Yes, I would like to see another mock.rs file where the StateRoot computation is enabled and a set of test to ensure it work.

Do you think you can ship it in this PR? I don't really like the idea of merging it first, and then right test. They should come together. But if it is really a problem we can do an exception

crates/pallets/starknet/src/state_root.rs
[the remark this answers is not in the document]
tdelabro: I would either return an option or, better move the check outside of the function call.
First you check if you EnableStateRoot is true, if yes you call the function that is meant to compute it.

crates/pallets/starknet/src/tests/mock.rs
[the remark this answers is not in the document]
EvolveArt: Adding the test setup + basic tests then

crates/pallets/starknet/src/lib.rs
```diff
@@ -1122,4 +1123,52 @@ impl<T: Config> Pallet<T> {
...
+    /// This function is called at the end of each block.
+    /// It iterates through all the pending storage changes and updates the storage trie.
+    /// It then computes the state root and stores it in the runtime storage.
+    ///
+    /// # Returns
+    ///
+    /// The global state root.
+    pub fn compute_and_store_state_root() -> Felt252Wrapper {
+        // Update contracts trie
+        let mut commitments = State::<T>::get();
```
tdelabro, line 1137: There is a getter method, `Self::state`

crates/pallets/starknet/src/lib.rs
```diff
@@ -1122,4 +1123,52 @@ impl<T: Config> Pallet<T> {
...
+
+            let nonce = Nonces::<T>::get(contract_address);
+            let class_hash = ContractClassHashes::<T>::get(contract_address).unwrap_or_default();
+            let hash = calculate_contract_state_hash::<T::SystemHash>(class_hash, state_root, nonce);
+            commitments.storage_commitment.set(contract_address, hash);
+
+            // Finally update the contracts trie in runtime storage.
+            State::<T>::mutate(|state| {
+                state.storage_commitment = commitments.clone().storage_commitment;
+            });
```
tdelabro, line 1163: I don't think you need to mutate it at each iteration of the loop.
It is only read once, before the loop. And updated in `commitments.storage_commitment.set(contract_address, hash);` at each iteration of the loop.

You can instead keep updating the state of `storage_commitment` in you in-memory `commitments` variable, and only write it to `State` once every pending update is processed, after the loop

crates/pallets/starknet/src/lib.rs
```diff
@@ -1122,4 +1123,52 @@ impl<T: Config> Pallet<T> {
...
+
+            PendingState::<T>::remove(contract_address);
```
tdelabro, line 1165: Same. Once you finished iterating over all of the pending states, you can just kill the storage itself, just bellow your loop. Rather than removing each entry 1 by 1

tdelabro (changes requested):

crates/pallets/starknet/src/lib.rs
EvolveArt: Is that more efficient, or substrate does it under the hood ?

crates/pallets/starknet/src/lib.rs
EvolveArt: we can't kill the storage cause its a `StorageMap` :(

LucasLvy: Use this ? `PendingState::<T>::remove_all(None);`

tdelabro: Does order matter to you?
Otherwise you can use drain and directly iterate on it.

`remove_all` is deprecated, you should use [`clear`](https://paritytech.github.io/substrate/master/frame_support/storage/types/struct.StorageMap.html#method.clear) instead, but you have to check the result and eventually call it again depending of the limit you used

crates/pallets/starknet/src/lib.rs
tdelabro: I believe it will be more efficient. 
Right now you right X time without ever reading between. So you just really need to write once, the last time

crates/primitives/starknet/src/crypto/commitment/mod.rs
```diff
@@ -39,15 +44,56 @@ impl<T: CryptoHasherT> CommitmentTree<T> {
...
+        let key = &index.0.to_bytes_be()[..31];
+        self.tree.set(&BitVec::from_vec(key.to_vec()), value)
+    }
+
+    /// Get the merkle root of the tree.
+    pub fn commit(&mut self) -> Felt252Wrapper {
+        self.tree.commit()
+    }
+
+    #[allow(dead_code)]
```
LucasLvy, line 90: why is it needed ?

crates/primitives/starknet/src/execution/contract_class_wrapper.rs
```diff
@@ -57,6 +58,11 @@ pub struct ContractClassWrapper {
 impl ContractClassWrapper {
     // This is the maximum size of a contract in starknet. https://docs.starknet.io/documentation/starknet_versions/limits_and_triggers/
     const MAX_CONTRACT_BYTE_SIZE: usize = 20971520;
+
+    /// Computes the hash of the contract class.
+    pub fn class_hash(&self) -> Felt252Wrapper {
+        Felt252Wrapper::ONE
```
LucasLvy, line 64: Is this intentional?

crates/primitives/starknet/src/tests/crypto.rs
```diff
@@ -121,10 +120,40 @@ fn test_merkle_tree() {
...
+    );
+}
+
+#[test]
+fn test_merkle_tree_class_commitment() {
+    let class_hashes = vec![Felt252Wrapper::from(0_u128), Felt252Wrapper::from(1_u128)];
+
+    let class_com = calculate_class_commitment_tree_root_hash::<PedersenHasher>(&class_hashes);
+
+    // The values we test ours against are computed from the sequencer test.
```
LucasLvy, line 138: is this true ?

crates/primitives/starknet/src/tests/crypto.rs
```diff
@@ -121,10 +120,40 @@ fn test_merkle_tree() {
...
+}
+
+#[test]
+fn test_merkle_tree_poseidon() {
+    let class_hashes = vec![Felt252Wrapper::from(0_u128), Felt252Wrapper::from(1_u128)];
+
+    let _class_com = calculate_class_commitment_tree_root_hash::<PoseidonHasher>(&class_hashes);
+
+    // The values we test ours against are computed from the sequencer test.
+    // assert_eq!(
```
LucasLvy, line 152: uncomment ?

tests/tests/test-rpc/test-starknet-rpc.ts
```diff
@@ -531,10 +531,10 @@ describeDevMadara("Starknet RPC", (context) => {
 
       const latestBlock = await providerRPC.getBlockHashAndNumber();
 
+      // TODO: Add real values
```
LucasLvy, line 534: add them

tests/tests/test-rpc/test-starknet-rpc.ts
```diff
@@ -563,10 +563,10 @@ describeDevMadara("Starknet RPC", (context) => {
         anteriorBlock.block_hash
       );
 
+      // TODO: Add real values
```
LucasLvy, line 566: add them

LucasLvy (changes requested):

crates/pallets/starknet/src/lib.rs
```diff
@@ -209,6 +211,31 @@ pub mod pallet {
     pub(super) type Pending<T: Config> =
         StorageValue<_, BoundedVec<(Transaction, TransactionReceiptWrapper), MaxTransactions>, ValueQuery>;
 
+    /// The Starknet pallet storage items.
+    /// STORAGE
+    /// State commitments of the current block.
+    #[pallet::storage]
+    #[pallet::unbounded]
+    #[pallet::getter(fn current_state_commitments)]
+    pub(super) type CurrentStateCommitments<T: Config> = StorageValue<_, StateCommitments, ValueQuery>;
```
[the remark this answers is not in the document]
EvolveArt: hm we discussed with Tim and this is more descriptive maybe `StarknetCurrentStateCommitments` ?

crates/pallets/starknet/src/lib.rs
```diff
@@ -339,7 +378,21 @@ pub mod pallet {
 
             for (key, value) in self.storage.iter() {
                 StorageView::<T>::insert(key, value);
+
+                // Update state tries if enabled in the runtime configuration
+                if T::EnableStateRoot::get() {
```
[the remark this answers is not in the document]
EvolveArt: why ? Genesis contracts/storage should modify the genesis state root no?

crates/pallets/starknet/src/lib.rs
```diff
@@ -1091,4 +1147,50 @@ impl<T: Config> Pallet<T> {
...
+
+            // Update the state trie for this contract in runtime storage.
+            StorageTries::<T>::set(contract_address, Some(state_tree.clone()));
+
+            // We then compute the state root
+            // And update the storage trie
+            let state_root = state_tree.commit();
+
+            let nonce = Self::nonce(contract_address);
+            let class_hash = Self::contract_class_hash_by_address(contract_address).unwrap_or_default();
```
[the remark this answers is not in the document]
EvolveArt: not really sure as this should not happen anyway, what do we do if compute state root fails do we just keep the previous one ? Set it to zero ?

crates/pallets/starknet/src/tests/state_root.rs
```diff
@@ -0,0 +1,28 @@
...
+use super::mock::*;
+use crate::tests::mock::state_root_mock::MockStateRootRuntime;
+
+#[test]
+fn given_default_runtime_with_state_root_disabled_default_value_is_correct() {
+    new_test_ext().execute_with(|| {
+        basic_test_setup_state_root::<MockRuntime>(2);
+
+        // Check the default state root value
+        pretty_assertions::assert_eq!(Starknet::compute_and_store_state_root(), Felt252Wrapper::default());
```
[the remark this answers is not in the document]
EvolveArt: It's just what we set in the block header when there is no state root computation

crates/primitives/starknet/src/crypto/commitment/mod.rs
EvolveArt: starknet_getProof not implemented yet

crates/primitives/starknet/src/tests/crypto.rs
EvolveArt: Actually wanted to ask what did you mean when writing this ?

tests/tests/test-rpc/test-starknet-rpc.ts
EvolveArt: in another PR that's not related to mine just added the TODO

EvolveArt (commented):
.

crates/pallets/starknet/src/lib.rs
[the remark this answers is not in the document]
LucasLvy: I don’t think that specifying current is relevant as when you’re querying you need to pass a block hash/number it’s quite explicit. WDYT @tdelabro

crates/pallets/starknet/src/lib.rs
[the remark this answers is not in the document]
LucasLvy: Yes right

crates/pallets/starknet/src/lib.rs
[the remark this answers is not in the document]
LucasLvy: I would tend to say that we should stop producing blocks as there is a big bug in the chain but maybe it’s too radical

crates/primitives/starknet/src/tests/crypto.rs
LucasLvy: I went in the Cairo lang repo and took the values from the python tests. I don’t think those tests are external though and I don’t think we test specifically the state root

LucasLvy (approved):

Issue #643, opened 2023-06-14
dev: KISS/YAGNI for Hasher traits 🔨


EvolveArt:
Right now we have two traits `CryptoHasherT` and `HasherT` both used at different places and which do not reuse correctly `starknet-rs`, cf this [function](https://github.com/xJonathanLEI/starknet-rs/blob/009bac782bd6d908ca5b1ede9938b38dcbd2fb31/starknet-core/src/crypto.rs#L25).

We should unify those implementations and keep it simple.

ApereLieZ:
I'm ready to address this issue

exp-table:
soon 😏

PR #672, opened 2023-06-16
chore: unifying Hasher traits

exp-table:
# Pull Request type

- Refactoring (API changes)

We currently have two traits, one for hashing `&[u8]/Felt252Wrapper` types and the other one for `FieldElement`.

Resolves: #643 

## What is the new behavior?

- Merged `HasherT` and `CryptoHasherT` into `HasherT`
- Functions naming changed for clarity

Yes, the trait's functions have been renamed for more clarity. All changes have been made in the codebase to reflect it.

## Other information

1. Given `starknet-rs` does not support `no_std` yet, its `compute_hash_on_elements` function has been copied into `PedersenHasher` with a mention to its source.


crates/primitives/starknet/src/crypto/hash/mod.rs
```diff
@@ -20,19 +22,27 @@ impl ThreadSafeCopy for Hasher {}
...
-            Self::Pedersen(p) => p.hash(data),
-            Self::Poseidon(p) => p.hash(data),
+            Self::Pedersen(p) => p.hash_bytes(data),
+            Self::Poseidon(p) => p.hash_bytes(data),
         }
     }
 
-    fn hash_elements(&self, data: &[Felt252Wrapper]) -> Felt252Wrapper {
+    fn compute_hash_on_wrappers(&self, data: &[Felt252Wrapper]) -> Felt252Wrapper {
         match self {
```
EvolveArt, line 33: cant't we just do `self. compute_hash_on_wrappers` tho ?

crates/primitives/starknet/src/crypto/hash/pedersen.rs
```diff
@@ -32,33 +34,17 @@ impl HasherT for PedersenHasher {
...
-/// The pedersen CryptoHasher implementation.
-impl CryptoHasherT for PedersenHasher {
     #[inline(always)]
-    fn hash(a: FieldElement, b: FieldElement) -> FieldElement {
+    fn hash_elements(a: FieldElement, b: FieldElement) -> FieldElement {
         pedersen_hash(&a, &b)
     }
 
-    /// Compute hash on elements, base on the [python implementation](https://github.com/starkware-libs/cairo-lang/blob/12ca9e91bbdc8a423c63280949c7e34382792067/src/starkware/cairo/common/hash_state.py#L6-L15).
+    /// Compute hash on elements, taken from [starknet-rs](https://github.com/xJonathanLEI/starknet-rs/blob/master/starknet-core/src/crypto.rs#L25) pending a no_std support.
```
EvolveArt, line 47: why not use it directly then ?

crates/primitives/starknet/src/traits/hash.rs
```diff
@@ -11,24 +11,15 @@ pub trait HasherT {
...
     /// The hash of the data.
-    fn hash_elements(&self, data: &[Felt252Wrapper]) -> Felt252Wrapper;
-}
-
-/// A trait for default hashing instance.
-pub trait DefaultHasher {
-    /// Get Hasher default instance.
-    fn hasher() -> Self;
-}
+    fn compute_hash_on_wrappers(&self, data: &[Felt252Wrapper]) -> Felt252Wrapper;
```
EvolveArt, line 21: I feel like `compute_hash_on_wrappers` is redundant since it's the same underlying structure. Let's keep it simple here and only keep the wrapper version.

EvolveArt (changes requested):
I know it's still in progress but left a few comments there

crates/primitives/starknet/src/crypto/hash/pedersen.rs
exp-table: `starknet_core` only works with `std`, we could have conditional compilation and do

```
#[cfg(feature = "std")]
use starknet_core::crypto::compute_hash_on_elements;
```

But I think it will require to have two diff. impls. (std and no_std) for functions where we want to use `compute_hash_on_elements`; std would use it and then no_std a "custom" impl.

Somewhat new to Rust so not sure ..

crates/primitives/starknet/src/crypto/hash/mod.rs
exp-table: Isn't this not possible because of the `Hasher` variants' field is what we need to actually match against?

crates/primitives/starknet/src/traits/hash.rs
exp-table: I somewhat need guidance on that ; it's true `compute_hash_on_wrappers` and `compute_hash_on_elements` essentially do the same but the latter doesn't need to be passed `&self`into the parameters.

We could remove it and just use the former but I think it would require refactoring [primitives/starknet/crypto/commitment](https://github.com/keep-starknet-strange/madara/blob/main/crates/primitives/starknet/src/crypto/commitment/mod.rs) and everywhere its functions are used. Basically, if using `compute_hash_on_wrappers`, would need an instance of a `HasherT` to be passed into every function where the compute is used.

crates/primitives/starknet/src/crypto/hash/pedersen.rs
EvolveArt: Hm I think starknet-rs is no_std compatible since we do use it everywhere in practice

crates/primitives/starknet/src/traits/hash.rs
EvolveArt: I think this would make a lot of sense, anyway I changed every `FieldElement` to `Felt252Wrapper` in the global state root [PR](https://github.com/keep-starknet-strange/madara/pull/667) so we can do it. It just makes sense to unify everything and since the repo clearly tends to use the wrapper everywhere, we should also do it here imho.

crates/primitives/starknet/src/crypto/hash/pedersen.rs
exp-table: Not all the crates! There is actually an issue to make `starknet-core` compatible with no_std [here](https://github.com/xJonathanLEI/starknet-rs/issues/403), @dbejarano820 is working on it 🙏

crates/primitives/starknet/src/traits/hash.rs
exp-table: What if we replace `compute_hash_on_elements` by a similarly named method that uses `Felt252Wrapper` but unlike `compute_hash_on_wrappers` does not need to be passed `&self`? In some places, it just seems to make more sense that the generic parameter `T: HasherT` is not instantiated and we just directly call the appropriate function.

EvolveArt: Sounds like a great idea, please do :) Overall I think we should focus on making starknet-rs core no_std compatible would also reduce the complexity of this PR and just make things more straightforward.

dbejarano820: Yup, I'll be tackling the starknet-core `no-std` support this week

exp-table:
There is no guarantee that `no_std` support for `starknet-core` is coming soon, although trying my best at it. I think it's reasonable to move forward and maybe merge?

EvolveArt (changes requested):
Can we implement the todo! ?

crates/primitives/starknet/src/crypto/hash/mod.rs
```diff
@@ -20,19 +22,33 @@ impl ThreadSafeCopy for Hasher {}
...
-            Self::Pedersen(p) => p.hash_elements(data),
-            Self::Poseidon(p) => p.hash_elements(data),
+            Self::Pedersen(p) => p.compute_hash_on_wrappers(data),
+            Self::Poseidon(p) => p.compute_hash_on_wrappers(data),
         }
     }
+
+    fn hash_elements(_a: FieldElement, _b: FieldElement) -> FieldElement {
+        todo!()
```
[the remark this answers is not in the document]
exp-table: Done, but this comes with some downsides I think?

To be able to match `self` (otherwise I don't think there is a way to distinguish which variant of the enum it is), need to change the parameters of `hash_elements` and `compute_hash_on_elements` to accept `&self`.

Now, in quite a few places we actually used `<T as HasherT>::compute_hash_on_elements` but now that we need `self`, it's changed to `<T>::default().compute_hash_on_elements()`. Now, this is where it could become a DX small bump because while it's fine if we use `PedersenHasher` or `PoseidonHasher`, we need to make a choice for `Hasher` (which also implements `HasherT`) because it's ambiguous and its default is `Hasher::Pedersen(pedersen::PedersenHasher::default())`.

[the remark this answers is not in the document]
EvolveArt: If we really want to KISS can't we just not implement the HasherT trait on Hasher and just on Pedersen and Poseidon ? Would be cleaner no?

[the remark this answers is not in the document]
EvolveArt: As you mentionned the current state is way too ambiguous and the use of `default` should be prohibited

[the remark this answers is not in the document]
exp-table: `Default` impl. for `Hasher` has been made `unreachable` to preserve that feature for `Pedersen` and `Poseidon`.

EvolveArt (approved):
lgtm

Issue #588, opened 2023-06-08
feat: replace hardcoded sequencer address with inherent data submission


lana-shanghai:
# Feature Request

Right now we have a [hardcoded value](https://github.com/keep-starknet-strange/madara/blob/main/crates/pallets/starknet/src/lib.rs#L105) for the block header's [sequencer address](https://github.com/keep-starknet-strange/madara/blob/main/crates/primitives/starknet/src/block/header.rs#L33). The idea is to investigate `pallet-authorship` and inherent transactions to find a good way to include the StarkNet sequencer address of the block author in the Substrate block. 

# Acceptance criteria 

We have a reliable way to replace the hardcoded value.

lana-shanghai:
WIP branch https://github.com/lana-shanghai/madara/commits/li/feat/remove_hardcoded_sequencer_addr 

Command to set sequencer address in the local offchain storage:

`curl -sS -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "offchain_localStorageSet", "params": {"kind": "PERSISTENT", "key": "0x737461726b6e65743a3a7365715f61646472", "value": "0x0000000000000000000000000000000000000000000000000000000000000020"}}' http://localhost:9944/`

The address is picked up by the client which propagates an inherent transactions to the pallet storage.

PR #679, opened 2023-06-19
Li/feat/remove hardcoded sequencer addr

lana-shanghai:
### Pull Request type

Feature

### What is the current behavior?

Currently the sequencer address is a hardcoded const. 
Resolves: [issue 588](https://github.com/keep-starknet-strange/madara/issues/588)

### What is the new behavior?

Now the node can set it's StarkNet sequencer address in it's local offchain storage (for the key `b"starknet::seq_addr")` either in Polkadot JS Apps or via terminal with the command `curl -sS -H "Content-Type: application/json" -d '{"id":1, "jsonrpc":"2.0", "method": "offchain_localStorageSet", "params": {"kind": "PERSISTENT", "key": "0x737461726b6e65743a3a7365715f61646472", "value": "0x_your_sequencer_address_32_bytes_long"}}' http://localhost:9944/`
Every block a mandatory dispatch inherent tx checks the local storage of an authority node and send the inherent tx to set the address in the pallet's storage. If not set, it sets it to the default const value. 

### Does this introduce a breaking change?

No, but in terms of tests, the sequencer address value must be set beforehand. The node, however, does not have to set it, but it's recommended.


crates/pallets/starknet/src/lib.rs
```diff
@@ -190,6 +189,15 @@ pub mod pallet {
...
 
+            let addr = StorageValueRef::persistent(b"starknet::seq_addr");
+
+            if let Ok(Some(address)) = addr.get::<[u8; 32]>() {
+                log::info!("Current set sequencer address: {:?}", address);
+                ()
+            } else {
+                log::info!("Sequencer address not set");
+            }
+
```
tdelabro, line 200: What is the purpose of this? You are reading but not doing anything

crates/pallets/starknet/src/lib.rs
```diff
@@ -875,9 +932,7 @@ impl<T: Config> Pallet<T> {
         let pending = Self::pending();
 
         let global_state_root = Felt252Wrapper::ZERO;
-        // TODO: use the real sequencer address (our own address)
-        // FIXME #243
-        let sequencer_address = SEQUENCER_ADDRESS;
+        let sequencer_address = SequencerAddress::<T>::get();
```
tdelabro, line 935: you can use the getter `Self::sequencer_address();` here too

crates/pallets/starknet/src/tests/declare_tx.rs
```diff
@@ -31,11 +33,12 @@ fn given_contract_declare_tx_works_once_not_twice() {
...
             signature: bounded_vec!(),
         };
-
+        assert!(true == true);
         assert_ok!(Starknet::declare(none_origin.clone(), transaction.clone()));
         // TODO: Uncomment once we have ABI support
         // assert_eq!(Starknet::contract_class_by_class_hash(erc20_class_hash), erc20_class);
-        assert_err!(Starknet::declare(none_origin, transaction), Error::<MockRuntime>::ClassHashAlreadyDeclared);
+        // assert_err!(Starknet::declare(none_origin, transaction),
+        // Error::<MockRuntime>::ClassHashAlreadyDeclared);
```
tdelabro, line 41: Did you forget to uncomment or is it intentional?

crates/pallets/starknet/src/tests/declare_tx.rs
```diff
@@ -102,6 +105,7 @@ fn given_contract_declare_tx_fails_wrong_tx_version() {
 #[test]
 fn given_contract_declare_on_openzeppelin_account_then_it_works() {
     new_test_ext().execute_with(|| {
+        SequencerAddress::<MockRuntime>::put(DEFAULT_SEQUENCER_ADDRESS);
```
tdelabro, line 108: Could you provide a utility method wich do all of the basic setup things we do in each test?

```
SequencerAddress::<MockRuntime>::put(DEFAULT_SEQUENCER_ADDRESS);
System::set_block_number(0);
run_to_block(2);
``` 

And use it everywhere.

It will help future contributors

tdelabro: Name it, basic_test_setup, or something along the lines

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+impl IsFatalError for InherentError {
+    fn is_fatal_error(&self) -> bool {
+        match self {
+            InherentError::WrongAddressFormat => true,
+        }
+    }
+}
+
+impl InherentError {
+    #[cfg(feature = "std")]
```
tdelabro, line 30: Why only `std`?

It really required you can put it above the whole `impl InherentError` block

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Try to create an instance ouf of the given identifier and data.
+    pub fn try_from(id: &InherentIdentifier, mut data: &[u8]) -> Option<Self> {
```
tdelabro, line 32: Why not using the `TryFrom` trait?

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+        inherent_data.put_data(INHERENT_IDENTIFIER, &self.sequencer_address)
+    }
+
+    async fn try_handle_error(
+        &self,
+        _identifier: &InherentIdentifier,
+        _error: &[u8],
+    ) -> Option<Result<(), sp_inherents::Error>> {
+        None
+    }
```
tdelabro, line 99: Why can't you handle the error like it is done for timestamp?

https://paritytech.github.io/substrate/master/src/sp_timestamp/lib.rs.html#240-248

tdelabro (changes requested):

crates/pallets/starknet/src/lib.rs
lana-shanghai: Fair enough, especially with the idea to provide the address in the future with a cli flag there is no need for the logging, will remove it

crates/primitives/starknet/src/sequencer_address/mod.rs
lana-shanghai: Good question, I will add that block

crates/primitives/starknet/src/sequencer_address/mod.rs
lana-shanghai: Looking into it, will push if it's easily replaceable!

crates/node/Cargo.toml
```diff
@@ -54,7 +55,7 @@ sp-trie = { workspace = true, features = ["default"] }
 # These dependencies are used for the node template's RPCs
 jsonrpsee = { workspace = true, features = ["server"] }
 
-# Substrate primtiives dependencies
+# Substrate primtives dependencies
```
dbejarano820, line 58: typo: `primitives`

lana-shanghai: Hahah, fixed one typo and got another in... XD

crates/pallets/starknet/src/lib.rs
```diff
@@ -271,6 +271,16 @@ pub mod pallet {
...
+    /// Current sequencer address.
+    #[pallet::storage]
+    #[pallet::getter(fn sequencer_address)]
+    pub type SequencerAddress<T: Config> = StorageValue<_, [u8; 32], ValueQuery>;
+
+    /// Ensure the sequencer address was updated for this block.
+    #[pallet::storage]
+    #[pallet::getter(fn seq_addr_update)]
+    pub type SeqAddrUpdate<T: Config> = StorageValue<_, bool, ValueQuery>;
+
```
dbejarano820, line 283: Is this necessary? Can't we use the `SequencerAddress` value itself to ensure it was updated for the block?

tdelabro: I think it allows the SequencerAddress to stay in the storage, to be available for the off-chain worker, even after the `SeqAddrUpdate` is killed.

crates/pallets/starknet/src/lib.rs
```diff
@@ -166,7 +163,10 @@ pub mod pallet {
     #[pallet::hooks]
     impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
         /// The block is being finalized.
-        fn on_finalize(_n: T::BlockNumber) {
+        fn on_finalize(n: T::BlockNumber) {
+            if UniqueSaturatedInto::<u64>::unique_saturated_into(n) >= 1 {
+                assert!(SeqAddrUpdate::<T>::take(), "Sequencer address must be set for the block");
+            }
```
tdelabro, line 169: Rather than doing this check (if >= 1) in every block, when it's only false for the first one, can't we consider setting the storage to the default address in `GenesisBuild::build`.

This way it is set for the genesis block too, and we can remove this otherwise useless branching

crates/pallets/starknet/src/tests/query_tx.rs
```diff
@@ -1,11 +1,14 @@
...
 
 use super::mock::*;
+use crate::SequencerAddress;
 
 #[test]
 fn estimates_tx_fee_successfully() {
     new_test_ext().execute_with(|| {
+        SequencerAddress::<MockRuntime>::put(DEFAULT_SEQUENCER_ADDRESS);
         System::set_block_number(0);
         run_to_block(2);
```
tdelabro, line 13: Can't you use `basic_test_setup` here too

crates/pallets/starknet/src/tests/query_tx.rs
```diff
@@ -21,6 +24,7 @@ fn estimates_tx_fee_successfully() {
 #[test]
 fn estimate_does_not_add_to_pending() {
     new_test_ext().execute_with(|| {
+        SequencerAddress::<MockRuntime>::put(DEFAULT_SEQUENCER_ADDRESS);
         System::set_block_number(0);
         run_to_block(2);
```
tdelabro, line 29: And also here?

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// Returns the sequencer address of this inherent data provider.
+    pub fn sequencer_address(&self) -> InherentType {
+        self.sequencer_address
+    }
+
+    /// Default address if sequencer address is not specified.
+    pub fn from_default() -> Self {
+        InherentDataProvider { sequencer_address: DEFAULT_SEQUENCER_ADDRESS }
+    }
```
tdelabro, line 86: I would rather use the trait Default

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// Convert storage vector into `[u8; 32]`.
+    pub fn from_vec(storage_val: Vec<u8>) -> Self {
+        let addr: [u8; 32] = slice_to_arr(&storage_val);
+        InherentDataProvider { sequencer_address: addr }
+    }
```
tdelabro, line 92: I would rather use the trait `TryFrom<Vec<u8>>`. And let the caller decide if he want to use `DEFAULT_SEQUENCER_ADDRESS` in case of an error

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/// Helper function to convert storage value.
+pub fn slice_to_arr(slice: &[u8]) -> [u8; 32] {
```
tdelabro, line 96: Should this function really be public?

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Submitted address must be `[u8; 32]`.
+    WrongAddressFormat,
+}
+
+#[cfg(feature = "std")]
+impl std::fmt::Display for InherentError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Inherent decoding error")
+    }
+}
```
tdelabro, line 33: You can use `thiserror_no_std::Error` to define impl display for you

```rust
#[derive(Error)]
pub enum InherentError {
    /// Submitted address must be `[u8; 32]`.
    #[error("Inherent decoding error")]
    WrongAddressFormat,
}

File: crates/primitives/starknet/src/sequencer_address/mod.rs [hunk 17]
```diff
@@ -0,0 +1,114 @@
...
+    [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 222, 173];
+
+/// The storage key for the sequencer address value.
+pub const SEQ_ADDR_STORAGE_KEY: &[u8] = b"starknet::seq_addr";
+
+/// The inherent type for the sequencer address.
+pub type InherentType = [u8; 32];
+
+#[derive(Encode, sp_runtime::RuntimeDebug)]
+#[cfg_attr(feature = "std", derive(Decode, Error))]
```
tdelabro, line 21: Because you are using `thiserror_no_std` you can derive Error even when not in `std`.

Event if it is not used, it's always good to have it implemented for an error type.

Same with `Decode`. Event if only `Encode` is used right now, it is not absurd to also derive `Decode`

tdelabro (changes requested):

crates/pallets/starknet/src/lib.rs
lana-shanghai: We can't know if it has been updated or not by just having a value in `SequencerAddress`

crates/primitives/starknet/src/sequencer_address/mod.rs
lana-shanghai: > I would rather use the trait `TryFrom<Vec<u8>>`. And let the caller decide if he want to use `DEFAULT_SEQUENCER_ADDRESS` in case of an error

Not sure I understand this bit, how can the caller decide that here?

tdelabro: your impl use `slice_to_arr` wich use TryFrom but igonre the error and use default value instead.
What I would like is for the error to be bubble up, and for the caller to decide if he want to use default value instead or to bubble the error up too

crates/node/src/service.rs
```diff
@@ -381,15 +387,25 @@ pub fn new_full(config: Configuration, sealing: Option<Sealing>) -> Result<TaskM
...
+                    let slot = sp_consensus_aura::inherents::InherentDataProvider::from_timestamp_and_slot_duration(
+                        *timestamp,
+                        slot_duration,
+                    );
+
+                    let ocw_storage = offchain_storage.clone();
+                    let prefix = &STORAGE_PREFIX;
+                    let key = SEQ_ADDR_STORAGE_KEY;
+                    let sequencer_address = SeqAddrInherentDataProvider::from_vec(
+                        ocw_storage.unwrap().get(prefix, key).unwrap_or(DEFAULT_SEQUENCER_ADDRESS.to_vec()),
```
tdelabro, line 404: `ocw_storage.unwrap()` why is this unwrap safe?

crates/pallets/starknet/src/tests/query_tx.rs
tdelabro: Yo forgot this one

tdelabro (changes requested):
you forgot a few comments

crates/primitives/starknet/src/sequencer_address/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+impl SequencerAddressInherentData for InherentData {
+    fn sequencer_address_inherent_data(&self) -> Result<Option<InherentType>, sp_inherents::Error> {
+        self.get_data(&INHERENT_IDENTIFIER)
+    }
+}
+
+#[derive(Copy, Clone, Decode, Encode, sp_runtime::RuntimeDebug)]
+#[cfg(feature = "std")]
```
tdelabro, line 59: Rather than putting every symbol down this line behind `#[cfg(feature = "std")]` individually, you should create a module, put it behind `#[cfg(feature = "std")]`, and pub use everything in this module if `#[cfg(feature = "std")]`

tdelabro (changes requested):

tdelabro (approved):

Issue #663, opened 2023-06-16
Add check in the rpc that the class isn't already declared


LucasLvy:
# Description

Everytime a declare transaction is sent madara will process it regardless of wether it's already declared or not. This should not be the case and we should check in the RPC if the class is already declared. If it is the rpc should return an error: `Class already declared`.

In order to check that you can use the runtime api like so:
```rs
self
      .overrides
      .for_block_hash(self.client.as_ref(), substrate_block_hash)
      .contract_class_by_class_hash(substrate_block_hash, class_hash.into())
```
If the value is `Some` then you should return an error `Class already declared` else you can proceed with the normal flow.

bajpai244:
I can take this ser 🙌

chirag-bgh:
@bajpai244 are you still working on this?

bajpai244:
Hello @chirag-bgh, I was travelling back from a conference, so was away from keyboard.

I will attempt this today :)

ClementWalter:
@LucasLvy we talked about this and the current behavior on starknet mainnet/testnet is just to process the tx as it if was the first time. If we definitely can check if it is already declared to avoid useless compute, I think that for now we need to stick to the ecosystem de facto standard, ie responding with a success and the expected object

bajpai244:
@ClementWalter, as discussed, according to the Starknet spec, there should be an error emitted by the sequencer.

Attaching the [link](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L146) here for people following this issue.

bajpai244:
Hello everyone, so update on the task and the current blocker.

**Update:** the logic is complete, I have a test, and we return with the correct error code.

**Blocker:** 

Since, the class hash **"0x372ee6669dc86563007245ed7343d5180b96221ce28f44408cff2898038dbd"**  has already been declared [here](https://github.com/keep-starknet-strange/madara/blob/3678b42e7d9560a365f4ef49e01910e14881f161/tests/tests/test-rpc/test-starknet-rpc.ts#L732), hence [this](https://github.com/keep-starknet-strange/madara/blob/3678b42e7d9560a365f4ef49e01910e14881f161/tests/tests/test-rpc/test-starknet-rpc.ts#L796) test will fail.

[...]

PR #686, opened 2023-06-20
feat: add support at rpc to check for already declared class 🚀

bajpai244:
Add support for checking at RPC to look for already declared classes, and not proceed if they exist.

# Pull Request type

- Feature

The current sequencer behavior is to proceed with all declare class calls without checking if the class is already declared or not. 

According to the s[pec](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L146), this should result in an error.

Resolves: #663 

## What is the new behavior?

- checks whether the class already exists or not
- if exists, returns the error code according the spec
- a test has also been included to check this behavior for RPC

no


abdelhamidbakhta (approved):
lgtm

bajpai244:
lgtm ser 🚀

Issue #636, opened 2023-06-14
dev: fix header hash tests 🧪


EvolveArt:
# Bug Report

**Madara version:**

[4a9bc0a008428c594388a375fc3be367940ef165](https://github.com/keep-starknet-strange/madara/commit/4a9bc0a008428c594388a375fc3be367940ef165)

**Current behavior:**

Right now the test don't really test anything apart that we haven't changed the implementation.

**Expected behavior:**

We should test against a block-hash computed with the starknet cli/starkli.

lambda-0x:
I can take this up

abdelhamidbakhta:
> I can take this up.
> 
> EDIT: Is there a tool to calculate `transaction_commitment` and `event_commiment` for a given block? its explained [here](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/header/) but i don't understand that.
> 
> EDIT2: Method to calculate this values is wrong in the docs, Waiting on pathfinder team to know whats the right way.

Did we figure out the right approach or still not yet ?

lambda-0x:
from @drspacemn 

> Ahhhh I see. Okay yeah I'm fairly sure the docs are incorrect and pathfinder has it correct, but yes I'll double check w/ Ariel

https://t.me/MadaraStarknet/6091

will start working on this today.

drspacemn:
Yes the docs are incorrect here. The return should be `h(tx_hash, h(0,0))` for non invoke txs.

PR #695, opened 2023-06-20
dev: add test for header hash

lambda-0x:
# Pull Request type

- Testing

Resolves: #636 

No

## Other information

Not yet working.

I am using the genesis block in the `alpha-mainnet` as example.

Here is what I am able to gather:
- there are no events so its commitment is `0x0` 
    - ref: https://t.me/MadaraStarknet/6057
- transaction commitment before `0.11.1` is calculated as:
   - h(tx_hash, tx_signature)  for invoke
   - h(tx_hash, h(0,0)) for non-invoke
   - empty_tx_signature = h(0,0) 
   - in genesis block all transaction have empty signature.


abdelhamidbakhta (approved):
lgtm

PR #696, opened 2023-06-20
refacto: DRY functions in starknet rpc api server

dbejarano820:
DRYing some code in the client rpc for cleaner code and duplication reduction.

# Pull Request type

Repetitive code

Resolves: #629 

## What is the new behavior?
DRY code

No


dbejarano820:
> @dbejarano820 looks nice! I am just wondering, we almost always convert and then submit. Do you think it's worthy to keep those two methods separated? Or the use case in `rpc/lib.rs` is always both at the same time, and one function should be enough?
> 
> Totally open questions, otherwise nice job refactoring this. 👍

It's a great question.. It could totally go both ways. Personally I like to keep functions small and with 1 purpose.

abdelhamidbakhta (approved):
Lgtm

Issue #697, opened 2023-06-20
bug: panic on genesis block when sending inherent


lana-shanghai:
# Bug Report

**Current behavior:**

Panic at genesis block

**Steps to reproduce:**

`cargo +nightly run --release -- --dev`
A check that introduced the bug: 
`assert!(!SeqAddrUpdate::<T>::exists(), "Sequencer address can be updated only once in the block");`

```
Version: 0.1.0-alpha-b9cc5edfe46

   0: backtrace::capture::Backtrace::new
   1: sp_panic_handler::set::{{closure}}
   2: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/18bfe5d8a9ca0e226171e98f8f4ef071790f3352/library/alloc/src/boxed.rs:1999:9
      std::panicking::rust_panic_with_hook
             at /rustc/18bfe5d8a9ca0e226171e98f8f4ef071790f3352/library/std/src/panicking.rs:695:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/18bfe5d8a9ca0e226171e98f8f4ef071790f3352/library/std/src/panicking.rs:580:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/18bfe5d8a9ca0e226171e98f8f4ef071790f3352/library/std/src/sys_common/backtrace.rs:150:18
   5: rust_begin_unwind
             at /rustc/18bfe5d8a9ca0e226171e98f8f4ef071790f3352/library/std/src/panicking.rs:578:5
   6: core::panicking::panic_fmt
             at /rustc/18bfe5d8a9ca0e226171e98f8f4ef071790f3352/library/core/src/panicking.rs:67:14
   7: <pallet_starknet::pallet::Call<T> as frame_support::traits::dispatch::UnfilteredDispatchable>::dispatch_bypass_filter::{{closure}}
   8: frame_support::dispatch_context::run_in_context
   9: <madara_runtime::RuntimeCall as sp_runtime::traits::Dispatchable>::dispatch
  10: <sp_runtime::generic::checked_extrinsic::CheckedExtrinsic<AccountId,Call,Extra> as sp_runtime::traits::Applyable>::apply
  11: frame_executive::Executive<System,Block,Context,UnsignedValidator,AllPalletsWithSystem,COnRuntimeUpgrade>::apply_extrinsic
  12: madara_runtime::api::dispatch
  13: environmental::using
[...]

PR #699, opened 2023-06-20
start check if inherent data has been updated at block 2

lana-shanghai:
### Pull Request type

Bugfix

### What is the current behavior?

Panics on block 1 because of the check inside the inherent transaction. 
Resolves: [bug](https://github.com/keep-starknet-strange/madara/issues/697)

### What is the new behavior?

In block 1, the inherent transaction expects `false` in `SeqAddrUpdate` storage. However, the `on_finalize` hook panics in that case, even though the inherent is expected to come in before it and set it from `false` to `true`. So the storage is set to `true` in the genesis build, which means we need to skip the check in block 1 for the inherent transaction, and start it from block 2.


dbejarano820 (approved):

crates/pallets/starknet/src/lib.rs
```diff
@@ -407,7 +407,9 @@ pub mod pallet {
         #[pallet::weight((0, DispatchClass::Mandatory))]
         pub fn set_sequencer_address(origin: OriginFor<T>, addr: [u8; 32]) -> DispatchResult {
             ensure_none(origin)?;
-            assert!(!SeqAddrUpdate::<T>::exists(), "Sequencer address can be updated only once in the block");
+            if UniqueSaturatedInto::<u64>::unique_saturated_into(frame_system::Pallet::<T>::block_number()) > 1 {
```
abdelhamidbakhta, line 410: please add a comment explaining the rationale of this new check

abdelhamidbakhta (changes requested):
left comment

LucasLvy (approved):

Issue #165, opened 2023-04-05
feat(quality): replace `unwrap` by error handling 🔨


EvolveArt:
# Feature Request

**Describe the Feature Request**

`unwrap`s are dangerous to have within the runtime as a `panic` would simply crash the blockchain. 
Moreover it would impact forkless upgrades in the future.

**Describe Preferred Solution**

Go through the codebase and look for `.unwrap()`, replace them with custom errors that then propage to the pallet as `RuntimeError`s.

**Acceptance Crtieria**

No need to replace **ALL** unwraps within a single PR, even small PRs would be very helpful.

0xEniotna:
Hi ! Can I try something for this one ?

EvolveArt:
hey @0xEniotna sure please go ahead!

0xEniotna:
To be sure I understand properly as I'm new to Rust, in the following [snippet](https://github.com/keep-starknet-strange/madara/blob/main/crates/pallets/starknet/src/lib.rs#L284) :
```
#[pallet::call_index(0)]
#[pallet::weight(0)]
pub fn ping(origin: OriginFor<T>) -> DispatchResult {
 // Make sure the caller is from a signed origin and retrieve the signer.
          let _deployer_account = ensure_signed(origin)?;
          Pending::<T>::try_append(Transaction::default()).unwrap();
          log!(info, "Keep Starknet Strange!");
          Self::deposit_event(Event::KeepStarknetStrange);
          Ok(())
 }
```
Would that be an acceptable solution ?
```
pub fn ping(origin: OriginFor<T>) -> DispatchResult {
            // Make sure the caller is from a signed origin and retrieve the signer.
            let _deployer_account = ensure_signed(origin)?;
            Pending::<T>::try_append(Transaction::default())
                .map_err(|_| Error::<T>::TooManyPendingTransactions)?;;
            log!(info, "Keep Starknet Strange!");
            Self::deposit_event(Event::KeepStarknetStrange);
            Ok(())
        }
```

Thanks amigos

Eikix:
Can I take it?

chee-chyuan:
can i take this pretty please??

abdelhamidbakhta:
> can i take this pretty please??

Assigned to you

chee-chyuan:
i did some very small work to handle unwraps. i see a majority of it is being used in chain_spec.rs. iiuc, chain_spec should only run once at startup. would it be worth it to also handle the unwrap here??

[...]

PR #709, opened 2023-06-21
dev: remove unwraps from crates/pallet and crates/primitives

lambda-0x:
# Pull Request type

- Minor refactor

Resolves: #165 

## What is the new behavior?

- moved tests to `tests/` folder
- changed return value for `calculate_commitments` to remove unnecessary casting to `H256`
- propagate `unwrap`'s as error where required, and add `expect` where its not possible for `unwrap` to panic which appropriate message

Yes


crates/pallets/starknet/src/lib.rs
```diff
@@ -721,8 +721,10 @@ pub mod pallet {
             let transaction_nonce = transaction.nonce;
             let sender_address = transaction.sender_address;
 
+            let nonce_for_priority: u64 = transaction_nonce.try_into().map_err(|_| InvalidTransaction::Future)?;
+
```
EvolveArt, line 725: I'd prefer `InvalidTransaction::Custom` here as it's quite specific to our wrapper and the transaction will NOT be valid in the future.

crates/pallets/starknet/src/lib.rs
```diff
@@ -808,10 +810,11 @@ impl<T: Config> Pallet<T> {
     fn get_block_context() -> BlockContext {
         let block_number = UniqueSaturatedInto::<u64>::unique_saturated_into(frame_system::Pallet::<T>::block_number());
         let block_timestamp = Self::block_timestamp();
-        // Get fee token address. Its value is checked when we set it so it's fine to unwrap
-        let fee_token_address =
-            ContractAddress::try_from(StarkFelt::new(Self::fee_token_address().into()).unwrap()).unwrap();
         let chain_id = Self::chain_id_str();
+        let provided_fee_token_addr = Into::<StarkFelt>::into(Self::fee_token_address());
```
EvolveArt, line 814: this would be clearer
```suggestion
        let provided_fee_token_addr: StarkFelt = Self::fee_token_address().into();
```

crates/pallets/starknet/src/lib.rs
```diff
@@ -928,7 +931,9 @@ impl<T: Config> Pallet<T> {
         let pending = Self::pending();
 
         let global_state_root = Felt252Wrapper::ZERO;
-        let sequencer_address = Self::sequencer_address();
+        // Maybe we can change sequencer_address to be Felt252Wrapper itself.
+        let sequencer_address =
```
EvolveArt, line 935: we should definitely yes, please create a `good first issue`

crates/pallets/starknet/src/lib.rs
```diff
@@ -952,19 +957,19 @@ impl<T: Config> Pallet<T> {
                 parent_block_hash,
                 block_number,
                 global_state_root,
-                Felt252Wrapper::try_from(&sequencer_address).unwrap(),
+                sequencer_address,
                 block_timestamp,
                 transaction_count,
-                transaction_commitment.try_into().unwrap(),
+                Felt252Wrapper::from(transaction_commitment),
```
EvolveArt, line 963: no need for this once #667 is merged

crates/pallets/starknet/src/lib.rs
```diff
@@ -952,19 +957,19 @@ impl<T: Config> Pallet<T> {
...
+                Felt252Wrapper::from(event_commitment),
                 protocol_version,
                 extra_data,
             ),
             // Safe because `transactions` is build from the `pending` bounded vec,
             // which has the same size limit of `MaxTransactions`
-            BoundedVec::try_from(transactions).unwrap(),
-            BoundedVec::try_from(receipts).unwrap(),
+            BoundedVec::try_from(transactions).expect("max(len(transactions)) <= MaxTransactions"),
+            BoundedVec::try_from(receipts).expect("max(len(receipts)) <= MaxTransactions"),
```
EvolveArt, line 972: `.expect` can still panic which is what we want to avoid in the runtime at all cost. Please look into `.map_err` instead

crates/primitives/starknet/src/crypto/commitment/mod.rs
```diff
@@ -57,7 +56,10 @@ impl<T: CryptoHasherT> CommitmentTree<T> {
 /// # Returns
 ///
 /// The transaction commitment, the event commitment and the event count.
-pub fn calculate_commitments<T: CryptoHasherT>(transactions: &[Transaction], events: &[EventWrapper]) -> (H256, H256) {
+pub fn calculate_commitments<T: CryptoHasherT>(
```
EvolveArt, line 59: No need to touch this file on this PR because of #667

EvolveArt (changes requested):
Overall please avoid `.expect` in the runtime, doesn't really solve the issue of potential panicking.

crates/pallets/starknet/src/lib.rs
lambda-0x: i used `Future` because i saw mention of nonce too high https://crates.parity.io/sp_runtime/transaction_validity/enum.InvalidTransaction.html#variant.Future.

for `Custom` what should be the value of u8?

crates/pallets/starknet/src/lib.rs
lambda-0x: > expect can still panic

agreed but the invariant here is that `len(receipts)` will always be <= `MaxTransaction` because of the way we are creating it.

 so this would never panic and thus is okay to `unwrap` it. 

i changed it to `except` just to make it more descriptive.

crates/primitives/starknet/src/crypto/commitment/mod.rs
lambda-0x: haha nice, you already fixed that.

crates/pallets/starknet/src/lib.rs
lambda-0x: okay we can wait till that is merged then

lambda-0x:
blocked on #667

crates/pallets/starknet/src/lib.rs
EvolveArt: Right but `Future` implies that transaction would be valid in the future which is not right here.

You can do this for `Custom` : `pub(crate) const NONCE_DECODE_FAILURE: u8 = 1;`

crates/pallets/starknet/src/lib.rs
EvolveArt: Alright

crates/pallets/starknet/src/lib.rs
LucasLvy: no it should not, we're using the sequencer address in each transaction and converting it from Felt252Wrapper to bytes is expensive. If you want to change it you can change it to starkfelt but i don't think it adds any value and it probably implies a refacto on the blockifier side

LucasLvy (approved):

LucasLvy:
rebase por favor

lambda-0x:
@LucasLvy this is blocked on the other PR, because @EvolveArt has refactors some stuff in there. So i will rebase and update this once its merged

crates/pallets/starknet/src/lib.rs
LucasLvy: The starknet-rs `FieldElement` type is internally using the Montgomery representation which is expensive to convert to regular bytes

lambda-0x: feel free to revert the PR then, or lmk if its not possible to revert it directly

lambda-0x: feel free to revert the PR then, or lmk if its not possible to revert it directly i will open a PR

Issue #190, opened 2023-04-10
feat: RPC - starknet_getStateUpdate


abdelhamidbakhta:
Get the information about the result of executing the requested block

samueloh99:
Hi !! Can I get this issue ?

abdelhamidbakhta:
> 

sure, thanks

sparqet:
@samueloh99 Are you still on it ?

sparqet:
@abdelhamidbakhta can I take it ?

d-roak:
@sparqet are you working on this?

sparqet:
@d-roak no you can take it mate, I was about to tell @abdelhamidbakhta to unassign me

d-roak:
While looking at how I could implement the StateDiffs I checked how Blockifier is [handling the states](https://github.com/tdelabro/blockifier/blob/4b4f6ff20e5e872fb4cbcdd238bf980281c58b85/crates/blockifier/src/state/cached_state.rs).
And I think we should refactor the DictStateReader, to add a state cache like Blockifier has.
https://github.com/keep-starknet-strange/madara/blob/9d1126ea9321a28944a4656167384df879ca025e/crates/primitives/starknet/src/state/mod.rs#L30-L39

What are your thoughts on that? Any other suggestions on how to handle the StateDiffs?

LucasLvy:
Right now we don't have the state root so get state update should return mocked values

d-roak:
So in get_state_update should I return mocked values? Like:

```
StateUpdate {
    block_hash: block.header().hash(),
    new_root: block.header().state_root(), // mocked? Can I use global_state_root?
    old_root: parent_block.header().state_root(), // mocked? Can I use global_state_root?
    state_diff: StateDiff { ...... } // totally mocked
}
```

d-roak:
I will drop this one. I think @rkdud007 wants to take over

rkdud007:
I'll take it ser 🫡

abdelhamidbakhta:
> I want to take it ser 🫡

Assigned to you!

abdelhamidbakhta:
Any news on this @rkdud007 ?

rkdud007:
@abdelhamidbakhta sorry had a busy week 😅 if this issue is urgent to be done can give it to someone else! or else, i will do at this weekend :))

abdelhamidbakhta:
[...]

PR #720, opened 2023-06-21
Feat/rpc starknet get state update

Pilouche:
# Feat/rpc starknet get state update

- Feature
- Testing

Resolves: #190 

## What is the new behavior?

- Method was already added in `crates/client/rpc-core/src/lib.rs`.
- Added logic in `crates/client/rpc/src/lib.rs`. Afaik, state root & state diff are not yet computed, hence the empty `Vec` for the `StateDiff`.
- Added relevant tests in `tests/tests/test-rpc/test-starknet-rpc.ts`.
- Added a ☑️ in `docs/starknet_features_compatibility.md`. Should it be removed as the actual values are not yet returned by the RPC endpoint ?

No, it doesn't.


abdelhamidbakhta (approved):
lgtm

PR #746, opened 2023-06-24
dev: fix compatibility with blockifier

lambda-0x:
# Pull Request type

Resolves: #745 

Maybe

## Other information

List of duplicate functions i found (lmk if there are any other):
- verify_tx_version (updated function + test)
- handle_nonce (updated function + calls)
- validate_tx (updated function + no change in behaviour)
- validate_entrypoint_calldata (is same)
- validate_entry_point_selector (is same)

### DIFF
--- a/updated.md
+++ b/updated.md
@@ -1,5 +1,0 @@
-- verify_tx_version (updated)
-- handle_nonce (updated)
-- validate_tx (updated)
-- validate_entrypoint_calldata (verified)
-- validate_entry_point_selector (verified)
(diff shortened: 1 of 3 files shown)


LucasLvy:
All the fee logic is a duplicate also do you mind integrating that also ?

updated.md
```diff
@@ -0,0 +1,5 @@
+- verify_tx_version (updated)
+- handle_nonce (updated)
+- validate_tx (updated)
+- validate_entrypoint_calldata (verified)
+- validate_entry_point_selector (verified)
```
lambda-0x, line 5: oops added this by mistake will remove it.

lambda-0x:
as discussed with @LucasLvy, there is not much we can remove here due to current implementation

PR #760, opened 2023-06-27
feat: better local chains

d-roak:
- use 2 authorities in local deployments
- make madara-path independent of testnet

@LucasLvy please confirm that this version of `commad.rs` works with your usual commands


abdelhamidbakhta (approved):
lgtm

Issue #693, opened 2023-06-20
feat: SN OS Output Facts + Proving


drspacemn:
Upon finalization of a Madara block:
- run the StarkNet OS via the CairoRunner w/ the starknet layout
- capture the SN OS program outputs and store them on the client
- extract the Position Independent Execution(PIE) from the runner

**Describe Preferred Solution**

DA Options:
- StarkNet Canonical -> submit the PIE to the SHARP -> query the job status -> write the DA facts to ETH L1 once block is proved
- Sovereign Rollup -> write the DA facts to a DAS of our choosing (Celestia) 

**Related Code**

[StarkNet OS Run](https://github.com/starkware-libs/cairo-lang/blob/361fe32d5930db340ea78fe05aedfe706f6c9405/src/starkware/starknet/core/os/syscall_handler.py#L887)

[StarkNet OS Fact Storage](https://github.com/starkware-libs/cairo-lang/blob/361fe32d5930db340ea78fe05aedfe706f6c9405/src/starkware/starknet/storage/starknet_storage.py#L69)

drspacemn:
To include [issue#35](https://github.com/keep-starknet-strange/madara/issues/350) in scope

EvolveArt:
Research
---

First we can get a `Cairo PIE` (Position Independent Execution) by running `cairo-run` with `--cairo_pie_output`.
Q: Is that available in Cairo 1 ?
Q: If yes, how is it different than Cairo 0 ?

https://github.com/starkware-libs/cairo-lang/blob/361fe32d5930db340ea78fe05aedfe706f6c9405/src/starkware/cairo/sharp/client_lib.py#L6

-> Wrapper around `SHARP` API

https://github.com/starkware-libs/cairo-lang/blob/361fe32d5930db340ea78fe05aedfe706f6c9405/src/starkware/cairo/sharp/sharp_client.py#L11

-> SHARP Client
1. init_client: Initializes a SharpClient instance.
2. submit: Command for submitting a Cairo job to SHARP.
3. get_job_status: Command for retrieving the status of a SHARP Cairo job.
4. is_verified: Command for verifying if a fact is registered on the SHARP fact-registry.

https://github.com/starkware-libs/cairo-lang/blob/361fe32d5930db340ea78fe05aedfe706f6c9405/src/starkware/cairo/lang/vm/cairo_pie.py#L2

-> Cairo PIE!

```python
metadata: CairoPieMetadata
memory: MemoryDict
[...]

PR #777, opened 2023-06-28
feat: initialise proving / data availability

drspacemn:
Initialize KV Store for: fact storage, last proven block, cairo job keys, and cairo job statuses

Spawn task for mock proving
- Store mock fact in madara db
- Submit mock pie to the GPS/Prover
- Store the cairo job key in the madara db 

Spawn task for mock update
- Query last proven block from core contracts
- Check the associated job status
- If the proved, mock the updateState write to the Core Contracts

# Pull Request type

- Bugfix
- [x] Feature
- Build-related changes
- Testing
- Other (please describe):

No concept of proving the state transition

Resolves: #693 

Yes 

Users need to remove their local database as a new column is initialized.

It appears the `purge chain` cli command is broken. Action item to open a ticket for bug fix


spapinistarkware:
Yo shouldn't compute the DA at prove_current_block(). It should be computed by running the StarkNet OS cairo program.
The state diff should be computed in terms of what the OS program expects. If the OS is run inside a cairo-vm runner, with a custom hint, then this can be any code that can retrieve thins like txs, contract definitions, block hash and timestamp, nonces, etc...

The OS program then outputs the DA using the "output builtin".

drspacemn:
> Yo shouldn't compute the DA at prove_current_block(). It should be computed by running the StarkNet OS cairo program. The state diff should be computed in terms of what the OS program expects. If the OS is run inside a cairo-vm runner, with a custom hint, then this can be any code that can retrieve thins like txs, contract definitions, block hash and timestamp, nonces, etc...
> 
> The OS program then outputs the DA using the "output builtin".

Okay yes understood. I added this comment in the place I expect to run the [OS](https://github.com/keep-starknet-strange/madara/blob/4faa30ab97d4413cb98fa7e4b583e3a151676261/crates/client/data-availability/src/lib.rs#L86).

I've submitted the issue for `cairo-vm` to output the PIE after the run of the [OS](https://github.com/lambdaclass/cairo-vm/issues/1305). In your opinion, is this something we can format manually from the vm run itself as a short term solution?

spapinistarkware:
> > Yo shouldn't compute the DA at prove_current_block(). It should be computed by running the StarkNet OS cairo program. The state diff should be computed in terms of what the OS program expects. If the OS is run inside a cairo-vm runner, with a custom hint, then this can be any code that can retrieve thins like txs, contract definitions, block hash and timestamp, nonces, etc...
> > The OS program then outputs the DA using the "output builtin".
> 
> Okay yes understood. I added this comment in the place I expect to run the [OS](https://github.com/keep-starknet-strange/madara/blob/a1082af16da1a40244e17eac1339e363fa841282/crates/client/data-availability/src/lib.rs#L48).
> 
Looks good
> Is there any point in supporting the pre v0.11.0 DA format(+ pre v0.11.0 SN OS I'm assuming) or just use the DA output from the current version of the OS?
> 
latter
> I've submitted the issue for `cairo-vm` to output the PIE after the run of the [OS](https://github.com/lambdaclass/cairo-vm/issues/1305). In your opinion, is this something we can format manually from the vm run itself as a short term solution?

It'll probably be the same amount of work as just making the feature in cairo-vm, which is not very hard I think.

tdelabro:
@drspacemn What should we do with this PR?

Issue #791, opened 2023-06-30
feat: re-order benchmarking data


bajpai244:
# Feature Request

re-order[ benchmarking](https://keep-starknet-strange.github.io/madara/dev/bench/) data to show ERC20 data first and then ERC721 data.

**Describe Preferred Solution**

Modify the index.html of `gh-pages` branch for it.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes!!!!
<img src="https://github.com/keep-starknet-strange/madara/assets/41180869/8479fe00-1e9e-48ec-a759-3baf23d3ea17" width="200px" />

PR #792, opened 2023-06-30
dev: re-order metrics, put all ERC20 data first, and then ERC721 mints data

bajpai244:
change the ordering of benchmark data on the website, show ERC20 data first, and then ERC721 data.

# Pull Request type

- Feature

The benchmark data for both ERC20 and ERC721 is mixed, i.e. we are ordering data only per the metric.

Resolves: #791 

## What is the new behavior?

- data is now ordered in a way, that firstly all ERC20 benchmarks are show, and then ERC721.

No

### DIFF
--- a/dev/bench/index.html
+++ b/dev/bench/index.html
@@ -292,19 +292,26 @@
             graphsElem.className = "benchmark-graphs";
             setElem.appendChild(graphsElem);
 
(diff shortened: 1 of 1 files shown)


abdelhamidbakhta (approved):
lgtm

Issue #685, opened 2023-06-20
dev: Integration Tests w/ different pallets' configuration 🔨


EvolveArt:
Right now, integration tests are run against one runtime configuration defined in `crates/runtime/src/pallets.rs`.
However, with the integration of the state root computation under a flag `EnableStateRoot` we'd want to run integration tests both with this flag on and off.

In order to do this, we would need to define multiple runtimes or a runtime that can be configured through node configuration.

One direction to explore is also the use of [feature-flags](https://substrate.stackexchange.com/questions/1361/attempting-to-build-multiple-runtimes-stopped-by-accountnonceapi-account-nonce), like `madara-state-root` for instance.

MdTeach:
@EvolveArt I would love to take this

apoorvsadana:
Can I pick this one?

lambda-0x:
i can take this up

PR #798, opened 2023-06-30
dev: add feature flag to enable state root

lambda-0x:
# Pull Request type

- Testing

Resolves: #685 

No

### DIFF
--- a/scripts/run_node.sh
+++ b/scripts/run_node.sh
@@ -2,4 +2,27 @@
 # This script is meant to be run on Unix/Linux based systems
 set -e
 
-exec ../target/release/madara --dev --tmp --rpc-external --execution native --pool-limit=100000 --pool-kbytes=500000 --rpc-methods=unsafe --rpc-cors=all --in-peers=0 --out-peers=1 --no-telemetry
+while [[ $# -gt 0 ]]; do
+  case $1 in
+    --with-state-root)
+      with_state_root=true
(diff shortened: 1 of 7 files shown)


scripts/run_node.sh
```diff
@@ -2,4 +2,27 @@
...
+# Check if the --with-state-root flag is provided
+if [[ $with_state_root ]]; then
+    echo "Running with state root..."
+    cargo build --release --features madara-state-root
+    exec ../target/release/madara --dev --tmp --rpc-external --execution native --pool-limit=100000 --pool-kbytes=500000 --rpc-methods=unsafe --rpc-cors=all --in-peers=0 --out-peers=1 --no-telemetry
+else
+    echo "Running without state root..."
+    cargo build --release
+    exec ../target/release/madara --dev --tmp --rpc-external --execution native --pool-limit=100000 --pool-kbytes=500000 --rpc-methods=unsafe --rpc-cors=all --in-peers=0 --out-peers=1 --no-telemetry
+fi
```
lambda-0x, line 27: strictly speaking this logic is not required for running benchmark with state-root enabled, because we are specifying the `cargo build --features --release madara-state-root` step in CI itself.

But i have kept it here so its easier for anyone else to use this. And also now we can run this script directly without first running `cargo build` manually

EvolveArt (approved):
lgtm

Issue #795, opened 2023-06-30
feat: update `estimateFee` according to new spec


apoorvsadana:
`estimateFee` takes an array of requests instead of a single request based on the new RPC spec - https://github.com/starkware-libs/starknet-specs. The feature will be to adapt Madara's `estimateFee` function according to the new spec.

As a part of this, we will also need to fix the TS estimateFee test cases which won't work after we merge #744

exp-table:
Count on me to do it 🏄‍♂️

abdelhamidbakhta:
> Count on me to do it 🏄‍♂️

Let’s go

PR #799, opened 2023-06-30
chore(rpc): update `estimateFee` according to specs

exp-table:
# Pull Request type

Updating `estimateFee`JSON RPC API method so it follows the latest specs.

Currently `estimateFee` only takes a single transaction, while it should take an array according to the new specs.

Resolves: #795

## What is the new behavior?

`estimateFee` now takes an array of transactions, according to the new specs.

Yes.

## Other information

While the TS tests have been updated, the `starknet`dependency does seem to support the latest spec regarding `estimateFee`, breaking the test given `provider.getEstimateFee` expects a single tx and not an array.


apoorvsadana:
Can you search for `FIXME: https://github.com/keep-starknet-strange/madara/issues/795` in the code and fix the two test cases according to the new spec as well?

exp-table:
> Can you search for `FIXME: https://github.com/keep-starknet-strange/madara/issues/795` in the code and fix the two test cases according to the new spec as well?

I would but as specified above, we rely on an external dependency `starknet` which doesn't follow the latest spec for `estimateFee` [here](https://www.starknetjs.com/docs/API/classes/RpcProvider#getestimatefee). 

I can put this as a draft while they eventually modify it!

apoorvsadana:
sorry, might have missed your original PR details but I think you've fixed the test now is that correct? i remember `starknet` sending an array of requests because of which it was failing for me intitially

exp-table:
> sorry, might have missed your original PR details but I think you've fixed the test now is that correct? i remember `starknet` sending an array of requests because of which it was failing for me intitially

Yeah, I mistook `SequencerProvider` with `RpcProvider` ...

EvolveArt (changes requested):
Just need to add a few more tests to test `getEstimateFeeBulk` edge cases otherwise LGTM

EvolveArt (approved):
lgtm

PR #817, opened 2023-07-04
feat: store sierra class in runtime when using declare v2

tdelabro:
# Current behaviour

using declare_v2, the sierra program is not used

## What is the new behavior?

- the sierra program is now stored in the runtime
- it is returned when calling the get_class rpc method

No

### DIFF
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -257,19 +257,26 @@ where
         })?;
 
         let contract_address_wrapped = contract_address.into();
-        let contract_class = self
+        if let Some(sierra_contract_class) = self
             .overrides
             .for_block_hash(self.client.as_ref(), substrate_block_hash)
-            .contract_class_by_address(substrate_block_hash, contract_address_wrapped)
-            .ok_or_else(|| {
-                error!("Failed to retrieve contract class at '{contract_address}'");
-                StarknetRpcApiError::ContractNotFound
-            })?;
-
-        Ok(to_rpc_contract_class(contract_class).map_err(|e| {
-            error!("Failed to convert contract class at '{contract_address}' to RPC contract class: {e}");
-            StarknetRpcApiError::ContractNotFound
-        })?)
+            .sierra_contract_class_by_address(substrate_block_hash, contract_address_wrapped)
+        {
+            Ok(ContractClass::Sierra(sierra_contract_class.into()))
+        } else {
+            let casm_contract_class = self
+                .overrides
+                .for_block_hash(self.client.as_ref(), substrate_block_hash)
+                .casm_contract_class_by_address(substrate_block_hash, contract_address_wrapped)
+                .ok_or_else(|| {
+                    error!("Failed to retrieve contract class at '{contract_address}'");
+                    StarknetRpcApiError::ContractNotFound
+                })?;
+            try_blockifier_contract_class_into_rpc_contract_class(casm_contract_class).map_err(|e| {
+                error!("Failed to convert contract class at '{contract_address}' to RPC contract class: {e}");
+                StarknetRpcApiError::ContractNotFound.into()
+            })
+        }
     }
 
     /// Get the contract class hash in the given block for the contract deployed at the given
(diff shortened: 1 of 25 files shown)


crates/primitives/starknet/src/transaction/utils.rs
```diff
@@ -1,14 +1,11 @@
-use alloc::vec::Vec;
-
-use crate::execution::types::{EntryPointTypeWrapper, EntryPointWrapper};
-
 #[cfg(feature = "std")]
 mod reexport_std_types {
+    use alloc::vec::Vec;
```
LucasLvy, line 3: In feature std ?

LucasLvy (changes requested):

crates/primitives/starknet/src/transaction/utils.rs
tdelabro: Oh yes. Weird one

crates/client/rpc/src/lib.rs
```diff
@@ -257,19 +257,26 @@ where
...
+                .overrides
+                .for_block_hash(self.client.as_ref(), substrate_block_hash)
+                .casm_contract_class_by_address(substrate_block_hash, contract_address_wrapped)
+                .ok_or_else(|| {
+                    error!("Failed to retrieve contract class at '{contract_address}'");
+                    StarknetRpcApiError::ContractNotFound
+                })?;
+            try_blockifier_contract_class_into_rpc_contract_class(casm_contract_class).map_err(|e| {
+                error!("Failed to convert contract class at '{contract_address}' to RPC contract class: {e}");
+                StarknetRpcApiError::ContractNotFound.into()
```
EvolveArt, line 277: `into` needed ?

crates/pallets/starknet/src/lib.rs
```diff
@@ -571,7 +580,11 @@ pub mod pallet {
             ensure!(ContractClassHashes::<T>::contains_key(transaction.sender_address), Error::<T>::AccountNotDeployed);
 
             // Check class hash is not already declared
-            ensure!(!ContractClasses::<T>::contains_key(class_hash), Error::<T>::ClassHashAlreadyDeclared);
+            ensure!(!CasmContractClasses::<T>::contains_key(class_hash), Error::<T>::ClassHashAlreadyDeclared);
+
+            if let Some(sierra_class) = sierra_program {
```
EvolveArt, line 585: so it can be none ? what happens in this case ? means its casm directly right ?

crates/primitives/starknet/Cargo.toml
```diff
@@ -26,7 +26,7 @@ blockifier = { workspace = true, default-features = false, features = [
 cairo-lang-casm = { workspace = true, default-features = false }
 cairo-lang-casm-contract-class = { workspace = true }
 cairo-vm = { workspace = true }
-starknet-core = { workspace = true, optional = true }
+starknet-core = { workspace = true }
```
EvolveArt, line 29: `default-features = false` no ?

crates/primitives/starknet/src/execution/sierra_contract_class.rs
```diff
@@ -0,0 +1,98 @@
...
+/// A Sierra entrypoint
+#[derive(Debug, Clone, Encode, Decode, PartialEq, Eq, scale_info::TypeInfo)]
+pub struct SierraEntryPoint {
+    /// A unique identifier of the entry point (function) in the program
+    pub selector: Felt252Wrapper,
+    /// The index of the function in the program
+    pub function_idx: u64,
+}
+
+// O(0) type conversion conversion
```
EvolveArt, line 43: O(0) ?

crates/primitives/starknet/src/transaction/types.rs
```diff
@@ -236,19 +229,28 @@ pub struct DeployAccountTransaction {
     pub max_fee: Felt252Wrapper,
 }
 
-impl DeployAccountTransaction {
-    /// converts the transaction to a [Transaction] object
-    pub fn from_deploy(self, chain_id: Felt252Wrapper) -> Result<Transaction, TransactionConversionError> {
-        let salt_as_felt: StarkFelt = StarkFelt(self.salt.into());
-        let stark_felt_vec: Vec<StarkFelt> = self.calldata.clone()
+impl TryFrom<(DeployAccountTransaction, Felt252Wrapper)> for Transaction {
```
EvolveArt, line 232: how idiomatic is that kind of pattern ? I find it quite misleading and not sure that rustdoc would make it easy to understand

EvolveArt (changes requested):
left few comments

PR #830, opened 2023-07-07
fix invalid urls for block header, event hash, starknet state

h3lio5:
# Pull Request type

## What is the new behavior?

No change in the behaviour. 

No


EvolveArt (approved):
lgtm

PR #832, opened 2023-07-07
Add Braavos wallet in genesis

ClementWalter:
# Pull Request type

- Feature

No Braavos wallet in the genesis. It prevents Braavos wallet from working out of the box.

## What is the new behavior?

Contract and class hashes set in the genesis as expected by the wallet extension.

No

## Other information

I was not sure if I should put these into cairo-contracts and recompile then or create a new
"genesis assets" folder. Actually we don't necessary want to recompile genesis assets because
they are meant to mimic a Starknet state, so we just "to have them", and especially not "to touch them".

The place of the folder could be update though.


EvolveArt (approved):
lgtm

Issue #824, opened 2023-07-05
dev: split integration tests into multiple files 🔨


EvolveArt:
**Current Behavior**

All integration tests currently sit in `test-starknet-rpc.ts` which is becoming quite large.

**Expected Behavior**

Split these tests within multiple files, you can follow the organization of the `starknet` pallet tests folder.

MdTeach:
@EvolveArt I would love to take this

abdelhamidbakhta:
> @EvolveArt I would love to take this

let's go

PR #836, opened 2023-07-09
dev: split integration tests into multiple files 🔨

MdTeach:
# Pull Request type

- Testing

Resolves: #824 
## What is the new behavior?

- Splits the integration tests within multiple files

No

## Other information
Initially while defining the test case each test case defined with `describeDevMadara("test case", ()=>{....})` would run the new instance of the server. With this, each test file would run a new instance of Madara. 

To mitigate this server is initiated before running any test cases. In some cases, if the test suite needs a new instance of the server then it can be achieved with `describeDevMadara("test case", ()=>{....}, { runNewNode: true })`


abdelhamidbakhta (approved):
lgtm

PR #866, opened 2023-07-12
added `is_query` to handle estimate fee

apoorvsadana:
# Pull Request type

- Bugfix
- Feature

Estimate fee with version `0x100000000000000000000000000000001` doesn't work. Starknet-rs has added support for it and this PR integrates it.

Resolves: #646 

## What is the new behavior?

- Estimate fee works correctly now
- Once `dev/query_tx_version` is merged in `starknet-rs`, I will update the packages in this PR and we can merge

No

### DIFF
--- a/crates/primitives/starknet/src/transaction/mod.rs
+++ b/crates/primitives/starknet/src/transaction/mod.rs
@@ -43,6 +43,7 @@ use self::types::{
     TransactionExecutionInfoWrapper, TransactionExecutionResultWrapper, TransactionReceiptWrapper,
     TransactionValidationErrorWrapper, TransactionValidationResultWrapper, TxType,
 };
+use self::utils::{calculate_transaction_version, calculate_transaction_version_from_u8};
 use crate::execution::types::{CallEntryPointWrapper, ContractAddressWrapper, Felt252Wrapper};
 use crate::fees::{self, charge_fee};
 use crate::state::StateChanges;
@@ -306,6 +307,7 @@ impl Transaction {
         contract_class: Option<ContractClass>,
         contract_address_salt: Option<U256>,
         max_fee: Felt252Wrapper,
+        is_query: bool,
     ) -> Self {
         Self {
             tx_type,
@@ -318,6 +320,7 @@ impl Transaction {
             contract_class,
             contract_address_salt,
             max_fee,
+            is_query,
         }
     }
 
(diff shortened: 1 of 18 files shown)


crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -753,7 +756,7 @@ impl Transaction {
         AccountTransactionContext {
             transaction_hash: tx.transaction_hash(),
             max_fee: tx.max_fee(),
-            version: TransactionVersion(StarkFelt::from(1_u8)),
+            version: calculate_transaction_version_from_u8(self.is_query, 1_u8),
```
EvolveArt, line 759: `tx.version` no?

crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -795,7 +798,7 @@ impl Transaction {
         AccountTransactionContext {
             transaction_hash: tx.transaction_hash(),
             max_fee: tx.max_fee(),
-            version: tx.version(),
+            version: calculate_transaction_version(self.is_query, tx.version()),
```
EvolveArt, line 801: just `tx.version` no ?

EvolveArt (changes requested):
would probably remove `calculate_transaction_version_from_u8` and add `TransactionVersion::ONE`, `TransactionVersion::ZERO`

crates/primitives/starknet/src/transaction/mod.rs
apoorvsadana: Actually `InvokeTransaction` doesn't have a version field which is why this is hardcoded to 1. And v0 is deprecated and we don't handle it. However, in case of v0 our code will mostly fail with the wrong message (I guess invalid signature). So I spoke to greg and was thinking to create a new issue to throw a good error for v0.

crates/primitives/starknet/src/transaction/mod.rs
apoorvsadana: Do you mean why is it a function and not just `tx.version`? `DeclareTransaction` doesn't have a `version` field as well and instead has a function which is why we use that. I am not sure if this was the question

crates/primitives/starknet/src/transaction/mod.rs
EvolveArt: please do

apoorvsadana: Ok nevermind, we do throw an error for V0 transactions [here](https://github.com/keep-starknet-strange/madara/blob/b99d8bc5a1aa4449bfbe8ddde0765dd802856f80/crates/primitives/starknet/src/transaction/types.rs#L624)

EvolveArt (approved):
lgtm

PR #881, opened 2023-07-18
Hackathon Work

hooperben:
Most of the work we have done as part of this project.


tdelabro:
@hooperben Can you explain what is happening in the PR in its description? It will really help me review it.
Especially because I don't know about Celestia or anything it seems related too. So you also will have to explain what all this about. Sorry :)
Also,, can you rebase and push from the latest commit?

tdelabro (changes requested):
@hooperben Can you explain what is happening in the PR in its description? It will really help me review it.
Especially because I don't know about Celestia or anything it seems related too. So you also will have to explain what all this about. Sorry :)
Also,, can you rebase and push from the latest commit?

Issue #853, opened 2023-07-10
bug: contract call on non existing contract returns the wrong error message


ClementWalter:
# Bug Report

## Current behavior

Taking a random contract address as example, the behavior on goerli is

```
starkli call 0x01e96d9d047081203a40639c05ba879be96e91e612e000fc10b94c17bbd61062 bytecode
WARNING: no valid provider option found. Falling back to using the sequencer gateway for the goerli-1 network.
Error: Contract not found
```

while on madara it's:

```
starkli call 0x01e96d9d047081203a40639c05ba879be96e91e612e000fc10b94c17bbd61062 bytecode --rp
c=http://127.0.0.1:9944
Error: Contract error
```

## Wat do

Fix the returned error here:
https://github.com/keep-starknet-strange/madara/blob/a24d36520d035fb7629a1ad11fe2f0968c0472ef/crates/client/rpc/src/lib.rs#L247

to return Contract not found if the contract does not exist.

**Other information:**

starkli is 🔥

MdTeach:
@ClementWalter I would love to take this

ayushtom:
@MdTeach if you're aren't working on this right now, can I pick it up?

MdTeach:
@ayushtom I'm still on it I should be done by this week.

PR #884, opened 2023-07-21
fix: contract call on non existing contract returning correct error message

MdTeach:
# Pull Request type

[x] Bugfix

Resolves: #853 

## What is the new behavior?
- Returns `Error: Contract not found` instead of `Error: Contract error` if the called contract doesn't exist.

No

## Other information
No


crates/client/rpc/src/lib.rs
```diff
@@ -245,6 +247,9 @@ where
             })?
             .map_err(|e| {
                 error!("Failed to call function: {:#?}", e);
+                if e == PalletError::<Runtime>::ContractNotFound.into() {
+                    return StarknetRpcApiError::ContractNotFound;
+                }
```
tdelabro, line 252: I would rather have implemented it using the `From` trait.
Is there a reason you cannot use it there?

tdelabro: This way you can use the `?` keyword

tdelabro (changes requested):

crates/client/rpc/src/errors.rs
```diff
@@ -42,6 +45,15 @@ pub enum StarknetRpcApiError {
     UnimplementedMethod = 501,
 }
 
+impl From<DispatchError> for StarknetRpcApiError {
+    fn from(err: DispatchError) -> Self {
+        match err {
+            _ if err == PalletError::<Runtime>::ContractNotFound.into() => StarknetRpcApiError::ContractNotFound,
+            _ => StarknetRpcApiError::ContractError,
```
tdelabro, line 52: Just do a if/esle, it will be more idiomatic, there is no reason to prefer match here

crates/client/rpc/src/errors.rs
```diff
@@ -42,6 +45,15 @@ pub enum StarknetRpcApiError {
     UnimplementedMethod = 501,
 }
 
+impl From<DispatchError> for StarknetRpcApiError {
+    fn from(err: DispatchError) -> Self {
+        match err {
+            _ if err == PalletError::<Runtime>::ContractNotFound.into() => StarknetRpcApiError::ContractNotFound,
```
tdelabro, line 51: Don't use `into` on the right arm of you comparison. 
Just use the thing err is actually equal to

tdelabro (changes requested):

crates/client/rpc/src/errors.rs
MdTeach: In this context, in order to compare the given `err: DispatchError` with the [pallet error](https://github.com/keep-starknet-strange/madara/blob/8bba0040b29f09e817d3205a6c1410e4b77d49d1/crates/pallets/starknet/src/lib.rs#L441), I think that explicit conversion with `into` is required. Is there another way to do this without using `into` that I'm missing ?

tdelabro: There is a way, you can manualy create the struct:
```rust
DispatchError::Module(ModuleError{...})
```
But it will be unmaintainable and unreadable. So keep it the way it is.

tdelabro (approved):
Good job !!
Ty

Issue #927, opened 2023-07-27
feat: display dev account info on startup 


glihm:
# Feature Request

**Describe the Feature Request**

In web3 tools starting local node we usually have quick access to dev accounts.
Would be great if madara can print this information out to avoid a user looking into
the code base for that.

**Describe Preferred Solution**

Ideally, printing the address and and private key associated to a pre-funded account.
From feedbacks I had from @apoorvsadana, [the account is argent class at address](https://github.com/glihm/madara/blob/6d92e605f0cb0784e6188f2779e0370a83cb621f/crates/node/src/constants.rs#L1C1-L1C1) `0x2`, with
[private and public key here](https://github.com/glihm/madara/blob/6d92e605f0cb0784e6188f2779e0370a83cb621f/tests/tests/constants.ts#L27).

**Related Code**

[Account info](https://github.com/glihm/madara/blob/6d92e605f0cb0784e6188f2779e0370a83cb621f/crates/node/src/constants.rs#L1C1-L1C1)

[Signer info](https://github.com/glihm/madara/blob/6d92e605f0cb0784e6188f2779e0370a83cb621f/tests/tests/constants.ts#L27)

I am not sure yet where this print must reside, [chain_spec](https://github.com/glihm/madara/blob/6d92e605f0cb0784e6188f2779e0370a83cb621f/crates/node/src/chain_spec.rs) can be a good starting point.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes :rocket:

apoorvsadana:
Let's only print this extra information when the `--dev` flag is enabled

a-moreira:
hey @glihm @apoorvsadana  ! since the node info printing at startup comes from Substrate itself (it's inside the `runner` [here](https://github.com/paritytech/substrate/blob/31491fa24164b1cc9fffd5f07f06f41c8fef437d/client/cli/src/runner.rs#L73)), I propose having this print inside the `new_full` function in Madara. I create a branch on my fork and will open a PR if you guys think it's good (file changes in my fork [here](https://github.com/a-moreira/madara/pull/1/files))

![Screenshot 2023-07-31 at 01 01 56](https://github.com/keep-starknet-strange/madara/assets/48456867/038ffa67-e8f3-4ffc-898b-53134d212236)

Note that in the print the address and key are being printed as Field elements, we need to convert back to hex.

apoorvsadana:
[...]

PR #941, opened 2023-07-31
Feat: print development account info on node startup

a-moreira:
# Pull Request type

- Feature

Resolves: #927 

## What is the new behavior?

- Prints address and PK of the development account when running in `--dev` mode

No


crates/node/src/chain_spec.rs
```diff
@@ -89,6 +89,17 @@ pub fn development_config(enable_manual_seal: Option<bool>) -> Result<DevChainSp
     ))
 }
 
+use sp_core::hexdisplay::HexDisplay;
+pub fn print_development_account() {
+    let argent_account_address = Felt252Wrapper::from_hex_be(ARGENT_ACCOUNT_ADDRESS).unwrap();
```
apoorvsadana, line 94: we have a couple of accounts inside `crates/node/src/constants.rs`, 0x1 is the no validate account, 0x2 is the argent account, 0x3 is a OZ account, 0x4 is a no validate cairo 1 account, let's display all of them with these details

crates/node/src/chain_spec.rs
```diff
@@ -89,6 +89,17 @@ pub fn development_config(enable_manual_seal: Option<bool>) -> Result<DevChainSp
...
+    let signer = get_storage_key(&argent_account_address, "_signer", &[], 0);
```
apoorvsadana, line 95: the signer here is the public key, we can just take the address and private from the constants file mentioned above

apoorvsadana:
Actually, now that I think of it, I think we should move the printing logic inside the `chain_spec.rs`. We can only print the information when we have `ChainType::Development`. This way, when we have multiple chain_specs in the future, we can choose what and how to print separately. Moreover, keeping the code inside `chain_spec` also makes it easier to remember to add new accounts when we modify the genesis state.

a-moreira:
@apoorvsadana the printing logic can be in `chain_spec.rs` but the actual printing needs to be called *after* the node has started (after the `runner` starts) somewhere. This is what I've done.

apoorvsadana:
Hmm, curious why is that though? Like why can't we print in in the `chain_spec` regardless of whether the node has started or not?

a-moreira:
@apoorvsadana we can do `println!` but not `log::info!`.

apoorvsadana:
@a-moreira I just tried `log::info!` inside `testnet_genesis` in `chain_spec.rs` and it works fine for me, can you check?

a-moreira:
@apoorvsadana lol this is crazy, it works, but I had tried before inserting the `log::info()` inside `pub fn development_config()` (at line 60) and it didn't work. Would you mind trying that as a sanity check please?

a-moreira:
@apoorvsadana sorry for the confusion, my mistake. I just updated the code. The `constants.rs` file only contain the address though, should I get the Pub and pk somewhere else?

apoorvsadana:
Hey @a-moreira, ya the log doesn't work inside `development_config`. I think this might have something to do with what you were saying regarding the node starting, will need to dig deeper as well.

You can log the address and pk. The pk for 0x2 is `0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d`. `0x1` and `0x4` are no validate accounts, they don't need any pk (there's no validation in these contracts and all txns pass through). The private key for `0x3` should be the same as `0x1` as the public key is same (given they have the same signature scheme), can you test this?

a-moreira:
@apoorvsadana should be good now. By testing you mean any specific check?

apoorvsadana:
@a-moreira I meant can you try to do a transaction with the OZ account using Argent's pk just to be sure it's correct.

a-moreira:
@apoorvsadana sure! I added a TS test to check that. Took the opportunity to make one of the utility functions to be more general so I can use it in the test ;-)

a-moreira:
Almost good to go, I just need to refactor some stuff on chain_spec.rs because it has been heavily changed (constants moved to a json file)

a-moreira:
@apoorvsadana @tdelabro I decided to lazily instantiate the JSON file, given that it's now needed in more than one place. I'm assuming here btw that reading the file is slower comparing to deserializing it. I tried deserializing it only once inside the `lazy_static!` macro but couldn't get it to work.

Let me know if you think this makes sense. TBH the constants were easier to deal with before moving them to this JSON file, so that's what I came up with :-D

crates/node/src/chain_spec.rs
```diff
@@ -80,6 +88,25 @@ pub fn development_config(enable_manual_seal: Option<bool>) -> Result<DevChainSp
...
+    let loader: GenesisLoader = serde_json::from_str(&STARKNET_GENESIS).unwrap();
+    let starknet_genesis: madara_runtime::pallet_starknet::GenesisConfig<Runtime> = loader.into();
+    let no_validate_account_address = starknet_genesis.contracts[0].0.0.to_string();
+    let argent_account_address = starknet_genesis.contracts[1].0.0.to_string();
+    let oz_account_address = starknet_genesis.contracts[2].0.0.to_string();
+    let cairo_1_no_validate_account_address = starknet_genesis.contracts[3].0.0.to_string();
+
+    const ARGENT_PK: &str = "0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d";
+    log::info!("🧪 Using the following development accounts:");
+    log::info!("🧪 NO VALIDATE with address: {} and pk: {}", no_validate_account_address, "");
```
tdelabro, line 104: Remove the `and pk: {}` part. Or replace it with "and no pk"

Same on line 107

tdelabro (changes requested):

crates/node/src/chain_spec.rs
a-moreira: done!

crates/node/src/chain_spec.rs
```diff
@@ -8,6 +9,10 @@ use sp_core::storage::Storage;
 use sp_core::{Pair, Public};
 use sp_state_machine::BasicExternalities;
 
+lazy_static! {
+    static ref STARKNET_GENESIS: String = read_file_to_string("crates/node/src/genesis_assets/genesis.json");
+}
```
tdelabro, line 14: I think we should not have it as a static.
Static means it will stay in memory for the whole execution of the node (potentially months).

From what I see it is only used during the startup of the node:
- to print it
- to initialize the runtime

The startup is not something we try to optimize. Reading it two times from the disk instead of once is not a large overhead and this way we don't have to keep it in memory for the rest of the execution. It is dropped when not used anymore

tdelabro (changes requested):

crates/node/src/chain_spec.rs
```diff
@@ -56,6 +61,9 @@ pub fn development_config(enable_manual_seal: Option<bool>) -> Result<DevChainSp
         "dev",
         ChainType::Development,
         move || {
+            // Logging the development account
```
tdelabro, line 64: Read it only once.
There. 

Then pass it by ref as an argument to `print_development_accounts` and ` testnet_genesis` methods.

This way you still only read it once, and it don't have to be `static`

tdelabro:
@a-moreira Any news on this one?

apoorvsadana:
@a-moreira are you still working on this?

a-moreira:
@apoorvsadana yes, I'm just busy, will come back in a couple of days.

crates/node/src/chain_spec.rs
a-moreira: You are right. I was thinking I should avoid changing the signature of `testnet_genesis`, but it was a stupid take on my part.

crates/node/src/chain_spec.rs
a-moreira: Yep, you are completely right.

a-moreira:
@tdelabro Thanks for the review, you are correct. Early (or in this case useless) optimization is indeed the root of all evil lol.

That said, in the last commits I focused on avoiding repetition of code only. LMK if it's OK or I can roll back the changes.

crates/node/src/chain_spec.rs
a-moreira: BTW, using only references everywhere for this didn't work, so I had to resort to some cloning anyway.

tdelabro: What was the problem with references?

crates/node/src/chain_spec.rs
```diff
@@ -60,9 +60,12 @@ pub fn development_config(enable_manual_seal: Option<bool>, madara_path: PathBuf
...
+            print_development_accounts(genesis_loader.clone());
+
             DevGenesisExt {
                 genesis_config: testnet_genesis(
-                    madara_path.clone(),
+                    genesis_loader.clone(),
```
tdelabro, line 68: You don't think the second clone is required. Weird that clippy didn't gave you a warning

tdelabro (approved):

crates/node/src/chain_spec.rs
a-moreira: Replied below

crates/node/src/chain_spec.rs
a-moreira: The clone is necessary because we cannot move out of `genesis_loader`, a captured variable in an `Fn` closure.

If we try to pass a reference, we get problems inside `testnet_genesis`, because the `.into()` takes ownership, which is behind a shared reference. Then the compiler suggest a clone anyway:
Screenshot 2023-08-29 at 18 40 04

crates/node/src/chain_spec.rs
```diff
@@ -85,9 +88,27 @@ pub fn development_config(enable_manual_seal: Option<bool>, madara_path: PathBuf
     ))
 }
 
+// helper to print development accounts info
+// accounts with addresses 0x1 and 0x4 are NO VALIDATE accounts (don't require PK)
+// accounts with addresses 0x2 and 0x3 have the same PK
+pub fn print_development_accounts(genesis_loader: GenesisLoader) {
+    let genesis_config: madara_runtime::pallet_starknet::GenesisConfig<Runtime> = genesis_loader.into();
```
tdelabro, line 95: Why do you have to convert to GenesisConfig. The data is the same in GenesisLoader. Only the type changes. But for display it shouldn't be an issue

crates/node/src/chain_spec.rs
```diff
@@ -85,9 +88,27 @@ pub fn development_config(enable_manual_seal: Option<bool>, madara_path: PathBuf
...
+    let oz_account_address = genesis_config.contracts[2].0.0.to_string();
+    let cairo_1_no_validate_account_address = genesis_config.contracts[3].0.0.to_string();
+
+    const ARGENT_PK: &str = "0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d";
+    log::info!("🧪 Using the following development accounts:");
+    log::info!("🧪 NO VALIDATE with address: 0x{} and no pk", no_validate_account_address);
+    log::info!("🧪 ARGENT with address: 0x{} and pk: {}", argent_account_address, ARGENT_PK);
+    log::info!("🧪 OZ with address: 0x{} and pk: {}", oz_account_address, ARGENT_PK);
+    log::info!("🧪 CAIRO 1 with address: 0x{} and no pk", cairo_1_no_validate_account_address);
+}
```
tdelabro, line 107: Rather than using `to_string` and manually prefixing by `0x{}`, you can use rust native https://doc.rust-lang.org/std/fmt/trait.LowerHex.html.

Implement it on your `HexFelt` type and enjoy

https://stackoverflow.com/questions/56485167/how-to-format-a-byte-into-a-2-digit-hex-string-in-rust

tdelabro (changes requested):

crates/node/src/chain_spec.rs
a-moreira: Good idea, will do it this week.

a-moreira: Done

crates/node/src/chain_spec.rs
```diff
@@ -91,19 +92,21 @@ pub fn development_config(enable_manual_seal: Option<bool>, madara_path: PathBuf
...
+    let argent_pk: HexFelt =
+        Felt252Wrapper::from_hex_be("0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d")
+            .unwrap()
+            .into();
     log::info!("🧪 Using the following development accounts:");
-    log::info!("🧪 NO VALIDATE with address: 0x{} and no pk", no_validate_account_address);
-    log::info!("🧪 ARGENT with address: 0x{} and pk: {}", argent_account_address, ARGENT_PK);
-    log::info!("🧪 OZ with address: 0x{} and pk: {}", oz_account_address, ARGENT_PK);
-    log::info!("🧪 CAIRO 1 with address: 0x{} and no pk", cairo_1_no_validate_account_address);
+    log::info!("{}", format!("🧪 NO VALIDATE with address: {no_validate_account_address:#x} and no pk"));
```
tdelabro, line 106: Can't you get rid of the format? Since you already have a `"{}"` in the `log::info!`

crates/node/src/chain_spec.rs
```diff
@@ -91,19 +92,21 @@ pub fn development_config(enable_manual_seal: Option<bool>, madara_path: PathBuf
...
+    log::info!("{}", format!("🧪 ARGENT with address: {argent_account_address:#x} and pk: {argent_pk:#x}",));
+    log::info!("{}", format!("🧪 OZ with address: {oz_account_address:#x} and pk: {argent_pk:#x}"));
+    log::info!("{}", format!("🧪 CAIRO 1 with address: {cairo_1_no_validate_account_address:#x} and no pk"));
```
tdelabro, line 109: Same for those. YOu don't need the format

crates/primitives/starknet/src/execution/felt252_wrapper.rs
```diff
@@ -9,6 +9,7 @@
 
 use alloc::string::String;
 
+
```
tdelabro, line 12: run cargo fmt, I think we don't need this extra line

crates/pallets/starknet/src/genesis_loader.rs
```diff
@@ -9,6 +11,7 @@ use serde_with::serde_as;
 use starknet_core::serde::unsigned_field_element::UfeHex;
 use starknet_crypto::FieldElement;
 
+use crate::genesis_loader::fmt::Formatter;
```
tdelabro, line 14: Import it from `std::fmt` rather than its genesis_loader re-export

crates/pallets/starknet/src/genesis_loader.rs
```diff
@@ -1,3 +1,5 @@
+use core::fmt::LowerHex;
```
tdelabro, line 1: You can import it from `std::fmt` there

tdelabro (changes requested):

tdelabro:
Als run the linters ;)

crates/node/src/chain_spec.rs
a-moreira: True, my bad.

tdelabro (approved):

tdelabro:
Great job!

PR #944, opened 2023-07-31
Add function docs for node service.rs

istommao:
Add some function docs for node/src/service.rs

# Pull Request type

- [x] Documentation content changes

No


crates/node/src/service.rs
```diff
@@ -158,6 +184,7 @@ where
 
     let madara_backend = Arc::new(MadaraBackend::open(&config.database, &db_config_dir(config))?);
 
+    // Call the `build_import_queue` function to create an import queue and block import for the service.
```
tdelabro, line 187: This adds nothing more than what you already know by reading the code.
I think you can remove it

crates/node/src/service.rs
```diff
@@ -64,6 +64,32 @@ type FullSelectChain = sc_consensus::LongestChain<FullBackend, Block>;
...
+/// - `build_import_queue`: A function that builds the import queue and block import for the
+///                          blockchain. It takes various parameters and returns a result
+///                          containing the import queue and block import components, or an
+///                          error if the construction fails.
+///
+/// # Generic Parameters
+///
+/// - `RuntimeApi`: The runtime API trait that should be used with the blockchain. It should be
+///                  a type that implements `ConstructRuntimeApi` trait for the `Block` and
+///                  `FullClient` types.
```
tdelabro, line 85: `It should be a type that implements `ConstructRuntimeApi` trait for the `Block` and FullClient` types.`
 This is already said by the code. You can remove it

crates/node/src/service.rs
```diff
@@ -64,6 +64,32 @@ type FullSelectChain = sc_consensus::LongestChain<FullBackend, Block>;
...
+/// - `BIQ`: The type of the `build_import_queue` function, which should be a closure that takes
```
tdelabro, line 86: "The type of the `build_import_queue` function," does not sound correct. Functions don't have types. At least not in the way you mean there.

crates/node/src/service.rs
```diff
@@ -244,10 +271,24 @@ where
...
+/// * `sealing` - An optional `Sealing` value that specifies the sealing mode to use.
+///
+/// # Returns:
+///
+/// * A `Result` containing a `TaskManager` object if the service was successfully created, or a `ServiceError` if an error occurred.
 pub fn new_full(config: Configuration, sealing: Option<Sealing>) -> Result<TaskManager, ServiceError> {
     let build_import_queue =
         if sealing.is_some() { build_manual_seal_import_queue } else { build_aura_grandpa_import_queue };
 
+    // Create a partial service by calling `new_partial` function with the chosen import queue function.
```
tdelabro, line 290: `    // Create a partial service by calling `new_partial` function with the chosen import queue function.`
It add nothing to what the code is already expressing

crates/node/src/service.rs
```diff
@@ -244,10 +271,24 @@ where
...
+    // This step sets up essential components for the full client, including the backend, transaction pool, import queue, and more.
```
tdelabro, line 291: `including the backend, transaction pool, import queue, and more.`
This is already expressed by the code

crates/node/src/service.rs
```diff
@@ -493,6 +534,24 @@ pub fn new_full(config: Configuration, sealing: Option<Sealing>) -> Result<TaskM
...
+/// * `block_import` - A boxed block import for the service.
+/// * `task_manager` - A reference to a `TaskManager` object that manages tasks for the service.
+/// * `prometheus_registry` - An optional reference to a `Registry` object for Prometheus metrics.
+/// * `commands_stream` - A stream of engine commands for the manual seal process.
+///
+/// # Constraints
+/// - `RuntimeApi`: Generic constraint, the RuntimeApi must implement the ConstructRuntimeApi trait and be Send + Sync with 'static lifetime.
+///
+/// # Returns
+/// Returns `Result<(), ServiceError>` indicating the success or failure of the function.
```
tdelabro, line 554: This add nothing to what the code is already saying

crates/node/src/service.rs
```diff
@@ -493,6 +534,24 @@ pub fn new_full(config: Configuration, sealing: Option<Sealing>) -> Result<TaskM
...
+/// * `client` - An `Arc` containing the client for the service.
+/// * `transaction_pool` - An `Arc` containing the transaction pool for the service.
+/// * `select_chain` - The select chain for the service.
+/// * `block_import` - A boxed block import for the service.
+/// * `task_manager` - A reference to a `TaskManager` object that manages tasks for the service.
+/// * `prometheus_registry` - An optional reference to a `Registry` object for Prometheus metrics.
+/// * `commands_stream` - A stream of engine commands for the manual seal process.
+///
+/// # Constraints
+/// - `RuntimeApi`: Generic constraint, the RuntimeApi must implement the ConstructRuntimeApi trait and be Send + Sync with 'static lifetime.
```
tdelabro, line 551: This is alreaady said by the code

tdelabro (changes requested):
Hey,
First thank you a lot for this initiative.
I know it is not easy to choose what and what not to put in the documentation.
My overall feedback is that you chose to put too many things on the comment and a lot of them are just repetition of what is already in the code itself.
Rust is a pretty expressive language, we don't have to explain it again all the time.
Writing doc is more about the "Why?" that the "What?"

The "What?" is really ofter self explanatory in Rust. You don't often see a piece of code where you have no idea what is happening. If so, just renaming the variable and the functions will most of the time be enough to make what is happening clear.

Nevertheless, the language cannot help to express the reason why we are doing those things. This is where good doc and comment are useful.

Another point is that documentation is not checked by the CI tools. I can really easily become outdated if a parameter is added or a type changed. The kind of doc you wrote (listing, the args, the generic types, etc) becomes a burden to maintain.

Issue #919, opened 2023-07-27
dev: Make the client runtime agnostic


tdelabro:
Substrate is designed so that the client does not have to know about the runtime.
This allows for two things:
- runtime upgrage
- reuse the same code for different versions of the chain (testnet, mainnet, ...)

We are currently using the `madara-runtime` crate as a dependency in multiple crates in `crates/client/`. This has to stop.

Tasks:
- [x] remove it from the `rpc` crate imports
- [x] remove it from the `storage` crate imports. This will require a rework of the way starknet events are read. Maybe we cannot read them directly from the pallet `system` storage and will have to relly only on the `RuntimeApi`

EvolveArt:
I'll work on this

EvolveArt:
@tdelabro Just to confirm, we should also remove `pallet-starknet` dependency from client right ?

tdelabro:
I think so, yes. What is being imported from it?

EvolveArt:
> I think so, yes. What is being imported from it?

`use pallet_starknet::runtime_api::{ConvertTransactionRuntimeApi, StarknetRuntimeApi};` so we cannot really remove it

tdelabro:
Ok so let's keep it. It's not the runtime but the pallet. It will stay the same across different runtimes

PR #946, opened 2023-07-31
💚 remove madara_runtime from client deps

EvolveArt:
# Pull Request type

- Bugfix
- Feature
X Refactoring (no functional changes, no API changes)
- Build-related changes
- Testing
- Other (please describe):

Resolves: #919 

## What is the new behavior?

- Events are fetched using the transaction receipts
- No need to use the storage schema to fetch events

No

## Other information

We could take inspiration from frontier by having some `block_data_cache`:
```rust
/// Manage LRU caches for block data and their transaction statuses.
/// These are large and take a lot of time to fetch from the database.
/// Storing them in an LRU cache will allow to reduce database accesses
/// when many subsequent requests are related to the same blocks.
pub struct EthBlockDataCacheTask<B: BlockT>(mpsc::Sender<EthBlockDataCacheMessage<B>>);
```

If that's a good idea I can create the related issue.


tdelabro (changes requested):

crates/client/storage/src/overrides/schema_v1_override.rs
```diff
@@ -148,20 +143,6 @@ where
         }
     }
 
-    fn events(&self, block_hash: <B as BlockT>::Hash) -> Option<Vec<EventWrapper>> {
```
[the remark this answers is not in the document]
EvolveArt: No it wasn't since we use the transaction receipts, was probably used long time ago

crates/primitives/starknet/src/starknet_serde/mod.rs
```diff
@@ -28,6 +28,49 @@ fn string_to_felt(hex_str: &str) -> Result<Felt252Wrapper, String> {
     }
 }
 
+#[allow(missing_docs)]
+#[derive(Error, Clone, Copy, Debug, scale_codec::Encode, scale_codec::Decode, scale_info::TypeInfo)]
+pub enum StarknetRpcApiError {
```
tdelabro, line 33: Why put this struct in a file named `starknet_serde`?

crates/client/rpc/src/errors.rs
```diff
@@ -1,56 +1,24 @@
...
-    ProofLimitExceeded = 10000,
-    #[error("Too many keys provided in a filter")]
-    TooManyKeysInFilter = 34,
-    #[error("Internal server error")]
-    InternalServerError = 500,
-    #[error("Failed to fetch pending transactions")]
-    FailedToFetchPendingTransactions = 38,
-    #[error("Unimplemented method")]
-    UnimplementedMethod = 501,
+pub struct TxError(pub StarknetRpcApiError);
```
tdelabro, line 4: Why do we need this wrapper struct?
Can't you impl `jsonrpsee::core::Error` directly on `StarknetRpcApiError`?

crates/runtime/src/lib.rs
```diff
@@ -322,6 +324,13 @@ impl_runtime_apis! {
             };
             Ok(UncheckedExtrinsic::new_unsigned(call.into()))
         }
+
+        fn convert_error(error: DispatchError) -> StarknetRpcApiError {
+            if error == PalletError::<Runtime>::ContractNotFound.into() {
+                return StarknetRpcApiError::ContractNotFound;
+            }
+            StarknetRpcApiError::ContractError
+        }
```
tdelabro, line 333: Is there a way to have other more descriptive errors? Not just for `ContractNotFound`

tdelabro (changes requested):

crates/primitives/starknet/src/starknet_serde/mod.rs
EvolveArt: Should we create a new file just for this?

crates/client/rpc/src/errors.rs
tdelabro: I meant implenting `jsonrpsee::core::Error` things directly where `StarknetRpcApiError` is defined

tdelabro: And maybe it should not be into primitve/starknet, but rather primitve/rpc or primitive/errors

EvolveArt: Can't because we would need to add jsonrpsee as a dep and it breaks everything

crates/primitives/starknet/src/starknet_serde/mod.rs
tdelabro: wdyt?

crates/client/rpc/src/errors.rs
tdelabro: wdym "it break everything"?
Put it behind a feature flag may solve this

crates/primitives/starknet/src/starknet_serde/mod.rs
EvolveArt: yup put it under an `errors` crate

crates/client/rpc/src/errors.rs
EvolveArt: Unfortunately not cause we also import it within the pallet `runtime_api.rs` so we can't have it as a std feature flag

tdelabro: not as a std feature.
In cargo.toml:
```rust
jsonrpsee = {version = "...", optional = true}

[features]
std = [...]
jsonrpsee = ["std", "dep:jsonrpsee"]
```

Then in the file you create the `StarknetRpcApiError`:
```rust
#[cfg(feature = "jsonrpsee"]
impl From<StarknetRpcApiError> for jsonrpsee::core::Error {
    fn from(err: TxError) -> Self {
        jsonrpsee::core::Error::Call(CallError::Custom(ErrorObject::owned(err.0 as i32, err.0.to_string(), None::<()>)))
    }
}
```

This way you can get rid of the wrapper type.

tdelabro: And I don't think the error type should go in `primitive/starknet`

tdelabro: What I would do instead is define in pallet/runtime_api.rs a `StarknetTransactionExecutionError`:
```rust
pub enum StarknetTransactionExecutionError {
  ContractNotFound,
  ClassAlreadyDeclared,
  ClassHashNotFound,
  InvalidContractClass
}
```

I would keep my `StarknetRpcApiError` in the client, where it was `crates/client/rpc/src/errors.rs`, but add this:
```rust
StarknetRpcApiError {
   ...,
   Runtime(#[from] StarknetTransactionExecutionError)
}
```

This way you can keep your `From<StarknetRpcApiError> for jsonrpsee::core::Error` in the client, without bothering the pallet with it.

CHANGELOG.md
```diff
@@ -5,5 +5,6 @@
 - ci: downgraded windows runner to windows-latest
 - ci: added windows binaries build and upload the binaries to the release page
 - ci: add `CHANGELOG.md` and enforce it is edited for each PR on `main`
+- fix: removed `madara_runtime` as a dependency in the client crates
```
tdelabro, line 8: Also better more expressive errors

crates/pallets/starknet/src/lib.rs
```diff
@@ -1178,13 +1178,13 @@ impl<T: Config> Pallet<T> {
...
-            // Update the state trie for this contract in runtime storage.
-            ContractTries::<T>::set(contract_address, Some(state_tree.clone()));
-
             // We then compute the state root
             // And update the storage trie
             let state_root = state_tree.commit();
 
+            // Update the state trie for this contract in runtime storage.
+            ContractTries::<T>::set(contract_address, Some(state_tree.clone()));
+
```
tdelabro, line 1187: Why moving it after?

crates/pallets/starknet/src/runtime_api.rs
```diff
@@ -18,6 +18,15 @@ use sp_runtime::DispatchError;
 use crate::types::{NonceWrapper, StateCommitments};
 use crate::StateTrie;
 
+#[derive(scale_codec::Encode, scale_codec::Decode, scale_info::TypeInfo)]
```
tdelabro, line 21: No derive Error?

Don't seem to be needed. But is there a reason we would like it to be implemented?

tdelabro (approved):
Looks good. Maybe improve a bit the changelog.

crates/pallets/starknet/src/lib.rs
EvolveArt: Otherwise the state trie stored in runtime was not committed

tdelabro: So it's a bug fix?

EvolveArt: Yes

CHANGELOG.md
```diff
@@ -5,5 +5,8 @@
 - ci: downgraded windows runner to windows-latest
 - ci: added windows binaries build and upload the binaries to the release page
 - ci: add `CHANGELOG.md` and enforce it is edited for each PR on `main`
+- fix: removed `madara_runtime` as a dependency in the client crates and
+make errors more expressive
```
tdelabro, line 9: Line break

tdelabro (changes requested):
Run prettier too

tdelabro (approved):

PR #953, opened 2023-08-01
cli: deprecate official dev flag and fetch chain spec from url

d-roak:
~merge after we are not dependent on `--dev` cc @tdelabro~

Breaking:
- `--dev` behaviour changed
- ~chain-specs folder deleted, now we depend on https://chains.madara.zone/sharingan/testnet-sharingan-raw.json to be updated (for sharingan)~
- now we retrieve the chain-spec for sharingan from https://chains.madara.zone/sharingan/testnet-sharingan-raw.json
- added new flag --chain-spec-url to retrieve the chainspec from a url


tdelabro:
Hi @d-roak,

Can you break down to me:
- why is it better to depend on a config hosted online, rather than in the repository itself? 
- what happens if the website is down?
- are we not using the `--dev` flag at all? Or does it still do something in our CI?

Also, what do you mean by "not dependent on `--dev`"? I think we can already run nodes without the `--dev`

d-roak:
> * why is it better to depend on a config hosted online, rather than in the repository itself?

We can have them in the repository. Still, we should fetch them from an URL (even if it's the gh URL). Currently, we are packaging the sharingan configuration in the binary for it to be copied to the madara configs directory the first time we run it. After that we are keeping a big binary with no reason. It will also not scale if we add more chain-specs for secondary testnets and mainnet

> * what happens if the website is down?

We can use the new flag `--chain-spec-url` to retrieve the chainspec from a different url

> * are we not using the `--dev` flag at all? Or does it still do something in our CI?

Most our docs encourage people to use the `--dev` flag, which is not a problem, because it creates a quick environment for people to run an authority node. The thing is, we are not aware of all the things the `--dev` flag is running. With this PR we control what are the behaviour of it

> Also, what do you mean by "not dependent on `--dev`"? I think we can already run nodes without the `--dev`

Yep, u can ignore that sentence

tdelabro:
I would rather have the spec hosted on github, in the repository, where we can check and modify them easily. Don't you think?

Love the explanation about reducing the binary size! Thx!

crates/node/src/cli.rs
```diff
@@ -15,7 +15,6 @@ pub enum Sealing {
 /// Available testnets.
 #[derive(Debug, Copy, Clone, PartialEq, clap::ValueEnum)]
 pub enum Testnet {
-    Local,
```
tdelabro, line 18: No more local testnet? How come?

crates/node/src/command.rs
```diff
@@ -197,14 +207,22 @@ pub fn run() -> sc_cli::Result<()> {
...
+                if cli.run.testnet.is_some() {
                     match cli.run.testnet {
-                        Some(Testnet::Local) => {
-                            cli.run.run_cmd.shared_params.chain = Some(madara_path + "/chain-specs/local-raw.json");
-                        }
                         Some(Testnet::Sharingan) => {
+                            // even if chain-spec url is specified, overwrite it
+                            fetch_chain_spec(
+                                madara_path.clone(),
+                                "https://chains.madara.zone/sharingan/testnet-sharingan-raw.json".to_string()).unwrap();
```
tdelabro, line 225: As said, I would feel more confident if we used the URL of our github repository. I don't know this website, who is managing it and how secure it is.

Also easier to keep the spec up to date, should they change one day.

tdelabro (changes requested):

crates/node/src/cli.rs
d-roak: We never used the flag `--testnet local`, the `--chain local` works as usual

crates/node/src/command.rs
d-roak: madara.zone is managed by keep-starknet-strange, there is a cron job to fetch the chain-spec from the main branch. That being said, I agree with you. I'll change it to the gh url

tdelabro (approved):

tdelabro (approved):

d-roak:
breaking down this PR into smaller PR's

PR #964, opened 2023-08-04
refactor: added configs folder in root directory

d-roak:
these folders are necessary in the root for #953, I'll delete the folders from the other places


configs/genesis-assets/genesis.json
```diff
@@ -0,0 +1,202 @@
+{
```
tdelabro, line 1: Is this compatible with the genesis loader developed by @greged93 ?

tdelabro (approved):

configs/genesis-assets/genesis.json
d-roak: yep, is copy pasted from his

d-roak: https://github.com/keep-starknet-strange/madara/tree/main/crates/node/src/genesis_assets

tdelabro (approved):

PR #969, opened 2023-08-05
✨ rust tests improvements (lazy load node + block creation with txs) & adds get_block_transaction_count

EvolveArt:
# Pull Request type

- Bugfix
X Feature
- Build-related changes
X Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

- Adds `get_block_transaction_count`
- Finishes `get_block_hash_and_number`
- Lazy load node through same test file 
- Adds new options for block creation
- Supports transaction execution


starknet-rpc-test/get_block_hash_and_number.rs
```diff
@@ -1,19 +1,52 @@
...
+        .create_block(
+            vec![transfer_tokens(
+                &account,
+                argent_account_address,
+                FieldElement::from_hex_be(MINT_AMOUNT).expect("Invalid Mint Amount"),
+                None,
+            )],
+            BlockCreation::new(None, true),
+        )
+        .await?;
```
tdelabro, line 44: Why are you sending a tx? What behaviour are you testing with this regarding `block_hash_and_number`.

starknet-rpc-test/get_block_hash_and_number.rs
```diff
@@ -1,19 +1,52 @@
...
 #[tokio::test]
-async fn work_ok_at_initialization() -> Result<(), anyhow::Error> {
+async fn work_ok_at_start_and_with_new_blocks() -> Result<(), anyhow::Error> {
     let madara = MadaraClient::new(ExecutionStrategy::Native).await;
     let rpc = madara.get_starknet_client();
 
     assert_eq!(
         rpc.block_hash_and_number().await?.block_hash,
-        FieldElement::from_hex_be("0x031ebd02657f940683ae7bddf19716932c56d463fc16662d14031f8635df52ad").unwrap()
+        FieldElement::from_hex_be("0x035d394a3808b432702261c2d66f1cf2f111dd0c691ad2a8246b86a5088f73fa").unwrap()
```
tdelabro, line 19: I would rather have you also call `get_block` and compute the hash of the block you received.
Otherwise, we have to trust you on the fact that this is its correct hash.
Also if we change the block structure or the way we compute it's hash. Those tests will break

starknet-rpc-test/get_block_transaction_count.rs
```diff
@@ -0,0 +1,88 @@
...
+async fn get_madara_client() -> &'static MadaraClient {
+    MADARA_CLIENT.get_or_init(|| async { MadaraClient::new(ExecutionStrategy::Native).await }).await
+}
+
+#[tokio::test]
+async fn work_ok_with_empty_block() -> Result<(), anyhow::Error> {
+    let madara = get_madara_client().await;
+    let rpc = madara.get_starknet_client();
+
+    assert_eq!(rpc.get_block_transaction_count(BlockId::Tag(BlockTag::Latest)).await?, 0);
```
tdelabro, line 25: This won't work if a previous test using the same static Madara client sent some transactions but did not seal the block.

tdelabro: Or is creating a block the only way to send transactions?

tdelabro: Still, you have no certainty that previous tests didn't send any tx and that the latest block does not contain any.
Except if you rely on the fact that this test should be executed first, on a untouched new node. Something we  shouldn't assume

starknet-rpc-test/src/constants.rs
```diff
@@ -0,0 +1,20 @@
...
+pub const SALT: &str = "0x0000000000000000000000000000000000000000000000000000000000001111";
+
+// https://github.com/keep-starknet-strange/madara/blob/main/crates/node/src/chain_spec.rs#L191-L192
+pub const TEST_CONTRACT_CLASS_HASH: &str = "0x0000000000000000000000000000000000000000000000000000000000001000";
+pub const MINT_AMOUNT: &str = "0x0000000000000000000000000000000000000000000000000000000000000001";
+pub const DEPLOY_ACCOUNT_COST: &str = "0x00000000000000000000000000000000000000000000000000000000ffffffff";
+pub const CONTRACT_ADDRESS: &str = "0x0000000000000000000000000000000000000000000000000000000000000001";
+pub const FEE_TOKEN_ADDRESS: &str = "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7";
+pub const TOKEN_CLASS_HASH: &str = "0x0000000000000000000000000000000000000000000000000000000000010000";
+pub const ARGENT_CONTRACT_ADDRESS: &str = "0x0000000000000000000000000000000000000000000000000000000000000002";
```
tdelabro, line 20: 👍

tdelabro: Except it will probably be better if it were `FieldElement` rather than `&str`

starknet-rpc-test/src/lib.rs
```diff
@@ -20,10 +30,22 @@ use url::Url;
 pub struct MadaraClient {
     process: Child,
     client: Client,
-    rpc_request_count: Cell<usize>,
+    rpc_request_count: AtomicUsize,
```
tdelabro, line 33: You plan to share the struct between threads?

starknet-rpc-test/src/lib.rs
```diff
@@ -20,10 +30,22 @@ use url::Url;
...
+    parent_hash: Option<String>,
+    finalize: bool,
+}
+
+impl BlockCreation {
+    pub fn new(parent_hash: Option<String>, finalize: bool) -> Self {
+        BlockCreation { parent_hash, finalize }
+    }
+}
+
```
tdelabro, line 48:
```rust
madara
        .create_block(
            vec![transfer_tokens(
                &account,
                argent_account_address,
                FieldElement::from_hex_be(MINT_AMOUNT).expect("Invalid Mint Amount"),
                None,
            )],
            BlockCreation::new(None, true),
        )
        .await?;
```

I find it to be poorly expressive. When calling I don't know what is `None` and what this `bool` stand for. 
I would rather have no `new` function so that the caller can use:

```rust
BlockCreation {
    parent_hash: None,
    finalize: true,
}
```

This way we know what is going on

starknet-rpc-test/src/lib.rs
```diff
@@ -20,10 +30,22 @@ use url::Url;
 pub struct MadaraClient {
     process: Child,
     client: Client,
-    rpc_request_count: Cell<usize>,
+    rpc_request_count: AtomicUsize,
     starknet_client: JsonRpcClient<HttpTransport>,
 }
 
+#[derive(Debug, Default)]
+pub struct BlockCreation {
```
tdelabro, line 38: rename to something along the lines of ` BlockCreationParameters`

starknet-rpc-test/get_block_transaction_count.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    madara.create_block(vec![], BlockCreation::default()).await?;
```
tdelabro, line 27: I think the caller shouldn't have to worry about the parameters he calls this method with. Those are details we should abstract.

I would rather have multiple methods

```rust
madara.create_empty_block();
madara.create_block_with_txs(vec![<my txs>]);
madara.create_block_with_parent("0xbabe".to_string());
```

For now, there is no use case for creating an unfinalized block (correct me if I'm wrong), so just don't worry about it, set it to `true` every time without consulting the caller.

starknet-rpc-test/get_block_transaction_count.rs
```diff
@@ -0,0 +1,88 @@
...
+    assert_eq!(rpc.get_block_transaction_count(BlockId::Tag(BlockTag::Latest)).await?, 1);
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn work_ok_with_block_multiple_txs() -> Result<(), anyhow::Error> {
+    let madara = get_madara_client().await;
+    let _rpcc = madara.get_starknet_client();
+
```
tdelabro, line 63: comment this too. No need to waste computation

starknet-rpc-test/src/lib.rs
tdelabro: Rather remove it and use the multiple methods solution I suggested earlier

starknet-rpc-test/src/utils.rs
```diff
@@ -0,0 +1,28 @@
+use starknet_accounts::{Account, Call, Execution, SingleOwnerAccount};
+use starknet_core::types::FieldElement;
+use starknet_core::utils::get_selector_from_name;
+use starknet_providers::jsonrpc::{HttpTransport, JsonRpcClient};
+use starknet_signers::LocalWallet;
+
+use crate::constants::FEE_TOKEN_ADDRESS;
+
+pub fn transfer_tokens<'a>(
+    account: &'a SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>,
```
tdelabro, line 10: It would be more developer friendly it was implemented in a more OO way:
```rust
alice_account.transfer_tokens();
```

Either by creating a wrapper type over `SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet>` or, even better, through a trait `AccountActions` that would define those methods

tdelabro (changes requested):

starknet-rpc-test/get_block_hash_and_number.rs
EvolveArt: > Return the correct block number at blocks and hash after diverse blocks with diverse tx are created

starknet-rpc-test/get_block_transaction_count.rs
EvolveArt: > Or is creating a block the only way to send transactions?

Since we are running the tests sequentially and that one madara instance is shared *only* throughout its test file then I think it's fair to assume this, no ? Otherwise I can just create a new empty block before this test, fair enough

starknet-rpc-test/src/constants.rs
EvolveArt: Not sure about that, maybe we'd want to have them used as str and not FieldElement, I think we shouldn't make this assumption in constants

starknet-rpc-test/src/lib.rs
EvolveArt: to have it lazy loaded it's necessary yes even if we run tests with one thread in practice

starknet-rpc-test/get_block_transaction_count.rs
```diff
@@ -0,0 +1,101 @@
...
+use starknet_providers::{Provider, ProviderError};
+use starknet_rpc_test::constants::{ARGENT_CONTRACT_ADDRESS, MINT_AMOUNT, SIGNER_PRIVATE};
+use starknet_rpc_test::utils::AccountActions;
+use starknet_rpc_test::{ExecutionStrategy, MadaraClient};
+use starknet_signers::{LocalWallet, SigningKey};
+
+#[fixture]
+async fn madara() -> MadaraClient {
+    MadaraClient::new(ExecutionStrategy::Native).await
+}
```
tdelabro, line 21: This will run a new node for each test. Is it the behavior you wanted?
Otherwise, you can use `#[once]`
https://docs.rs/rstest/latest/rstest/attr.fixture.html#once-fixture

starknet-rpc-test/src/lib.rs
tdelabro: Still required with fixtures?

starknet-rpc-test/get_block_hash_and_number.rs
tdelabro: There is no `compute_hash` method easily accessible on the `starknet_core::Block`?

starknet-rpc-test/src/lib.rs
```diff
@@ -30,50 +57,72 @@ pub enum ExecutionStrategy {
...
+pub enum TestError {
+    #[error("No free ports")]
+    NoFreePorts,
+}
+
 impl Drop for MadaraClient {
     fn drop(&mut self) {
-        if let Err(e) = self.process.kill() {
-            eprintln!("Could not kill Madara process: {}", e)
+        self.process.kill().expect("Cannot kill process");
```
tdelabro, line 68: you lose the info in the `e` error.

starknet-rpc-test/src/constants.rs
```diff
@@ -0,0 +1,23 @@
...
+
+pub const MIN_PORT: u16 = 1025;
+pub const MAX_PORT: u16 = 65_535;
```
tdelabro, line 23: Document how you came up with those values

starknet-rpc-test/src/lib.rs
```diff
@@ -30,50 +57,72 @@ pub enum ExecutionStrategy {
...
-        );
-
-        let child_handle = Command::new(madara_path.to_str().unwrap())
+    async fn init(execution: ExecutionStrategy) -> Result<Self, TestError> {
+        // we keep the reference, otherwise the mutex unlocks immediately
+        let _mutex_guard = BACKGROUND_MADARA_MUTEX.lock().await;
+
+        let free_port = get_free_port()?;
+
+        let root = Path::new("../");
```
tdelabro, line 90: `repository_root`

starknet-rpc-test/src/lib.rs
```diff
@@ -30,50 +57,72 @@ pub enum ExecutionStrategy {
...
+        std::env::set_current_dir(&root).expect("Failed to change working directory");
```
tdelabro, line 91: Will this continue for the whole test execution?
Because in those tests we will probably want to access other files by doing things like "../compiled_contracts/".

starknet-rpc-test/get_block_hash_and_number.rs
EvolveArt: Unfortunately not, I suggest we let that for another PR since it might need a bit more work

starknet-rpc-test/get_block_transaction_count.rs
EvolveArt: Not usable for async fixtures, trying to find another solution..

starknet-rpc-test/src/lib.rs
```diff
@@ -1,17 +1,43 @@
...
+
+type TransactionExecution<'a> = Execution<'a, SingleOwnerAccount<&'a JsonRpcClient<HttpTransport>, LocalWallet>>;
+
+lazy_static! {
+        /// This is to prevent TOCTOU errors; i.e. one background madara node might find one
+        /// port to be free, and while it's trying to start listening to it, another instance
+        /// finds that it's free and tries occupying it
+        /// Using the mutex in `get_free_port_listener` might be safer than using no mutex at all,
+        /// but not sufficiently safe
+        static ref BACKGROUND_MADARA_MUTEX: Mutex<()> = Mutex::new(());
```
tdelabro, line 38: `Mutex::new(())` is `const` Is it possible to make it a const rather than a lazy_static, or does it cause trouble with the way we are using it?

tdelabro, line 38: No a really good name. FREE_PORT_ATTRIBUTION_MUTEX, or something along the lines will be more descriptive

tdelabro (changes requested):

PR #978, opened 2023-08-09
Li/dev/initial gas config param

lana-shanghai:
## Bugfix

View calls to deployed Cairo 1 contracts were failing with Out of Gas. 

Solves [issue](https://github.com/keep-starknet-strange/madara/issues/710)

## What is the new behavior?

We give a default value to `initial_gas` in the entrypoint wrapper.


d-roak (approved):
LGTM

CHANGELOG.md
```diff
@@ -28,3 +28,5 @@
 - fix: update RPC URL to use localhost instead of 0.0.0.0 in hurl.config file
 - fix: update the default port for running Madara locally in getting-started.md
   file from 9933 to 9944.
+- dev: replace the 0 initial gas value with u128::MAX because view call
```
tdelabro, line 31: Is this OK for a long time?
Shouldn't the value be something else than u128::MAX?

lana-shanghai: @tdelabro We discussed that this is customizable for appchains, but we leave this value to be u128::MAX. We don't know which contracts to benchmark to get a different value.

PR #993, opened 2023-08-11
fix: rebase fixes and adaptations

apoorvsadana:
Rebase of 
 - cairo-vm
 - cairo
 - blockifier
 - starknet-api

New features/bugs fixed
- VM follows the step limit set in the pallet
- Cairo 2.1.0


crates/pallets/starknet/src/tests/block.rs
```diff
@@ -121,7 +122,7 @@ fn get_block_context_works() {
             block_context.fee_token_address
         );
         // correct vm_resource_fee_cost
-        assert_eq!(HashMap::<String, f64>::new(), block_context.vm_resource_fee_cost);
+        assert_eq!(Arc::new(HashMap::<String, f64>::new()), block_context.vm_resource_fee_cost);
```
tdelabro, line 125: You can probably use `Default::default()` instead for readability

crates/primitives/starknet/src/execution/call_entrypoint_wrapper.rs
```diff
@@ -162,7 +161,7 @@ impl TryInto<CallEntryPoint> for CallEntryPointWrapper {
             // starknet-lib is constantly breaking it's api
             // I hope it's nothing important ¯\_(ツ)_/¯
             code_address: None,
-            initial_gas: Felt252::from_bytes_be(&self.initial_gas.0.to_bytes_be()),
+            initial_gas: self.initial_gas.try_into().unwrap(),
```
tdelabro, line 164: Why is this unwrap safe?

tdelabro (changes requested):

crates/pallets/starknet/src/tests/block.rs
apoorvsadana: I actually tried that before but assert wasn't able to get the type due to multiple implementations of `PartialEq`. I have now changed it to the following

```rust
let vm_resoursce_fee_cost: Arc<HashMap<String, f64>> = Default::default();
assert_eq!(vm_resoursce_fee_cost, block_context.vm_resource_fee_cost);
```

crates/primitives/starknet/src/execution/call_entrypoint_wrapper.rs
apoorvsadana: So we have initial_gas set to 0 everywhere which should be fixed over here #710. However, even after that, `initial_gas` should be less than u64 and if it's not I feel that error should be caught somewhere at the RPC level. Should I handle it here as well?

tdelabro: No need to handle the error if it cannot happen. But you need to explain why on a comment above the unwrap.

tdelabro: So just say what you tlod me in a comment

apoorvsadana: done!

tdelabro (approved):

tdelabro:
@apoorvsadana I think this one can be rebased now that the CI is fixed

PR #1011, opened 2023-08-17
fix: changed log::info to println

apoorvsadana:
The file copy was being logged using `log::info`. However, the logger is not initialised until the node starts (inside `create_runner`). Hence, I have changed the code to use println! instead


d-roak:
I don't think this is a good solution

apoorvsadana:
Ya I am not as well about what other implementations are possible, open to new approaches. What do you think are the issues though @d-roak?

d-roak:
@apoorvsadana i think the log::info is the way to go. The thing is, we should discover why it's not logging in that scope. It might be related with it being in the pallet crate (?)

apoorvsadana:
It doesn't log in that scope because the logger isn't initialized until `create_runner`. We can  maybe figure out a way to initialise it before (I googled a little but didn't find much, will need to probably go deeper). However, event frontier seems to use the same approach for logging before the node starts

https://github.com/paritytech/frontier/blob/aab7cfc3e038ec65ca4d1b2bc3d68e51aa162514/template/node/src/command.rs#L144

tdelabro:
We can maybe have another logger? For launch?

apoorvsadana:
Right, but I am just trying to understand what's the main issue with `println` that we want to tackle - `std` dependency, logging levels etc. Then we can ideally decide what's the correct solution for us.

tdelabro:
My problem with println is that it will print everytime, when most of the time the user won't care about what will be displayed. So we should be able to hide it easily. Log levels are a good solution for this.

apoorvsadana:
Got it, this makes sense. Will find what logger can we use.

tdelabro:
Maybe we should just move the logger initialization from the node to the CLI part of the code? Wdyt?

PR #1016, opened 2023-08-18
Feat/hash function

stompesi:
# Pull Request type

- Feature: Truncate the hash_bytes functions in `poseidon` and `pedersen` to a fixed number of bytes so that they can be hashed using all the data instead of using them dynamically.
- Documentation content changes: Fix typo
- Testing: Add `dynamic_string_hashing` test code

[72, 101, 108, 108, 111, 44, 32, 109, 97, 100, 97, 114, 97, 33, 33, 46, 32, 73, 116, 32, 105, 115, 32, 112, 101, 100, 101, 114, 115, 101, 110, 32, 104, 97, 115, 104, 46]

hash_bytes function of `poseidon` is error.
hash_bytes function of `pedersen` uses the array [72, 101, 108, 108, 111, 44, 32, 109, 97, 100, 97, 114, 97, 33, 33, 46, 32, 73, 116, 32, 105, 115, 32, 112, 101, 100, 101, 114, 115, 101, 110]. It means that the array which length is only 31 length is used.

However, hash functions don't work well enough.

## What is the new behavior?

I convert the bytes vector to Vec<Felt252Wrapper>.
But 32 bytes can be overflow, So I add one padding, In an array of 32, fill the 0th index with 0 and truncate by 31.

For example, 
[72, 101, 108, 108, 111, 44, 32, 109, 97, 100, 97, 114, 97, 33, 33, 46, 32, 73, 116, 32, 105, 115, 32, 112, 101, 100, 101, 114, 115, 101, 110, 32, 104, 97, 115, 104, 46]

is converted to

[
  [0, 72, 101, 108, 108, 111, 44, 32, 109, 97, 100, 97, 114, 97, 33, 33, 46, 32, 73, 116, 32, 105, 115, 32, 112, 111, 115, 101, 105, 100, 111, 110],
  [0, 32, 104, 97, 115, 104, 46, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
]

and the array is used for hash.

Yes (maybe?)

## Other information


tdelabro:
Hey @stompesi,

Thanks a lot for this really interesting PR. I have a few questions.

Shouldn't we follow some kind of standard for both hashing functions? Is your impl this standard? Or is it something you come up with yourselves? 

Isn't it a problem if our Poseidon doesn't behave in the same way that the Starknet one? Or I am missing something?

If it's too complex to explain by message, we can have a call, with me and other members of the Madara team, in order to better understand all the implications.

Tell me what you think is best :)

crates/primitives/starknet/src/crypto/hash/pedersen.rs
```diff
@@ -20,10 +21,27 @@ impl HasherT for PedersenHasher {
...
+        for i in 0..chunks {
+            let start = i * 31;
+            let end = cmp::min(start + 31, data.len());
+
+            // Create a buffer for our 32-byte chunk.
+            let mut buffer = [0u8; 32];
+            buffer[1..end - start + 1].copy_from_slice(&data[start..end]);
+
+            // Convert the buffer to a FieldElement and then to a Felt252Wrapper.
+            let field_element = FieldElement::from_bytes_be(&buffer).unwrap();
```
tdelabro, line 40: How are you sure it is ok to unwrap here?

crates/primitives/starknet/src/crypto/hash/pedersen.rs
```diff
@@ -20,10 +21,27 @@ impl HasherT for PedersenHasher {
...
-        let field_element = FieldElement::from_byte_slice_be(&data[..31]).unwrap();
-        Felt252Wrapper(pedersen_hash(&FieldElement::ZERO, &field_element))
+        // Calculate the number of 31-byte chunks we'll need, rounding up.
+        // (1 byte is used padding to prevent the value of field from being greater than modular)
+        // TODO: It is need a way to truncate bytes to fit into values smaller than modular (optimization)
+        let chunks = (data.len() + 30) / 31;
+
+        let mut hash_value = FieldElement::ZERO;
+
+        for i in 0..chunks {
```
EvolveArt, line 31: could we use the `.chunks` method here ?

stompesi: Thanks for the advice, I've found that implementing it using the `.chunks` that exist in std::slice will result in more concise code. Do you mind if I ask you to review it one more time?

crates/primitives/starknet/src/crypto/hash/pedersen.rs
stompesi: Yes Currently, I have rewritten the code by changing the existing code from_bytes_be to from_byte_slice_be. The `.chunks` function that @EvolveArt pointed out does not exceed the number of field's modulus mentioned in the existing starknet-ff because it divides the chunk into 31bytes. Therefore, it is considered safe.

stompesi: 1. Shouldn't we follow some kind of standard for both hashing functions? Is your impl this standard? Or is it something you come up with yourselves?

I'll have to check on that, it's something we implemented because we need to use hashes for danamic bytes. 
However, the hash function shouldn't be a problem if you use the hashes as they are implemented to determine the input value needed for the hash.

2. Isn't it a problem if our Poseidon doesn't behave in the same way that the Starknet one? Or I am missing something?

We have not modified the hash function implemented by Starknet, but we have fixed the existing error when using large input values(bytes).
The existing implementation of Poseidon is as follows, which causes the following error when large bytes are passed.

`let data = FieldElement::from_byte_slice_be(data).unwrap(); `

```
thread 'crypto::hash::poseidon::dynamic_string_hashing' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidLength', 
crates/primitives/starknet/src/crypto/hash/poseidon.rs:21:59
```

So we modified it by separating it into each chunk and hashing them using the existing poseidon_hash_many function.

If you have question, please reply~ Thank you

EvolveArt (approved):
lgtm

EvolveArt:
Once @tdelabro approves we will be able to merge!

From my understanding of the code it's safe to `unwrap` in that case, unless the `FieldElement` is of a different field size.

crates/primitives/starknet/src/crypto/hash/pedersen.rs
tdelabro: @stompesi, can you comment on your unwrap explaining why it is safe? Or even, when will it stop being safe? So that future developers know this and don't do any mistakes

stompesi: @tdelabro I added a comment on the reason that the unwrap is safe.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Madara Changelog
 
+- feat: modify the hash_bytes functions in `poseidon` and `pedersen` for dynamic
```
tdelabro, line 3: This goes under the `next release` section

tdelabro (approved):

crates/primitives/starknet/src/crypto/hash/pedersen.rs
```diff
@@ -65,3 +78,39 @@ impl DefaultHasher for PedersenHasher {
...
+    let hasher = PedersenHasher::hasher();
+
+    let message = "Hello, madara!!. It is pedersen hash.".to_string(); // 37 bytes
+    let message = message.as_bytes();
+    let hash_value = hasher.hash_bytes(message);
+
+    assert_eq!(
+        hash_value,
+        Felt252Wrapper(
+            FieldElement::from_str("0x03501abfd0e0606ecab6702213a03294b81837e4d48232df3c39a62b88cc6f73").unwrap()
```
apoorvsadana, line 95: just curious, what did we use to calculate this value?

apoorvsadana (commented):
added two small comments, otherwise lgtm

crates/primitives/starknet/src/crypto/hash/pedersen.rs
stompesi: I convert the `"Hello, madara!!. It is pedersen hash."` to bytes array and split the bytes with chunks.

crates/primitives/starknet/src/crypto/hash/pedersen.rs
```diff
@@ -20,10 +20,23 @@ impl HasherT for PedersenHasher {
...
+        for chunk in data.chunks(CHUNK_SIZE) {
+            // It is safe to unwrap here because we know that the chunk size is 31 and the value can not
+            // overflow than the field's modulus value. In more detail, the FieldElement Maximum value is 2^251
+            // + 17 * 2^192. So the chunk (31 bytes is 248 bits) is smaller than the maximum value (== 2^248 - 1
+            // < 2^251 + 17 * 2^192). So it is safe to unwrap here.
+            let field_element = FieldElement::from_byte_slice_be(chunk).unwrap();
+            hash_value = pedersen_hash(&hash_value, &field_element);
+        }
+
+        Felt252Wrapper(hash_value)
```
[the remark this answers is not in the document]
stompesi: In my opinion, the `hash_bytes_in_chunks` function is unnecessary because `hash_bytes` handles bytes of various lengths and `CHUNK_SIZE` can be a maximum of 32 to handle efficiently. 

However, if you give `CHUNK_SIZE` as 32, the value can be 2^256 - 1, which can exceed the maximum value of the FieldElement and cause an error. Therefore, I set `CHUNK_SIZE` to 31, which is 2^248 - 1, which does not exceed the maximum value of the FieldElement and processes it with maximum efficiency. So, I think the `hash_bytes_in_chunks` function is not necessary to handle CHUNK_SIZE as a separate parameter.

tdelabro (approved):

crates/primitives/starknet/src/crypto/hash/pedersen.rs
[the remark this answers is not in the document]
d-roak: @apoorvsadana I don't think that is an issue here. The logic we are using for computing the hash_bytes is different in pedersen and poseidon. Trying to find common ground will result in an over-engineered solution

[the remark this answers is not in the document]
apoorvsadana: Understoof @stompesi. Ya makes sense @d-roak, looks good then

crates/primitives/starknet/src/crypto/hash/pedersen.rs
apoorvsadana: Oh ok, what I meant is did you use any tool/script which does this for you and the got the value?

d-roak (approved):
lgtm

Issue #982, opened 2023-08-10
feat: Index config files and add a CI task to ensure that the CI files are correctly listed


d-roak:
This task is related with the PR #974 

For fetching the config files from the URLs we need to know the files that are available for download. The idea is to create a `configs/index.json` with the folder structure of the configs subdirs, and download everything with the same folder structure to `madara_path`. We should also add a CI task to ensure the `index.json` is always updated

d-roak:
Also the `cairo-contracts` should build to `configs/cairo-contracts` instead of `cairo-contracts/build`

PR #1018, opened 2023-08-18
feat: Indexing configs

d-roak:
- pr (on close) ci commented (was not working)
- we are now indexing the files that we retrieve online, and validating  their hashes
- ci now verifies if the hashes are correct
- ci enforces changes in the `docs/genesis.md`, if we had changes in the `configs/genesis-assets/genesis.json`
- documentation on our custom flags and parameters customizations
- we have flags for disabling URL fetching
- we have flags for retrieving/loading custom chain specs and config indexers (currently, the config indexer must have the same format as ours)

Resolves: #982

Following PR required to change: `index.json` and `constants.rs` to point to the official madara repo


.github/workflows/pull-request-close.yml
```diff
@@ -12,27 +12,26 @@ jobs:
...
-          echo "Deleting caches..."
-          for cacheKey in $cacheKeysForPR
-          do
-            echo "Deleting cache $cacheKey"
-            gh actions-cache delete $cacheKey -B $BRANCH --confirm
-          done
-          echo "Done"
-        env:
-          GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
+      # - name: Generate a token
```
tdelabro, line 15: Did you keep it commented on purpose?

crates/node/src/configs.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub url: Option<String>,
+}
+
+pub fn fetch_and_validate_file(
+    remote_base_path: String,
+    file: File,
+    dest_path: String,
+    force_fetching: bool,
+) -> Result<(), String> {
+    if let Some(url) = file.url {
```
tdelabro, line 24:
```rust
let full_url = file.url.or_else(|| dest_path.split("configs/").collect::<Vec<&str>>()[1].split('/').collect::<Vec<&str>>().join("/"))

utils::fetch_from_url(full_url, dest_path, force_fetching)?;

File: crates/node/src/command.rs [hunk 3]
```diff
@@ -210,59 +208,74 @@ pub fn run() -> sc_cli::Result<()> {
...
+                    // If testnet flag is specified, right now it's only Sharingan
+                    for chain_spec in madara_configs.chain_specs {
+                        configs::fetch_and_validate_file(
+                            madara_configs.remote_base_path.clone(),
+                            chain_spec,
+                            madara_path.clone() + "/configs/chain-specs/",
+                            cli.run.update_configs,
+                        )?;
+                    }
                 }
```
tdelabro, line 279: This could be a match of `cli.run.testnet`, `cli.run.chain_spec_url` and `cli.run.disable_url_fetch`

crates/node/src/command.rs
```diff
@@ -210,59 +208,74 @@ pub fn run() -> sc_cli::Result<()> {
...
-                    utils::fetch_from_url(
-                        constants::SHARINGAN_CHAIN_SPEC_URL.to_string(),
+                utils::fetch_from_url(
+                    chain_spec_url.clone(),
+                    madara_path.clone() + "/chain-specs",
+                    cli.run.update_configs,
+                )?;
+                let chain_spec = chain_spec_url.split('/').last().expect("Chain spec file name not found");
+                cli.run.run_cmd.shared_params.chain = Some(madara_path + "/chain-specs/" + chain_spec);
+            } else if let (Some(Testnet::Sharingan), false) = (cli.run.testnet, cli.run.disable_url_fetch) {
```
tdelabro, line 261: Why does `cli.run.disable_url_fetch` has to be `false`?

crates/node/src/command.rs
```diff
@@ -210,59 +208,74 @@ pub fn run() -> sc_cli::Result<()> {
...
+                    configs::fetch_and_validate_file(
+                        madara_configs.remote_base_path.clone(),
+                        asset,
+                        madara_path.clone() + "/configs/genesis-assets/",
+                        cli.run.update_configs,
                     )?;
                 }
             }
 
             if let (Some(chain_spec_url), None) = (cli.run.chain_spec_url.clone(), cli.run.testnet) {
```
tdelabro, line 253: What if `cli.run.disable_url_fetch` is true at this point? Do we still fetch from URL?

tdelabro (changes requested):
the `pub fn run() -> sc_cli::Result<()>` function is getting gigantic. And heavily nested. Which is terrible for readability. Do you think you can find a way to split it into different coherent sub-functions?
Maybe by putting the body of each individual subcommand match arm in its dedicated function.

crates/node/src/cli.rs
```diff
@@ -36,17 +36,34 @@ pub struct ExtendedRunCmd {
     #[clap(flatten)]
     pub run_cmd: RunCmd,
 
+    /// Load a custom chain-spec from an url
     #[clap(long)]
-    pub testnet: Option<Testnet>,
+    pub chain_spec_url: Option<String>,
```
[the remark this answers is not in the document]
d-roak: What might be the benefits? We currently don't have the `reqwest` imported on the node crate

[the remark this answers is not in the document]
tdelabro: Value hint will help for autocompletion in the terminal.

Having the type be an Url would probably comes with automatic validation of the input format

crates/node/src/command.rs
d-roak: if the `disable_url_fetch` is enabled, but we set manually an URL, it should be fetched. E.g.
`--disable-url-fetch --chain-spec-url https://domain.com/chain-spec.json`

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
-            base_path.to_str().unwrap().to_string()
+            cli.run.madara_path = Some(base_path);
         }
         _ => {
             let home_path = std::env::var("HOME").unwrap_or(std::env::var("USERPROFILE").unwrap_or(".".into()));
             let path = format!("{}/.madara", home_path);
             cli.run.run_cmd.shared_params.base_path = Some((path.clone()).into());
-            cli.run.madara_path = Some((path.clone()).into());
-            path
+            cli.run.madara_path = Some(path.into());
```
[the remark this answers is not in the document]
tdelabro: Also this would be better written like this:

```rust
let path = cli.run.madara_path
                       .or(cli.run.run_cmd.shared_params.base_path)
                       .or_else(|| {
           let home_path = std::env::var("HOME").unwrap_or(std::env::var("USERPROFILE").unwrap_or(".".into()));
            format!("{}/.madara", home_path);
});

cli.run.run_cmd.shared_params.base_path = path;
cli.run.madara_path = path
```

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
+
+fn set_dev_environment(cli: &mut Cli) {
+    // create a reproducible dev environment
+    cli.run.run_cmd.shared_params.dev = false;
+    cli.run.run_cmd.shared_params.chain = Some("dev".to_string());
+
+    cli.run.run_cmd.force_authoring = true;
+    cli.run.run_cmd.alice = true;
+
+    // we can't set `--rpc-cors=all`, so it needs to be set manually if we want to connect with external
```
tdelabro, line 81: How comes?

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
+    let local_path = utils::get_project_path();
+
+    match (
+        cli.run.chain_spec_url.clone(),
+        cli.run.testnet,
+        cli.run.disable_url_fetch,
+        cli.run.disable_madara_configs,
+        local_path,
+    ) {
+        (Some(chain_spec_url), _, _, _, _) => {
```
tdelabro, line 99: This one can be taken out of the `match`,  as a `if let` with an early return

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
+                configs::fetch_and_validate_file(
+                    madara_configs.remote_base_path.clone(),
+                    chain_spec,
+                    madara_path.clone() + "/configs/chain-specs/",
+                    cli.run.update_configs,
                 )?;
             }
+            cli.run.run_cmd.shared_params.chain = Some(madara_path + "/chain-specs/testnet-sharingan-raw.json");
         }
+        _ => {}
```
tdelabro, line 128: So if either ` cli.run.disable_url_fetch` or `cli.run.disable_madara_configs` if `true` we will skip the two previous arms and go directly to this last "do nothing" arm?

Are you sure this is the behavior you want?

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
-            let src_path = src_path + "/configs/cairo-contracts/" + file;
-            utils::copy_from_filesystem(src_path, madara_path.clone() + "/cairo-contracts")?;
-        } else {
-            utils::fetch_from_url(
-                constants::CAIRO_CONTRACTS_URL.to_string() + file,
-                madara_path.clone() + "/cairo-contracts",
+    if cli.run.run_cmd.shared_params.chain.is_some() {
+        cli.run.run_cmd.rpc_external = true;
+        cli.run.run_cmd.rpc_methods = RpcMethods::Unsafe;
+    }
```
tdelabro, line 134: I need a comment to understand why are we doing this

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
+
+    Ok(())
+}
+
+fn fetch_madara_configs(cli: &mut Cli) -> Result<(), String> {
+    let madara_path = cli.run.madara_path.clone().expect("Failed retrieving madara_path").to_str().unwrap().to_string();
+    let local_path = utils::get_project_path();
+
+    if let Some(configs_url) = cli.run.configs_url.clone() {
+        utils::fetch_from_url(configs_url, madara_path.clone() + "/configs", cli.run.update_configs)?;
```
tdelabro, line 144: the fn name `utils::fetch_from_url` imply (to me) that the fn will return the fetched values that I can then use to do whatever I want.

This function instead write the fetched values directly to the disk. A more descriptive name is needed. Maybe "write_madara_confings_to_disk_from_url".

Is it madara_configs or chain_specs that are fetched?

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
+    } else if let Ok(ref src_path) = local_path {
+        let src_path = src_path.clone() + "/configs/index.json";
+        utils::copy_from_filesystem(src_path, madara_path.clone() + "/configs", cli.run.update_configs)?;
+    } else if !cli.run.disable_url_fetch {
```
tdelabro, line 148: What happend at this point if `cli.run.disable_url_fetch` is true. Does madara still work ? If not we should return an error message enjoining the use to pass one of the 3 required arg

tdelabro: My understanding is that we will panic at l158 `.expect("Failed to load madara configs");`
I would rather return an error earlier if madara can't work with this combination of args

crates/node/src/command.rs
```diff
@@ -52,69 +52,138 @@ impl SubstrateCli for Cli {
...
+            madara_path.clone() + "/configs",
+            cli.run.update_configs,
+        )?;
+    }
+
+    let madara_configs: configs::Configs =
+        serde_json::from_str(&utils::read_file_to_string(madara_path.clone() + "/configs/index.json")?)
+            .expect("Failed to load madara configs");
+    for asset in madara_configs.genesis_assets {
+        if let Ok(ref src_path) = local_path {
```
tdelabro, line 160: if we don't plan on handling the error return by `get_project_path`, `local_path` should be an option rather than an Error.

But the true question is : are we absolutely sure we don't want to handle this error?

tdelabro, line 160: Invert those two lines.
First you match your local path, then you iterate your assets and use for each
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

crates/node/src/configs.rs
```diff
@@ -0,0 +1,49 @@
...
+            Some(
+                remote_base_path
+                    + &dest_path.split("configs/").collect::<Vec<&str>>()[1]
+                        .split('/')
+                        .collect::<Vec<&str>>()
+                        .join("/")
+                    + &file.name,
+            )
+        })
+        .unwrap();
```
tdelabro, line 36: Wild unwrap. Do error management instead

tdelabro: or rather use `unwrap_or_else` instead

crates/node/src/constants.rs
```diff
@@ -1,46 +1,2 @@
...
-    "upgradable.json",
-    "cairo_1/erc20.casm.json",
-    "cairo_1/erc20.sierra.json",
-    "cairo_1/HelloStarknet.casm.json",
-    "cairo_1/HelloStarknet.sierra.json",
-    "cairo_1/NoValidateAccount.casm.json",
-    "cairo_1/NoValidateAccount.sierra.json",
-];
+pub const DEFAULT_CONFIGS_URL: &str =
+    "https://raw.githubusercontent.com/d-roak/madara/feat/configs-index/configs/index.json";
```
tdelabro, line 2: Will it be hosted on you github workspace lastingly?

tdelabro: Or is it just for development?

crates/pallets/starknet/src/genesis_loader.rs
```diff
@@ -56,7 +56,7 @@ impl<T: crate::Config> From<GenesisLoader> for GenesisConfig<T> {
                 match class {
                     ContractClass::Path { path, version } => {
                         let contract_path = match loader.madara_path.clone() {
-                            Some(madara_path) => madara_path + "/" + &path,
+                            Some(madara_path) => madara_path + "/configs/" + &path,
                             None => {
                                 let project_path = utils::get_project_path().expect("Project path not found");
```
tdelabro, line 61: A better expect message could be: "A project path should be present for the node to start"

https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

crates/pallets/starknet/src/utils.rs
```diff
@@ -11,6 +11,14 @@ impl From<Error> for sc_cli::Error {
     }
 }
 
+impl From<Error> for String {
```
tdelabro, line 14: `Error` should rather implement `Display`

tdelabro: It's a more rustic way

tdelabro (changes requested):

crates/node/src/constants.rs
d-roak: for the CI, once we merge I'll do a PR to replace `constants.rs` and `index.json`, to point to the official repo

crates/node/src/cli.rs
```diff
@@ -36,17 +36,38 @@ pub struct ExtendedRunCmd {
...
+    /// There is an alias for madara_path <> base_path <> tmp
+    /// Default: ~/.madara
     #[clap(long)]
     pub madara_path: Option<PathBuf>,
 
+    /// Choose a supported testnet chain which will load some default values
+    /// current supported testnets: sharingan
     #[clap(long)]
-    pub chain_spec_url: Option<String>,
+    pub testnet: Option<Testnet>,
```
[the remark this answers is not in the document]
d-roak: This is not new in this PR, I just reordered the flags. Testnets can have default configs, but currently it just sets the file that should be retrieved with the chainspec (an sets some substrate flags)

crates/node/src/cli.rs
```diff
@@ -36,17 +36,38 @@ pub struct ExtendedRunCmd {
...
+    #[clap(long)]
+    pub disable_madara_configs: bool,
+
+    /// Disable automatic url fetching for madara config files
+    #[clap(long)]
+    pub disable_url_fetch: bool,
 
+    /// Path to the folder where all configuration files and data are stored
+    /// There is an alias for madara_path <> base_path <> tmp
+    /// Default: ~/.madara
```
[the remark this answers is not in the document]
d-roak: No, madara-path is always setted

crates/node/src/cli.rs
```diff
@@ -36,17 +36,38 @@ pub struct ExtendedRunCmd {
...
-    pub testnet: Option<Testnet>,
+    pub chain_spec_url: Option<String>,
+
+    /// Load a custom index.json file for downloading custom assets
+    /// Check documentation in
+    /// https://github.com/keep-starknet-strange/madara/blob/main/docs/configs.md
+    #[clap(long)]
+    pub configs_url: Option<String>,
+
+    /// Disable madara default configs
```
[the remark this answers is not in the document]
d-roak: Kakarot has an use case where they don't use our flags and put directly the files in the madara path. In order for us not to overwrite anything, or load unnecessary stuff on madara path, I created that flag

crates/node/src/command.rs
d-roak: That is not new. You need to set an Cors enum that doesn't have public visibility

crates/node/src/command.rs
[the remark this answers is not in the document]
d-roak: We have been using that architecture for some time now. We can't merge both arguments. We fully manage the Madara path and we know when and how it is set up. The same doesn't apply to base_path. That argument is managed by substrate and can change during the program's runtime (e.g. when you set the `-tmp`) flag. Wdyt?

crates/node/src/command.rs
d-roak: No, I'm wrong

crates/node/src/command.rs
d-roak: We had that since the --testnet flag was created. The reasoning was for authority node operators to run their nodes with a command that we had set, between iterations. So they would always run `madara --testnet sharingan`, and if we needed to enable some extra stuff, we could without an issue. We can remove that, if u want

crates/node/src/command.rs
d-roak: both madara_configs and chain_specs use that function to fetch files from the URL and add them in the madara path

crates/node/src/command.rs
d-roak: the configs can be manually added in the madara path, or they can be already there from a previous run

crates/node/src/command.rs
d-roak: get_project_path will only work in the case we are in a dev environment where we have the cargo project setted. We only look if it's erroring, and if so, we assume that we are not in a dev environment, so we shouldn't try to copy the files

d-roak: Agree with the inverse

Why use a for each? I will need to convert to an iterator before and it will not handle easily the errors

crates/node/src/command.rs
d-roak: Recheck the logic, pls

crates/pallets/starknet/src/utils.rs
d-roak: I think I'm doing this in the wrong way. Even if i implement `Display`, the compiler still throws an error that I'm missing `String`

crates/pallets/starknet/src/genesis_loader.rs
d-roak: That message may be misleading. What about "Project path not found while trying to load the genesis contracts"

crates/node/src/command.rs
tdelabro: so maybe just `fetch_from_url_and_write_to_disk`

crates/pallets/starknet/src/genesis_loader.rs
tdelabro: "A Project path shoud be present in order to load the genesis contracts"

crates/node/src/constants.rs
tdelabro: Great

crates/node/src/command.rs
tdelabro: It sound usefull. We can keep it, but we have to document it heavily and flag it as TO: remove when going in production

crates/node/src/command.rs
tdelabro: We are in the function `fetch_madara_configs`. It's not clear that this function should only be called in a dev env.

I think you should handle the error

crates/node/src/cli.rs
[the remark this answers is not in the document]
tdelabro: I think it's a missuesed of clap.
 you can do:

```rust
#[clap(long, default_value="~/.madara"})]
pub madara_path: PathBuf,
```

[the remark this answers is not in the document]
tdelabro: Or some variation around this in order to support windows too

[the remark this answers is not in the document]
tdelabro: Same things for the place I told you "if you always set it to Some in the end, use a default value in the CLI". That what I meant

crates/node/src/command.rs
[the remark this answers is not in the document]
tdelabro: But if they are always the same. Can't we just get rid of madara_path and just use base_path for everything instead?

crates/node/src/command.rs
tdelabro:
```rust
if let (Some(Testnet::Sharingan), false) = (cli.run.testnet, cli.run.disable_madara_configs) {
   if let Some(ref src_path) = local_path {
     ...
   } else if !cli.run.disable_url_fetch {
     ...
   }
}
```

This is the same logic you wrote. I find it more readable this way.
Is this the logic you want this time?

tdelabro: A lot of this logic only is relevant when we are running a sharingan testnet.
Can we extract it in a separated function?
I think it will make the code much clearer

```rust
if let (Some(Testnet::Sharingan) = cli.run.testnet {
   set_sharingan_chain_specs(cli);
}
```

crates/pallets/starknet/src/utils.rs
tdelabro: The true fix will be to use anyhow rather than `String` as the error type

crates/node/src/command.rs
tdelabro: Oh you can go with a for loop. Especially if there are errors to handle. You are right!

crates/node/src/cli.rs
```diff
@@ -36,17 +36,41 @@ pub struct ExtendedRunCmd {
...
+
+    /// Disable madara default configs:
+    /// - Fetching index.json, genesis.json and genesis assets
+    /// - Fetching default chain specs
+    #[clap(long)]
+    pub disable_madara_configs: bool,
+
+    /// Disable automatic url fetching for madara config files
+    #[clap(long)]
+    pub disable_url_fetch: bool,
```
tdelabro, line 57: Hey, I have been thinking and I would rather prefer to disable fetch URL being the default.
By default do nothing. It seems more logic.

And then you can pass `--fetch-chain-specs` or  `--fetch-chain-specs=<your_url>`.
The first one will use our default value, the second the one specified by the user.
If you don't pass it, it will be set to None.

To do so use https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_missing_value
```rust
 #[arg(long, num_args = 0..=1, default_missing_value("my_default_url"))]
pub fetch-chain-specs: Option<String>,
```

tdelabro: Same for disable_madara_config.

crates/node/src/cli.rs
```diff
@@ -36,17 +36,41 @@ pub struct ExtendedRunCmd {
...
+    /// current supported testnets: sharingan
     #[clap(long)]
-    pub chain_spec_url: Option<String>,
+    pub testnet: Option<Testnet>,
 
+    /// If the files currently exist in your madara_path, the default behaviour will skip the file
+    /// fetching, it's possible to force an update with this flag
     #[clap(long)]
-    pub genesis_url: Option<String>,
+    pub update_configs: bool,
```
tdelabro, line 73: rename to `force_update_config`

crates/node/src/command.rs
```diff
@@ -52,69 +52,133 @@ impl SubstrateCli for Cli {
...
+    match (cli.run.madara_path.clone(), cli.run.run_cmd.shared_params.base_path.clone()) {
         (Some(madara_path), _) => {
-            cli.run.run_cmd.shared_params.base_path = Some(madara_path.clone());
-            madara_path.to_str().unwrap().to_string()
+            cli.run.run_cmd.shared_params.base_path = Some(madara_path);
         }
         (_, Some(base_path)) => {
-            cli.run.madara_path = Some(base_path.clone());
-            base_path.to_str().unwrap().to_string()
+            cli.run.madara_path = Some(base_path);
```
tdelabro, line 61: And if as you told "madara_path is always set" because it has a default value of `~/.madara`, we won't ever enter this arm.

Just do:
```rust
cli.run.run_cmd.shared_params.base_path = cli.run.madara_path.clone();
```

crates/pallets/starknet/src/utils.rs
tdelabro: The true fix would be to use anyhow instead of string as your error type

docs/configs.md
```diff
@@ -0,0 +1,54 @@
...
+  retrieve configuration files
+- `--madara-path`: Changes the default [madara-path](#madara-path) (default:
+  `$HOME/.madara`)
+- `--node-key-file`: Key responsible for assigning the peer_id (default:
+  `madara-path/p2p-key.ed25519`)
+- `--testnet`: Allows you to join an official testnet, currently only Sharingan
+  is supported
+- `--update-configs`: By default, if configuration files are already downloaded
+  into `--madara-path`, the client will not try to overwrite them. With this
+  flag, you can force an update
```
tdelabro, line 36: This will be hard to maintain as the CLI evolve.
Remove it and focus on having a top-notch in-code `clap` documentation, that will appear when you run `.madara --help`.

d-roak: I don't mind deleting it and adding to clap. But it's usual to have people not knowing what flags are available and what is the behavior of the flags we changed (we can't change the clap docs for those). Imo it's better to have the documentation for it

tdelabro: The problem is that it will 100% get outdated at some point. While clap `--help` will always be up to date.
And every program in the planet know that he should run `--help` to get the flags of any binary

crates/node/src/cli.rs
```diff
@@ -37,52 +47,45 @@ pub struct ExtendedRunCmd {
...
+    /// If you want to load a chain spec that is present in your filesystem, use `--chain=<PATH>`
+    #[clap(long, conflicts_with = "testnet")]
+    pub fetch_chain_spec: Option<String>,
 
     /// Path to the folder where all configuration files and data are stored
-    /// There is an alias for madara_path <-> base_path
-    /// Default: ~/.madara
-    #[clap(long)]
+    /// base_path will always be overwritten by madara_path
+    /// in the case you use the --tmp, the base_path will be changed during the runtime
```
tdelabro, line 56: the `--tmp` flag

tdelabro: `he base_path will be changed during the runtime` to what?

crates/node/src/command.rs
```diff
@@ -83,41 +66,24 @@ fn set_dev_environment(cli: &mut Cli) {
...
+            utils::fetch_from_url(
+                constants::SHARINGAN_CHAIN_SPEC_URL.to_string(),
+                madara_path.clone() + "/configs/chain-specs/",
+            )?;
             cli.run.run_cmd.shared_params.chain = Some(madara_path + "/chain-specs/testnet-sharingan-raw.json");
         }
-        _ => {}
     }
 
     if cli.run.run_cmd.shared_params.chain.is_some() {
```
tdelabro, line 89: Is this compatible with `Testnet` can you have both `--testnet` and `--chain`?

crates/node/src/command.rs
```diff
@@ -83,41 +66,24 @@ fn set_dev_environment(cli: &mut Cli) {
     cli.run.run_cmd.rpc_methods = RpcMethods::Unsafe;
 }
 
-fn set_chain_spec(cli: &mut Cli) -> Result<(), String> {
+fn set_testnet(cli: &mut Cli, testnet: Testnet) -> Result<(), String> {
     // checks if it should retrieve and enable a specific chain-spec
     let madara_path = cli.run.madara_path.clone().expect("Failed retrieving madara_path").to_str().unwrap().to_string();
     let local_path = utils::get_project_path();
```
tdelabro, line 73: In all arms testnet is required to to be `Testnet::Sharingan`

```rust
if testsnet == Testnet::Sharingan {
   if let Some(ref src_path) = local_path {
   
   } else {

   }
}

File: crates/node/src/command.rs [hunk 29]
```diff
@@ -128,45 +94,50 @@ fn set_chain_spec(cli: &mut Cli) -> Result<(), String> {
...
-        let src_path = src_path.clone() + "/configs/index.json";
-        utils::copy_from_filesystem(src_path, madara_path.clone() + "/configs", cli.run.update_configs)?;
-    } else if !cli.run.disable_url_fetch {
-        utils::fetch_from_url(
-            constants::DEFAULT_CONFIGS_URL.to_string(),
-            madara_path.clone() + "/configs",
-            cli.run.update_configs,
-        )?;
-    }
+    match (cli.setup.fetch_madara_configs.clone(), local_path) {
```
tdelabro, line 111: you are never using the two matched variable in conjunction.


```rust
if let Ok(ref src_path) = local_path {

} else if let Some(config_url) = cli.setup.fetch_madara_configs {

}
```

tdelabro, line 111: I think you can probably use a ref of madara_configs rather than clone it

crates/node/src/command.rs
```diff
@@ -313,12 +277,27 @@ pub fn run() -> sc_cli::Result<()> {
                     Some((madara_path + "/p2p-key.ed25519").into());
             }
 
-            set_chain_spec(&mut cli)?;
+            if cli.run.run_cmd.shared_params.dev {
+                set_dev_environment(&mut cli);
+            }
+
+            if let Some(chain_spec_url) = cli.run.fetch_chain_spec.clone() {
```
tdelabro, line 284: Do you have to clone? Cant `set_chain_spec` take a ref?

docs/configs.md
```diff
@@ -1,54 +0,0 @@
-# Configurations
```
tdelabro: Some additional doc is probably needed somewhere in order to explain that we now have a `setup` command.
Some examples may need to change too. Maybe in the README.md

tdelabro (changes requested):

crates/node/src/cli.rs
d-roak: non deterministic

crates/node/src/command.rs
d-roak: you right

crates/node/src/command.rs
d-roak: It becomes conflicts with the mut cli there

crates/pallets/starknet/src/utils.rs
d-roak: Can we leave that refactor to another PR? (I'll open an issue for that if u agree)

crates/node/src/command.rs
```diff
@@ -37,7 +37,7 @@ impl SubstrateCli for Cli {
     fn load_spec(&self, id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
         Ok(match id {
             "dev" => {
-                let enable_manual_seal = self.sealing.map(|_| true);
+                let enable_manual_seal = self.run.sealing.map(|_| true);
```
tdelabro, line 40: sealing is of type
```rust
#[derive(Debug, Copy, Clone, clap::ValueEnum, Default)]
pub enum Sealing {
    // Seal using rpc method.
    #[default]
    Manual,
    // Seal when transaction is executed.
    Instant,
}
```

Why do we ignore the `Instant` variant and make it a manual sealing anyway?

crates/node/src/command.rs
```diff
@@ -56,6 +56,7 @@ impl SubstrateCli for Cli {
 }
 
 fn set_dev_environment(cli: &mut Cli) {
+    println!("Setting up dev environment");
```
tdelabro, line 59: debug log?

d-roak: okay, this was not supposed to exist at all, one of my debugging probably. I will delete it

crates/node/src/command.rs
d-roak: Always have been like this. I just adapted it to work with our current code. @EvolveArt might be able to answer that question

tdelabro: @EvolveArt do you know something?

crates/node/src/command.rs
```diff
@@ -55,71 +55,97 @@ impl SubstrateCli for Cli {
...
+
+    // we can't set `--rpc-cors=all`, so it needs to be set manually if we want to connect with external
+    // hosts
+    cli.run.run_cmd.rpc_external = true;
+    cli.run.run_cmd.rpc_methods = RpcMethods::Unsafe;
+}
+
+fn set_testnet(cli: &mut Cli, testnet: Testnet) -> Result<(), String> {
+    // checks if it should retrieve and enable a specific chain-spec
+    let madara_path = cli.run.madara_path.clone().expect("Failed retrieving madara_path").to_str().unwrap().to_string();
```
tdelabro, line 74: Why is it safe to unwrap?
Why can we `expect` it?

tdelabro: Also recommended message  format for `expect` is defined here: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

crates/node/src/command.rs
```diff
@@ -55,71 +55,97 @@ impl SubstrateCli for Cli {
...
+    if cli.run.run_cmd.shared_params.chain.is_some() {
+        cli.run.run_cmd.rpc_external = true;
+        cli.run.run_cmd.rpc_methods = RpcMethods::Unsafe;
     }
 
+    Ok(())
+}
+
+fn set_chain_spec(cli: &mut Cli, chain_spec_url: String) -> Result<(), String> {
+    let madara_path = cli.run.madara_path.clone().expect("Failed retrieving madara_path").to_str().unwrap().to_string();
```
tdelabro, line 100: Same thing

crates/node/src/command.rs
```diff
@@ -55,71 +55,97 @@ impl SubstrateCli for Cli {
...
+
+    utils::fetch_from_url(chain_spec_url.clone(), madara_path.clone() + "/chain-specs")?;
+    let chain_spec = chain_spec_url.split('/').last().expect("Chain spec file name not found");
+    cli.run.run_cmd.shared_params.chain = Some(madara_path + "/chain-specs/" + chain_spec);
+
+    Ok(())
+}
+
+fn fetch_madara_configs(cli: &Cli) -> Result<(), String> {
+    let madara_path = cli.run.madara_path.clone().expect("Failed retrieving madara_path").to_str().unwrap().to_string();
```
tdelabro, line 110: This exact code appears three time, maybe it could be encapsulated in its own function
https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

crates/node/src/command.rs
```diff
@@ -55,71 +55,97 @@ impl SubstrateCli for Cli {
...
+    let local_path = utils::get_project_path();
+
+    if let Ok(ref src_path) = local_path {
+        let index_path = src_path.clone() + "/configs/index.json";
+        utils::copy_from_filesystem(index_path, madara_path.clone() + "/configs")?;
+
+        let madara_configs: configs::Configs =
+            serde_json::from_str(&utils::read_file_to_string(madara_path.clone() + "/configs/index.json")?)
+                .expect("Failed to load madara configs");
```
tdelabro, line 119: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

crates/node/src/command.rs
```diff
@@ -55,71 +55,97 @@ impl SubstrateCli for Cli {
...
+        for asset in madara_configs.genesis_assets {
+            let src_path = src_path.clone() + "/configs/genesis-assets/" + &asset.name;
+            utils::copy_from_filesystem(src_path, madara_path.clone() + "/configs/genesis-assets")?;
+        }
+    } else if let Some(configs_url) = &cli.setup.fetch_madara_configs {
+        utils::fetch_from_url(configs_url.to_string(), madara_path.clone() + "/configs")?;
+
+        let madara_configs: configs::Configs =
+            serde_json::from_str(&utils::read_file_to_string(madara_path.clone() + "/configs/index.json")?)
+                .expect("Failed to load madara configs");
```
tdelabro, line 129: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

crates/node/src/command.rs
```diff
@@ -238,42 +264,26 @@ pub fn run() -> sc_cli::Result<()> {
...
-                cli.run.run_cmd.force_authoring = true;
-                cli.run.run_cmd.alice = true;
-
-                // we can't set `--rpc-cors=all`, so it needs to be set manually if we want to connect with external
-                // hosts
-                cli.run.run_cmd.rpc_external = true;
-                cli.run.run_cmd.rpc_methods = RpcMethods::Unsafe;
+        Some(Subcommand::Run(cmd)) => {
+            let madara_path =
+                cli.run.madara_path.clone().expect("Failed retrieving madara_path").to_str().unwrap().to_string();
```
tdelabro, line 269: Again same code

crates/node/src/command.rs
tdelabro: `set_chain_spec` is only used once, with an input that comes form the `cli` iteslf. 

So, `chain_spec_url` has no need to be passed as an argument.

I would change this func signatrue so it only takes `cli` as arg and does the if some check internaly

crates/node/src/command.rs
```diff
@@ -238,42 +264,26 @@ pub fn run() -> sc_cli::Result<()> {
...
-                    Some(madara_path.clone() + "/chain-specs/testnet-sharingan-raw.json");
+            if let Some(chain_spec_url) = cli.run.fetch_chain_spec.clone() {
+                set_chain_spec(&mut cli, chain_spec_url)?;
+            }
 
-                // This should go apply to all testnets when applying a match pattern
-                cli.run.run_cmd.rpc_external = true;
-                cli.run.run_cmd.rpc_methods = RpcMethods::Unsafe;
+            if let Some(testnet) = cli.run.testnet {
+                set_testnet(&mut cli, testnet)?;
```
tdelabro, line 286: Same approach can probaly be used for `set_testnet`


```rust
fn set_testnet(cli: &mut Cli) -> Result<(), String> {
if let Some(testnet) = cli.run.testnet {
<do your thing>
}

}

Comment (tdelabro) [reply crates/node/src/command.rs:286]: Maybe with a change in fn name, `try_to_set_testnet` or smth

File: crates/pallets/starknet/src/utils.rs (hunk 18 above)
Comment (tdelabro) [reply crates/pallets/starknet/src/utils.rs:14]: Sure

Comment (tdelabro) [reply crates/pallets/starknet/src/utils.rs:14]: Don't forget to open the issue

File: crates/pallets/starknet/src/utils.rs [hunk 41]
```diff
@@ -70,12 +78,13 @@ pub fn copy_from_filesystem(src_path: String, dest_path: String) -> Result<(), E
 
 pub fn fetch_from_url(target: String, dest_path: String) -> Result<(), Error> {
     log::info!("Trying to fetch {} to {} from url", target, dest_path);
-    let dst = std::path::PathBuf::from(dest_path);
+    let mut dst = std::path::PathBuf::from(dest_path);
     std::fs::create_dir_all(&dst)?;
+    dst.push(target.split('/').last().expect("File name not found"));
```
tdelabro, line 83: Expect message standard

tdelabro (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Next release
 
+- ci: disable pr close workflow
```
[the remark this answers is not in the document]
d-roak: current approach not working

configs/index.json
```diff
@@ -0,0 +1,57 @@
+{
+  "remote_base_path": "https://raw.githubusercontent.com/d-roak/madara/feat/configs-index/configs/",
+  "genesis_assets": [
+    {
+      "name": "Account.json",
+      "md5": "bbdfc1b69f76e9c2784db6d8b697dc56"
```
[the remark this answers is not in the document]
d-roak: hash of the contents of the whole file. Easier to check on CI

crates/node/src/command.rs
d-roak: because the value is always a string (is it not?)

crates/node/src/command.rs
d-roak: Is it worth having a function for return the madara path as a string?

configs/index.json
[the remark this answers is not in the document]
tdelabro: But two different files (indentation, spacing, order of the keys) can lead to two different md5 hashes, while still representing the same contract.
Won't cause a problem?

What exactly are we checking in the CI?

tdelabro (changes requested):

configs/index.json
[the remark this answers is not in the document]
d-roak: That the files that we are retrieving in the URL match the md5 that is in the `index.json` (simple security validation). Since you can define custom url's for each file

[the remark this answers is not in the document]
d-roak: Answering the first question, that's no problem. We want to verify the integrity of the file, not if it's contract X or Y

crates/node/src/command.rs
d-roak: applied the `into_os_string` approach with an `expect` error msg. Seems cleaner now

crates/node/src/command.rs
EvolveArt: Seems to be linked to this [todo](https://github.com/keep-starknet-strange/madara/blob/main/crates/node/src/service.rs#L319)

README.md
```diff
@@ -141,16 +141,6 @@ production environment, the relative changes and trends over time are what we
 focus on. This way, we can ensure that Madara is always improving, and that we
 maintain a high standard of performance as the project evolves.
 
-One can use flamegraph-rs to generate flamegraphs and look for the performance
```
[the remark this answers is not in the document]
d-roak: I'll re-add it here. It didn't stopped working

EvolveArt (approved):

tdelabro (approved):

PR #1031, opened 2023-08-22
Feat/estimate fee correct version

tdelabro:
# Pull Request type

- Bugfix
- Feature
- Build-related changes
- Testing
- Other (please describe):

Resolves: #NA

## What is the new behavior?

-
-
-


crates/client/rpc/src/errors.rs
```diff
@@ -19,28 +19,30 @@ pub enum StarknetRpcApiError {
...
     #[error("Invalid contract class")]
     InvalidContractClass = 50,
-    #[error("Too many storage keys requested")]
-    ProofLimitExceeded = 10000,
-    #[error("Too many keys provided in a filter")]
-    TooManyKeysInFilter = 34,
+    #[error("Class already declared")]
+    ClassAlreadyDeclared = 51,
+    #[error("The transaction version is not supported")]
+    UnsupportedTxVersion = 61,
```
tdelabro, line 39: Did you choose those numbers yourself?

crates/client/rpc/src/lib.rs
```diff
@@ -528,9 +528,17 @@ where
         request: Vec<BroadcastedTransaction>,
         block_id: BlockId,
     ) -> RpcResult<Vec<FeeEstimate>> {
-        // TODO:
-        //      - modify BroadcastedTransaction to assert versions == "0x100000000000000000000000000000001"
-        //      - to ensure broadcasted query signatures aren't valid on mainnet
+        let smthng = request.iter().any(|tx| match tx {
```
tdelabro, line 531: `smthng` is not a good name

crates/client/rpc/src/lib.rs
```diff
@@ -528,9 +528,17 @@ where
...
+            BroadcastedTransaction::Invoke(BroadcastedInvokeTransaction::V0(tx_v0)) => tx_v0.is_query,
+            BroadcastedTransaction::Invoke(BroadcastedInvokeTransaction::V1(tx_v1)) => tx_v1.is_query,
+            BroadcastedTransaction::Declare(BroadcastedDeclareTransaction::V1(tx_v1)) => tx_v1.is_query,
+            BroadcastedTransaction::Declare(BroadcastedDeclareTransaction::V2(tx_v2)) => tx_v2.is_query,
+            BroadcastedTransaction::DeployAccount(deploy_tx) => deploy_tx.is_query,
+        });
+
+        if smthng {
+            return Err(StarknetRpcApiError::UnsupportedTxVersion.into());
+        }
```
tdelabro, line 541: Why are those not supported for `estimate_fee`?

crates/client/rpc/src/errors.rs
gaetbout: Comes from:
https://github.com/starkware-libs/starknet-specs/blob/0e859ff905795f789f1dfd6f7340cdaf5015acc8/api/starknet_write_api.json#L272

crates/client/rpc/src/lib.rs
gaetbout: What do you mean, isn't the point of the issue to block/prevent all the one that aren't query_version?

gaetbout: These are the cases:
 - Deploying: `__validate_deploy__` should ensure tx version is  2**128 + 1
 - Invoke: `__validate_` should ensure tx version is  2**128 + 1
 - Declaring: `__validate_declare__` should ensure tx version is  2**128 + 2
 
I think both BroadcastedInvokeTransaction::V0 and BroadcastedDeclareTransaction::V1 shouldn't even be supported 
so the another error should be thrown for those?

gaetbout: ref:
https://docs.starknet.io/documentation/starknet_versions/version_notes/#:~:text=New%20transaction%20version%20(version%201)%20for%20invoke%20and%20declare%20transactions%3A

gaetbout: https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/Blocks/transactions/

gaetbout: Still working on this, so just ignore it at this point :)

gaetbout: GM @tdelabro , @apoorvsadana ,
As stated in the documentation 
https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/Blocks/transactions/
Invoke v0 and declared v1 are deprecated and will be removed.
Should those already return some different error?
Or should this be part of another issue for when they will effectively be removed?

crates/client/rpc/src/errors.rs
tdelabro: maybe mention it in a comment

Issue #1038, opened 2023-08-24
bug: ci is failing due to the current cache mechanism


d-roak:
Improve the caching mechanism in the ci

PR #1046, opened 2023-08-28
fix: CI cache

kasteph:
# Pull Request type

- CI

Resolves: #1038 

## What is the new behavior?

- Add prefix to the Cargo lockfile path

No.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -27,7 +27,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           fail-on-cache-miss: true
       - name: Run benchmark
         run: |
@@ -67,7 +67,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           fail-on-cache-miss: true
       - name: Setup rust toolchain
         if: steps.cache.outputs.cache-hit != 'true'
--- a/.github/workflows/linters-cargo.yml
+++ b/.github/workflows/linters-cargo.yml
@@ -20,7 +20,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           fail-on-cache-miss: true
       - name: Setup build deps
         run: |
--- a/.github/workflows/madara-commands.yml
+++ b/.github/workflows/madara-commands.yml
@@ -18,7 +18,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           fail-on-cache-miss: true
       - name: Create build-spec (plain)
         run: |
--- a/.github/workflows/rust-build.yml
+++ b/.github/workflows/rust-build.yml
@@ -21,7 +21,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Setup rust toolchain
         if: steps.cache.outputs.cache-hit != 'true'
--- a/.github/workflows/tests-coverage.yml
+++ b/.github/workflows/tests-coverage.yml
@@ -29,7 +29,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           fail-on-cache-miss: true
       - name: Run test (native)
         run: |-
@@ -53,7 +53,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
           fail-on-cache-miss: true
       - name: Setup build deps
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - ci: don't enforce changelog on PR's with label `dependencies`
 - feat: rebase of core deps and 0.12.1
 - chore: remove global state root
+- ci: fix cache
 
 ## v0.1.0
 


kasteph:
Hi @d-roak, thanks for the feedback! My first instinct was to think that there might have been an issue with the matching. However, when I printed to `stdout` the values of `hashFile(Cargo.lock)`  and `hashFile(**/Cargo.lock)` it was the same since there's only one lockfile in the project. (ref: [hashFile doc](https://docs.github.com/en/actions/learn-github-actions/expressions#hashfiles))

To be honest, this was a bit of a stab in the dark as I wasn't sure what improve the caching mechanism exactly meant in the original issue. There's a few things I can think of, off the top of my head:

- a looser cache retrieval strategy
   - having another list of keys to restore stale cache if there is a miss with `restore-keys`
- a short lived cache 
   - add commit hash to cache key so that even if one is working on the same branch, it will invalidate the cache

Let me know which direction you'd like me to go on and I'd be happy to keep working on this!

d-roak:
Ping me on tg (d_roak) to discuss some thoughts on this

kasteph:
Checking the list of caches, it looks like my branch is using the key `Linux-cargo-f4c36fe519218b6992218cc91d236c23d53548dc424cffd8280f11d207e381d4` which makes sense since the Cargo.lockfile has not changed and if there's a cache miss on the branch, it will check the base branch of my branch (which in this case is `main`).

Below, one can see that the same key exists for this cache from PR 980 and now on `main` which could be the source for various PRs potentially using stale builds.

```
❯ gh actions-cache list -R git@github.com:keep-starknet-strange/madara
Total caches size 9.72 GB

Showing 7 of 7 cache entries in keep-starknet-strange/madara

node-cache-Linux-npm-876ef5908887861cffefac81a13234821af02450f33c402f1bb43e0aef854169  30.14 MB  refs/heads/main       16 minutes ago
Linux-cargo-f4c36fe519218b6992218cc91d236c23d53548dc424cffd8280f11d207e381d4           1.93 GB   refs/heads/main       16 minutes ago
node-cache-Linux-npm-71e37eef1dac7f0b420fab5d31bbacf4be45547a63de6d3264bef6111978704b  30.11 MB  refs/pull/1018/merge  3 hours ago
Linux-cargo-342744392c550a83e271e910a661d83838f1640d0a20e7f4f17e39bd89b20764           1.93 GB   refs/pull/1018/merge  3 hours ago
Linux-cargo-e292a7f4a73880ce990a50da3dc3b366237f8b153aafc00065b02e112ec6f5fc           1.94 GB   refs/pull/992/merge   a day ago
Linux-cargo-f4c36fe519218b6992218cc91d236c23d53548dc424cffd8280f11d207e381d4           1.93 GB   refs/pull/980/merge   4 days ago
Linux-cargo-9e8d279ba728b0ceb759fe7cdf9ce55513af2388e3481e972034b3cf9e41e0cc           1.93 GB   refs/pull/992/merge   4 days ago
```

.github/workflows/clean-cache.yml
d-roak, line 1: This probably won't work without a personal token. And personal tokens will not work on this workflow (github security policies)

.github/workflows/benchmarks.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
d-roak, line 30: With the previous considerations of us having a single `Cargo.lock`, why should we keep this change? (this applies to every `**/Cargo.lock` change)

.github/workflows/rust-build.yml
```diff
@@ -21,7 +20,10 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          # scope the cache by branch s.t. it doesn't fall back on base branch if there's a cache miss
+          # see: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
+          key:
+            ${{ runner.os }}-cargo-${{ github.ref_name }}-${{ hashFiles('**/Cargo.lock') }}
```
d-roak, line 26: The force update I had in mind was an update with the same key. I already saw that it is not possible. Let's try to use some approach for matching keys like u suggested
ref: https://github.com/actions/cache/blob/main/tips-and-workarounds.md#update-a-cache

The one in this commit does not look into the main cache (it should)

d-roak (changes requested):

.github/workflows/rust-build.yml
```diff
@@ -21,7 +20,11 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
```
d-roak, line 27: shouldn't we start from the `${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{ github.run_id }}` until `${{ runner.os }}-cargo`?

d-roak (changes requested):
Pls add the restore-keys to all the places where we are retrieving the cargo cache

.github/workflows/rust-build.yml
```diff
@@ -21,7 +20,12 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
+          fail-on-cache-miss: true
+          restore-keys: |
+            ${{ runner.os }}-cargo
```
d-roak, line 28: match restore-keys from the highest level to the lowest 
check this: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#comparing-artifacts-and-dependency-caching

.github/workflows/tests-coverage.yml
```diff
@@ -29,8 +29,12 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
           fail-on-cache-miss: true
+          restore-keys: |
+            ${{ runner.os }}-cargo
```
d-roak, line 37: exact match (we want to execute on the build of the run)

.github/workflows/tests-coverage.yml
```diff
@@ -53,8 +57,12 @@ jobs:
...
```
d-roak, line 65: exact match (we want to execute on the build of the run)

.github/workflows/madara-commands.yml
```diff
@@ -18,8 +18,12 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
           fail-on-cache-miss: true
+          restore-keys: |
+            ${{ runner.os }}-cargo
```
d-roak, line 26: exact match (we want to execute on the build of the run)

.github/workflows/linters-cargo.yml
```diff
@@ -20,8 +20,12 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
           fail-on-cache-miss: true
+          restore-keys: |
+            ${{ runner.os }}-cargo
```
d-roak, line 28: exact match (we want to execute on the build of the run)

.github/workflows/benchmarks.yml
```diff
@@ -67,8 +71,12 @@ jobs:
             ~/.cargo/registry/cache/
             ~/.cargo/git/db/
             target/
-          key: ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
           fail-on-cache-miss: true
+          restore-keys: |
+            ${{ runner.os }}-cargo
```
d-roak, line 79: exact match (we want to execute on the build of the run)

.github/workflows/benchmarks.yml
```diff
@@ -27,8 +27,12 @@ jobs:
...
```
d-roak, line 35: exact match (we want to execute on the build of the run)

d-roak (changes requested):
Sry for the delay

.github/workflows/benchmarks.yml
kasteph: Do I understand correctly that you'd like multiple restore keys?

```suggestion
     ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
            github.run_id }}
     ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
     ${{ runner.os }}-cargo
```

d-roak: on the build yes, on the retrieve no

d-roak (approved):
lgtm

Issue #904, opened 2023-07-25
Tracking issue for E2E tests


tdelabro:
Currently, our integration tests are written in JS and we struggle to maintain them. We want to move over to rust.

# Working branch
In the first time of developing this test suite, there will be a lot of changes, from many different contributors at the same time.
We don't want to disrupt the main branch flow until we have something robust to merge. Therefore, in a first time, we will be working on this branch: https://github.com/keep-starknet-strange/madara/tree/feature/e2e-rust-tests.
All contributors' PR should be merged in this one.

# Roadmap:
- [x] Create a `test/` folder on the workspace root
- [x] add it in the `[test]` section of the workspace
- [x] in it create a crate that will contain the utility methods, mostly wrappers around `starknet-rs`
- [x] create a test file by RPC route and fill it with our new tests
- [x] remove js tests and clean up the related tooling

## Required lib functionalities
- [x] send rpc call
- [x] send Txs
- [x] seal block
- [x] jump N blocks

### Skiped RPC routes
- syncing

### Special cases

> Starknet currently supports calls about a [pending block](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/transaction-life-cycle/#the_pending_block), the block being built. Madara doesn't, so any request about `BlockId::Tag(BlockTag::Pending))` should return an error 

### Routes to test

<details>
<summary>blockNumber</summary>

#### Rust impl
```rust
fn block_number(&self) -> RpcResult<u64>;
```

#### Things to test
- [x] Return 0 at block 0
- [x] Return the correct block number after creating 1, 10, 100, 1000, 1,000,000 (only if it does not take too long to execute) empty blocks
</details>

<details>
<summary>blockHashAndNumber</summary>

#### Rust impl
```rust
 fn block_hash_and_number(&self) -> RpcResult<BlockHashAndNumber>;
```

#### Things to test
- [x] Return the correct block number at blocks and hash at block 0
- [x]  Return the correct block number at blocks and hash after an empty block is created
- [x]  Return the correct block number at blocks and hash after diverse blocks with diverse tx are created
</details>

<details>
<summary>getBlockTransactionCount</summary>
#### Rust impl
```rust
fn get_block_transaction_count(&self, block_id: BlockId) -> RpcResult<u128>;
```

#### Things to test
- [x] Return the correct number of tx for empty blocks
- [x]  Return the correct number of tx in a block with multiple tx
- [x] Return an error for a nonexisting block
</details>

<details>
[...]

PR #1053, opened 2023-08-30
Feature/e2e rust get nonce

tchataigner:
Add test cases in rust for getNonce route.

Parent issue: #904 

# Pull Request type

- Testing

Resolves: #NA

## What is the new behavior?

In this draft state it contains working behavior for:
- [x] Return an error for a nonexisting block
- [x] Return a nonce 0 for a non-used contract address
- [x] Return a nonce 0 if the contract deployed at this address is not an account contract
- [x] Otherwise, return the nonce of the account deployed at this address

No

## Other information

The tested behavior differs from #904 checklist but after some discussions the one implemented is the correct one.


tchataigner:
Hey @tdelabro, I am currently facing some behavior that goes against what was mentioned in #904. 

Currently when I test for errors on non-used contract address or non-account contract I do not get errors but an `Ok()` with a nonce 0.

Which one is the correct one? The one from the issue or the one I'm experiencing?

You should be able to test that by running the `fail_non_used_contract_address` and `fail_non_account_contract` tests.

tdelabro:
@tchataigner Hey,
I think this has been done in order to respect Starknet current behavior. Any non-initialized memory slot of a contract will return 0 if read.
So nothing to fix, the behaviour I described in the issue was incorrect. 
You can put the PR as ready it that was the only thing bothering you :)

tchataigner:
Alright ! Thanks, will clean up the code, fix tests and I'll set it as ready.

starknet-rpc-test/get_nonce.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    assert_eq!(
+        rpc.get_nonce(
+            BlockId::Number(0),
+            FieldElement::from_hex_be("0x4269DEADBEEF").expect("Invalid Contract Address")
+        )
+        .await
+        .ok(),
+        Some(
+            FieldElement::from_hex_be("0x0000000000000000000000000000000000000000000000000000000000000000")
```
tdelabro, line 58: `FieldElement::ZERO` instead

starknet-rpc-test/get_nonce.rs
```diff
@@ -0,0 +1,115 @@
...
+            argent_account_address,
+            FieldElement::from_hex_be(MINT_AMOUNT).expect("Invalid Mint Amount"),
+            None,
+        )])
+        .await?;
+
+    assert_eq!(
+        rpc.get_nonce(BlockId::Tag(BlockTag::Latest), account.address(),).await.ok(),
+        Some(
+            FieldElement::from_hex_be("0x0000000000000000000000000000000000000000000000000000000000000001")
```
tdelabro, line 109: `FieldElement::ONE`

tdelabro (changes requested):

tdelabro (approved):

PR #1055, opened 2023-08-30
Starknet call tests

apoorvsadana:
- adds test cases for starknet_call
- moves mutex lock acquisition to the right place
- moves account creation function to utils
- makes create_block_with_txs generic to accept Invoke and Declare transactions


starknet-rpc-test/call.rs
```diff
@@ -0,0 +1,232 @@
...
+use starknet_core::types::contract::SierraClass;
+use starknet_core::types::{BlockId, BlockTag, FunctionCall, StarknetError};
+use starknet_core::utils::get_selector_from_name;
+use starknet_ff::FieldElement;
+use starknet_providers::{MaybeUnknownErrorCode, Provider, ProviderError, StarknetErrorWithMessage};
+use starknet_rpc_test::constants::{ARGENT_CONTRACT_ADDRESS, FEE_TOKEN_ADDRESS, SIGNER_PRIVATE};
+use starknet_rpc_test::utils::create_account;
+use starknet_rpc_test::{ExecutionStrategy, MadaraClient, Transaction};
+
+#[fixture]
```
tdelabro, line 19: I think a previous PR moved this fixture in a fixture.rs file. You will be able to use it if you rebase

starknet-rpc-test/call.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    madara.create_empty_block().await?;
+    let account = create_account(rpc, SIGNER_PRIVATE, ARGENT_CONTRACT_ADDRESS);
+
+    let contract_artifact: SierraClass = serde_json::from_reader(
+        std::fs::File::open(env!("CARGO_MANIFEST_DIR").to_owned() + "/contracts/HelloStarknet.sierra.json").unwrap(),
+    )
+    .unwrap();
+
+    let declaration = account.declare(
```
tdelabro, line 172: can you rename it to `declare_tx` or any name that make it clear it is a transaction

starknet-rpc-test/call.rs
```diff
@@ -0,0 +1,232 @@
...
+        contract_artifact.clone().flatten().unwrap().into(),
+        FieldElement::from_hex_be("0xdf4d3042eec107abe704619f13d92bbe01a58029311b7a1886b23dcbb4ea87").unwrap(), // compiled class hash
+    );
+    let contract_factory = ContractFactory::new(contract_artifact.class_hash().unwrap(), account.clone());
+
+    let mut deployment = contract_factory.deploy(vec![], FieldElement::ZERO, true);
```
tdelabro, line 178: Same here, make it clear it is a transaction object

starknet-rpc-test/call.rs
```diff
@@ -0,0 +1,232 @@
...
+    // manually incrementing nonce else as both declare and deploy are in the same block
+    // so automatic nonce calculation will fail
+    let nonce = rpc.get_nonce(BlockId::Tag(BlockTag::Latest), account.address()).await.unwrap();
+    deployment = deployment.nonce(nonce + FieldElement::ONE);
+
+    // declare and deploy contract
+    madara
+        .create_block_with_txs(vec![Transaction::Declaration(declaration), Transaction::Execution(deployment)])
+        .await?;
+    // madara.create_block_with_txs(vec![Transaction::Execution(deployment)]).await?;
```
tdelabro, line 193: Forgotten comment

starknet-rpc-test/contracts/HelloStarknet.cairo
```diff
@@ -0,0 +1,19 @@
+#[starknet::contract]
+mod HelloStarknet {
```
tdelabro, line 2: Not a gib fan of this contract name and file name.
Can you try something more descriptive: `Counter` or `Balance`

starknet-rpc-test/get_block_transaction_count.rs
```diff
@@ -57,14 +57,13 @@ async fn work_ok_with_block_one_tx(#[future] madara: MadaraClient) -> Result<(),
     let signer = LocalWallet::from(SigningKey::from_secret_scalar(FieldElement::from_hex_be(SIGNER_PRIVATE).unwrap()));
     let argent_account_address = FieldElement::from_hex_be(ARGENT_CONTRACT_ADDRESS).expect("Invalid Contract Address");
     let account = SingleOwnerAccount::new(rpc, signer, argent_account_address, chain_id::TESTNET);
+    let execution = account.transfer_tokens(
```
tdelabro, line 60: Same `token_transfer_tx`

starknet-rpc-test/get_block_hash_and_number.rs
```diff
@@ -27,17 +24,15 @@ async fn work_ok_at_start_and_with_new_blocks() -> Result<(), anyhow::Error> {
...
-    madara
-        .create_block_with_txs(vec![account.transfer_tokens(
-            argent_account_address,
-            FieldElement::from_hex_be(MINT_AMOUNT).expect("Invalid Mint Amount"),
-            None,
-        )])
-        .await?;
+    let account = create_account(rpc, SIGNER_PRIVATE, ARGENT_CONTRACT_ADDRESS);
+
+    let execution = Transaction::Execution(account.transfer_tokens(
```
tdelabro, line 29: Same, change name to `transfer_tokens_tx`

starknet-rpc-test/src/lib.rs
```diff
@@ -128,6 +155,9 @@ impl MadaraClient {
     }
 
     pub async fn new(execution: ExecutionStrategy) -> Self {
+        // we keep the reference, otherwise the mutex unlocks immediately
+        let _mutex_guard = FREE_PORT_ATTRIBUTION_MUTEX.lock().await;
+
```
tdelabro, line 160: Out of curiosity, why was the first impl not working?

tdelabro (changes requested):

starknet-rpc-test/call.rs
apoorvsadana: fixed for all tests

starknet-rpc-test/call.rs
apoorvsadana: fixed

starknet-rpc-test/call.rs
apoorvsadana: fixed

starknet-rpc-test/call.rs
apoorvsadana: fixed

starknet-rpc-test/contracts/HelloStarknet.cairo
apoorvsadana: fixed

starknet-rpc-test/get_block_transaction_count.rs
apoorvsadana: fixed

starknet-rpc-test/get_block_hash_and_number.rs
apoorvsadana: fixed

starknet-rpc-test/src/lib.rs
apoorvsadana: So previously, we acquired the lock during the process of finding the port and released the lock when we found a free port. We didn't wait till the current thread launched the madara instance and acquired this free port. So what would happen is

1. Thread 1 took the lock
2. Thread 1 found out port 1025 is open
3. Thread 1 released the lock
4. Thread 1 fired up the madara instance
5. Thread 2 started look for ports and found 1025 open as well (thread 1 is still starting the madara instance so it hasn't acquired port 1025 yet)
6. Thread 1 starts on port 1025. Thread 2 madara instance starts on some other port but it makes calls to 1025 only thinking that's the instance assigned to it.
7. If Thread 1 test cases gets over before Thread 2, it will close the Madara instance and Thread 2 will fail.

So the fix was just to release the lock once Madara is fired up and has acquired the port.

starknet-rpc-test/call.rs
tdelabro: Bless you 🙏

starknet-rpc-test/src/lib.rs
tdelabro: Thx

starknet-rpc-test/call.rs
```diff
@@ -165,36 +161,33 @@ async fn works_on_mutable_call_without_modifying_storage(#[future] madara: Madar
...
     // manually setting fee else estimate_fee will be called and it will fail
     // as contract is not declared yet (declared in the same block as deployment)
-    deployment = deployment.max_fee(FieldElement::from_hex_be("0x1000000000").unwrap());
+    deploy_tx = deploy_tx.max_fee(FieldElement::from_hex_be("0x1000000000").unwrap());
 
     // manually incrementing nonce else as both declare and deploy are in the same block
     // so automatic nonce calculation will fail
     let nonce = rpc.get_nonce(BlockId::Tag(BlockTag::Latest), account.address()).await.unwrap();
-    deployment = deployment.nonce(nonce + FieldElement::ONE);
+    deploy_tx = deploy_tx.nonce(nonce + FieldElement::ONE);
```
tdelabro, line 183: Can't you chain those:

```rust
let nonce = rpc.get_nonce(BlockId::Tag(BlockTag::Latest), account.address()).await.unwrap();

let deploy_tx = contract_factory.deploy(vec![], FieldElement::ZERO, true)
     // manually setting fee else estimate_fee will be called and it will fail
    // as contract is not declared yet (declared in the same block as deployment
    .max_fee(FieldElement::from_hex_be("0x1000000000").unwrap())
    // manually incrementing nonce else as both declare and deploy are in the same block
    // so automatic nonce calculation will fail
   .nonce(nonce + FieldElement::ONE);

````

tdelabro: I think it's a more common usage of the builder pattern

tdelabro (changes requested):

starknet-rpc-test/call.rs
tdelabro: And I check your other usages of this `contract_factory`, maybe it can be done in a similar way there too

apoorvsadana: yep, makes sense! fixed

tdelabro (approved):

PR #1060, opened 2023-09-02
Initial gas fix

apoorvsadana:
- fixes incorrect initial_gas for transactions which would cause transactions to go through in the blockifier even if enough gas was not present
- fixed a bug where fees wouldn't be charged if max_fee was 0

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,7 +18,7 @@
       // As such, this path would have to change accordingly.
       "program": "${workspaceFolder}/target/debug/madara",
       // If you wish to supply arguments/parameters to the program, supply them below:
-      "args": ["--dev", "--sealing=instant"],
+      "args": ["--dev", "--sealing=instant", "--tmp", "--execution=Native"],
       // Working folder for execution. Change as necessary if program requires a different value:
       "cwd": "${workspaceFolder}",
       "terminal": "integrated",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - chore: cairo-contracts compilation scripts & docs are updated, cairo_0
   contracts recompiled
 - chore: rebase of core deps and 0.12.1
+- fix: initial_gas set to max_fee and fixed fee not being charged when max_fee=0
 
 ## v0.1.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - fix: estimate_fee should make sure all transaction have a version being
   2^128 + 1 or 2^128+2 depending on the tx type
+- fix: initial_gas set to max_fee and fixed fee not being charged when max_fee=0
 
 ## v0.2.0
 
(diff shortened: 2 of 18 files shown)


crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -429,7 +429,7 @@ impl Transaction {
         };
 
         // FIXME 710
-        let mut initial_gas = super::constants::INITIAL_GAS;
+        let mut initial_gas = self.max_fee.try_into().unwrap(); // unwrap safe as >u64 case will be handled at client side
```
d-roak, line 432: shouldn't it be independent of the client?

.vscode/launch.json
```diff
@@ -18,7 +18,7 @@
       // As such, this path would have to change accordingly.
       "program": "${workspaceFolder}/target/debug/madara",
       // If you wish to supply arguments/parameters to the program, supply them below:
-      "args": ["--dev", "--sealing=instant"],
+      "args": ["--dev", "--sealing=instant", "--tmp", "--execution=Native"],
```
d-roak, line 21: why are u using native execution?

crates/primitives/starknet/src/transaction/mod.rs
```diff
@@ -555,8 +555,11 @@ impl Transaction {
         // Verify the transaction version.
         self.verify_tx_version(&tx_type)?;
 
-        // FIXME 710
-        let mut initial_gas = super::constants::INITIAL_GAS;
+        // if it's an estimate fee then use max initial_gas
+        let mut initial_gas = match self.is_query {
+            true => u64::MAX,
+            false => self.max_fee.try_into().unwrap(), // unwrap safe as >u64 case will be handled at client side
```
d-roak, line 561: same as above

.vscode/launch.json
apoorvsadana: this is for launch.json which is used when using the vs code debugger. the debugger won't work if the execution isn't native

crates/primitives/starknet/src/transaction/mod.rs
apoorvsadana: what should be independent of the client?

d-roak: the safeness (or not) of the unwrap

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 - chore: cairo-contracts compilation scripts & docs are updated, cairo_0
   contracts recompiled
 - chore: rebase of core deps and 0.12.1
+- fix: initial_gas set to max_fee and fixed fee not being charged when max_fee=0
```
tdelabro, line 32: Wrong spot. It goes under `Next release`

tdelabro (approved):

crates/primitives/starknet/src/transaction/mod.rs
apoorvsadana: makes sense, fixed

crates/primitives/starknet/src/transaction/mod.rs
apoorvsadana: fixed

CHANGELOG.md
apoorvsadana: fixed

PR #1068, opened 2023-09-04
upgrade: bump rustc nightly version to 1.74 date

d-roak:
to fix #1061 problems


tdelabro (approved):
Why are we using the nightly toolchain? DO you remember?

PR #1075, opened 2023-09-06
Pending tx and get tx by hash

apoorvsadana:


starknet-rpc-test/get_transaction_by_hash.rs
```diff
@@ -0,0 +1,51 @@
...
+    Ok(())
+}
+
+#[rstest]
+#[tokio::test]
+async fn fail_invalid_transaction_hash(#[future] madara: MadaraClient) -> Result<(), anyhow::Error> {
+    let madara = madara.await;
+    let rpc = madara.get_starknet_client();
+
+    assert!(rpc.get_transaction_by_hash(FieldElement::ZERO).await.is_err());
```
EvolveArt, line 48: cannot we have a more precise match here ?

starknet-rpc-test/get_transaction_by_hash.rs
```diff
@@ -0,0 +1,51 @@
...
+        .await;
+
+    let rpc_response = match txs.unwrap().remove(0).unwrap() {
+        TransactionResult::Execution(rpc_response) => rpc_response,
+        _ => panic!("expected execution result"),
+    };
+
+    // there is a delay between the transaction being available at the client
+    // and the sealing of the block, hence sleeping for 100ms
+    tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
EvolveArt, line 33: That's quite unexpected behavior, I'm not sure I understand why this is happening

starknet-rpc-test/pending_transactions.rs
```diff
@@ -0,0 +1,42 @@
...
+    let rpc = madara.get_starknet_client();
+
+    let pending_txs = rpc.pending_transactions().await?;
+
+    // not validating the fields inside the transaction as
+    // that is covered in get_block_with_txs
+    assert_eq!(pending_txs.len(), 0);
+
+    Ok(())
+}
```
EvolveArt, line 42: please add another test with 100+ pending txs

starknet-rpc-test/src/utils.rs
```diff
@@ -67,18 +94,18 @@ pub trait AccountActions {
 }
 
 impl AccountActions for SingleOwnerAccount<&JsonRpcClient<HttpTransport>, LocalWallet> {
-    fn transfer_tokens(
+    fn transfer_tokens_u256(
         &self,
         recipient: FieldElement,
-        transfer_amount: FieldElement,
+        transfer_amount: [FieldElement; 2],
```
EvolveArt, line 100: let's create a U256 type

EvolveArt (changes requested):
left few comments

starknet-rpc-test/get_transaction_by_hash.rs
apoorvsadana: fixed

starknet-rpc-test/pending_transactions.rs
apoorvsadana: fixed

starknet-rpc-test/src/utils.rs
apoorvsadana: done

starknet-rpc-test/get_transaction_by_hash.rs
apoorvsadana: So what I found from debugging is that `block_hash_from_transaction_hash` inside `mapping_db` returns None if you call it instantly after adding a transaction. And the db only writes the hashes when `sync_block` is called, which is called as a part of `mapping_sync_worker`. So, while `engine_createBlock` returns 200 it takes a few milli seconds for the blocks to be synced on the client db which is why we need to wait.

EvolveArt: Interesting, we didn't have this behavior within typescript tests. I suggest we move this to the create_block util then as it's quite related

apoorvsadana: The delay is actually very small. Like, even if I add a print statement instead of a sleep, it still works. However, since this time would be machine dependent I added around 100ms.

I was thinking of adding it inside create_block but I think that will slow all other tests as well which is why I added it here. wdyt?

apoorvsadana: Hmm, makes sense, will check how to do this

starknet-rpc-test/add_declare_transaction.rs
```diff
@@ -0,0 +1,154 @@
...
+    let (declare_tx, _, _) =
+        account.declare_contract("./contracts/Counter.sierra.json", "./contracts/Counter.casm.json");
+
+    let mut txs = madara.create_block_with_txs(vec![Transaction::Declaration(declare_tx)]).await?;
+
+    assert_eq!(txs.len(), 1);
+
+    let invoke_tx_result = txs.remove(0);
+    assert_matches!(
+        invoke_tx_result.err(),
```
tdelabro, line 33: replace by `txs[0].unwrap_err()`
It won't panic we can go full unwrap on it

tdelabro: And if it does, the test fails, which is what we want

starknet-rpc-test/add_declare_transaction.rs
```diff
@@ -0,0 +1,154 @@
...
+    assert_eq!(txs.len(), 1);
+    let declare_tx_result = txs.remove(0);
+    assert!(declare_tx_result.is_ok());
+
+    // transaction failed during execution, no change in storage
+    assert!(rpc.get_class(BlockId::Tag(BlockTag::Latest), expected_class_hash).await.is_err());
+
+    // doesn't get included in block
+    let included_txs = rpc.get_block_transaction_count(BlockId::Tag(BlockTag::Latest)).await?;
+    assert_eq!(included_txs, 0);
```
tdelabro, line 79: That's for cairo0 tx. Should we get a cairo1 test too? Where it revert but is included in the block

starknet-rpc-test/add_declare_transaction.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    let account = create_account(rpc, SIGNER_PRIVATE, ARGENT_CONTRACT_ADDRESS, true);
+    let (declare_tx, expected_class_hash, _) =
+        account.declare_contract("./contracts/Counter.sierra.json", "./contracts/Counter.casm.json");
+
+    let mut txs = madara.create_block_with_txs(vec![Transaction::Declaration(declare_tx)]).await?;
+
+    assert_eq!(txs.len(), 1);
+    let declare_tx_result = txs.remove(0);
+    match declare_tx_result {
```
tdelabro, line 98: `txs[0]`

starknet-rpc-test/add_declare_transaction.rs
```diff
@@ -0,0 +1,154 @@
...
+        Ok(TransactionResult::Declaration(DeclareTransactionResult { transaction_hash, class_hash })) => {
+            assert_eq!(
+                transaction_hash,
+                FieldElement::from_hex_be("0x01fc4c0d8f82edfd74ef83c5db42203fe4a70243a76e88e0a4a6ade9753d8ec9")
```
tdelabro, line 102: Do we have a way to compute it from `declare_tx`? Rather than having it hardcoded?

starknet-rpc-test/add_declare_transaction.rs
```diff
@@ -0,0 +1,154 @@
...
+    let declare_tx_result = txs.remove(0);
+    assert!(declare_tx_result.is_ok());
+
+    // second declaration fails
+    let (declare_tx, _, _) =
+        account.declare_contract("./contracts/Counter.sierra.json", "./contracts/Counter.casm.json");
+
+    let mut txs = madara.create_block_with_txs(vec![Transaction::Declaration(declare_tx)]).await?;
+    assert_eq!(txs.len(), 1);
+    let declare_tx_result = txs.remove(0);
```
tdelabro, line 142: `txs[0]`

starknet-rpc-test/add_declare_transaction.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    // first declaration works
+    let account = create_account(rpc, SIGNER_PRIVATE, ARGENT_CONTRACT_ADDRESS, true);
+    let (declare_tx, _, _) =
+        account.declare_contract("./contracts/Counter.sierra.json", "./contracts/Counter.casm.json");
+
+    let mut txs = madara.create_block_with_txs(vec![Transaction::Declaration(declare_tx)]).await?;
+
+    assert_eq!(txs.len(), 1);
+    let declare_tx_result = txs.remove(0);
```
tdelabro, line 133: txs[0]

tdelabro: no need to make it `mut`

starknet-rpc-test/add_deploy_account_transaction.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[rstest]
+#[tokio::test]
+async fn fail_execution_step_with_no_storage_change(#[future] madara: MadaraClient) -> Result<(), anyhow::Error> {
+    let madara = madara.await;
+    let rpc = madara.get_starknet_client();
+
+    // deploy account
+    let oz_factory =
+        build_oz_account_factory(rpc, "0x123", FieldElement::from_hex_be(CAIRO_1_ACCOUNT_CONTRACT_CLASS_HASH).unwrap())
```
tdelabro, line 23: Isn't "0x123" also an invalid private key. Could it fail because of that rather than the lack of fees?

tdelabro (changes requested):
Gotta go I will finish the review later ;)

starknet-rpc-test/add_deploy_account_transaction.rs
tdelabro: My question is does it fail because of lack of fee, or because of an invalid signature before this?

tdelabro: But maybe you are right and `0x123` is a valid signature. But earlier we have this tests
```rust
async fn fail_validation_step(#[future] madara: MadaraClient) -> Result<(), anyhow::Error> {
    let madara = madara.await;
    let rpc = madara.get_starknet_client();

    // using incorrect private key to generate the wrong signature
    let account = create_account(rpc, "0x1234", ARGENT_CONTRACT_ADDRESS, true);
``` 

Which make me think you signature is not valid in this test too, and we fail before the fee thing, just because of bad signature

apoorvsadana: Right. Actually "0x123" is misleading here, I get what you meant now. So in the context of the Argent Account, 0x2 will be an incorrect key but since DeployAccount doesn't have any validation (it happens during the contract run I beleive?), entering any private key here will pass the validation step. I guess I will use the original private key to avoid this confusion.

tdelabro: I think you can use whatever, but just mention why it is ok in a comment :)

starknet-rpc-test/add_declare_transaction.rs
tdelabro: My (personal) opinion is that we don't really care about this kind of perf optimization for tests. The main focus should be the readability of the tests, so there is no mistake possible on what is happening and what it guarantees.
The less complexity, the less distraction the better. Technical complexity should be removed in order for the attention to focus on the test purpose.

So I would be totally ok to add an extra `.clone` if it allows for a more straightforward impl. 

What do you think?

starknet-rpc-test/add_declare_transaction.rs
tdelabro: I think we can let things as they are :)

starknet-rpc-test/add_declare_transaction.rs
apoorvsadana: I understand your point about just focussing on readability for test cases, but what's the issue with `txs.remove(0)`?

tdelabro: Remove gives you an `Option`. `[0]` is equivalent to `.remove(0).unwrap()` and it's ok because you know it is there.
Also `txs` does not have to be mutable`.

Imo it makes the intention clear by striping those unnecessary impl details.

But it's nothing truly important. If it causes other problems we can keep it

apoorvsadana: Since I am using `remove` I need to make txs mutable. In this specific case, I could use `txs[0].as_ref` and clone the transaction hash when comparing. However, in some tests, I am comparing a lot of fields and will need to clone all of them to compare, so I feel having a `mut` with `remove` will be more readable. So I am keeping it for now, let me know though if you want me to change is particularly in some places.

starknet-rpc-test/add_deploy_account_transaction.rs
apoorvsadana: I have fixed this but it's in the new PR, sorry!

starknet-rpc-test/add_declare_transaction.rs
tdelabro: That's good to me

starknet-rpc-test/get_transaction_by_hash.rs
tdelabro: Maybe a less platform-dependent thing would be to retry every 100ms, for 2 sec. And fail the test if after two sec it still returns an err.

apoorvsadana: Ya I couldn't find any way to wait for the syncing either. @EvolveArt if you also agree, I can implement what @tdelabro mentioned

apoorvsadana: @tdelabro, I discussed with @EvolveArt and have implemented the polling solution

Issue #1049, opened 2023-08-29
dev: Use actuall contract class hash for the genesis ones


tdelabro:
As you can see here: https://github.com/keep-starknet-strange/madara/blob/main/docs/genesis.md#contract-classes-node-genesis, and in the code, some of the contracts we deploy during the genesis step of the chain have funky ClassHash (`0x1000`, `0x90000`, etc).

Those are not the result of a regular hash computation based on their class code. We set those arbitrary values manually because it was easier at the time.

This had to change. Your mission will be to compute the actual class of those contracts, in the canon starknet way, and modify code and documentation in order to use those new, actual class hashes

ayushtom:
I can pick this

PR #1079, opened 2023-09-09
chore: update class hash in genesis config

ayushtom:
# Pull Request type

Chore: added actual class hashes for contracts which had dummy values.

- Bugfix

Currently we add dummy values from certain contracts listed below - 
- UniversalDeployer.json
- test.json
- erc20.json
- erc721.json
Here the values were mocked. We add real values generated from starkli for the contracts and add these in the genesis config

Resolves: #1049 

## What is the new behavior?

- We add the class hashes for contracts with the help of starkli
- We add this from starkli since this also matches with the declared contract class hash on goerli

No


ayushtom:
closes #1049

ayushtom:
@tdelabro can you review these changes?

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Madara Changelog
 
+- chore: add real class hash values for genesis config
```
tdelabro, line 3: Goes under the `next release` section

tdelabro (changes requested):
I thought those values also exit as hardcoded constants somewhere in the Rust code. Maybe I'm wrong, can you double check tho

CHANGELOG.md
ayushtom: done

ayushtom:
> I thought those values also exit as hardcoded constants somewhere in the Rust code. Maybe I'm wrong, can you double check tho

I think i see that rust code normally serialises this json and reads value from there. These values were not hardcoded in rust code implementation.

ayushtom:
> I thought those values also exit as hardcoded constants somewhere in the Rust code. Maybe I'm wrong, can you double check tho

I think i see that rust code normally serialises this json and reads value from there. These values were not hardcoded in rust code implementation. @tdelabro

tdelabro (approved):

PR #1083, opened 2023-09-10
RPC cache system

### LINKED ISSUE #947: dev: add LRU cache layer to block fetching client-side
```rust
/// Manage LRU caches for block data and their transaction statuses.
/// These are large and take a lot of time to fetch from the database.
/// Storing them in an LRU cache will allow to reduce database accesses
/// when many subsequent requests are related to the same blocks.
pub struct EthBlockDataCacheTask<B: BlockT>(mpsc::Sender<EthBlockDataCacheMessage<B>>);
```

This code was taken from [frontier](https://github.com/paritytech/frontier/blob/194f6bb06152402ba44b340c3d401ae6e0670d96/client/rpc/src/eth/cache/mod.rs#L76) and is extensively used on their RPC layer. We should investigate adding this on our side and evaluate the performance gain using https://github.com/keep-starknet-strange/gomu-gomu-no-gatling

tchataigner:
# Pull Request type

- Feature

In the current implementation of the madara client, we fetch block data by going through the generated logs of our pallet. As the block data is important and we need fast retrieval, it seems important to:
- Properly store the block data in a dedicated `StorageValue` in our pallet
- When possible retrieve the block data through storage query instead of digest API
- Add a cache layer on top of it to ease retrieval of recurrent queried data

Resolves: #947

## What is the new behavior?

Right now, the update to the code base are as follow:
- Added a `CurrentBlock` type in the starknet pallet, representing a `StorageValue` containing our block data. Also added a getter to retrieve it through our runtime API.
- Added necessary method over our storage implementation to both retrieve the data through the runtime API and directly from our storage.
- Implemented a base LRUCache limited by allocated memory, along with an asynchronous task responsible for the handling of the cache and retrieving data.
- Updated the necessary calls in our RPC implementation, to use the cache task instead of directly retrieving the block data through our logs.
-  Added an argument over the run command to set the allocated memory size limit for the cache.

The new flow is:
1. Run a madara run while specifying `--starknet-log-block-cache-size`. For now, the default value is 200 (most likely will cache little if not nothing)
2. On RPC call, we will send the requested block hash to our data cache task.
3. If the data cache has the block, returns it.
4. Otherwise, we will directly query our storage for it.

No

## Other information

Original issue: #947


tchataigner:
Hey @tdelabro, here is a first draft for the RPC cache. 

As we discussed under the issue, I've implemented the cache to work with allocated memory size instead of values size to make it more precise.

Right now, it seems to fail integration test, which is strange as it passes on my computer. I'll look into it.

The only problem that I have is that I think the benchmark tests to evaluate the cache efficiency are not there. I wanted to add some to try and measure it but I'm having trouble in identifying how to pull that off.

 Could I ask for your input on that ? Ideally, I think that we want to try and spam a request on loop to our RPC methods that concern block fetch (e.g.: `getblockWithTxHashes`)

crates/client/rpc/src/cache/lru_cache.rs
```diff
@@ -0,0 +1,141 @@
...
+    #[test]
+    fn test_work_allocated_size_limit() {
+        let mut cache = LRUCache::new("name", 150, None);
+        // Allocated memory will keep being 148 bytes for the three first insert.
+        assert!(cache.insert(1, "a"));
+        let memory_usage_step_1 = cache.inner.memory_usage();
+        assert!(cache.insert(2, "b"));
+        assert!(cache.insert(3, "c"));
+        // Memory allocation will try to reach 280 bytes, blocked by our limiter. Insertion should be
+        // succesful but first item should be removed from cache.
```
EvolveArt, line 134:
```suggestion
        // successful but first item should be removed from cache.
```

crates/client/rpc/src/cache/lru_cache.rs
```diff
@@ -0,0 +1,141 @@
...
+                    format!("Allocated memory size of starknet {} data cache.", cache_name),
+                )?,
+                registry,
+            )?,
+        })
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
EvolveArt, line 115: is there a way we can add tests for the metrics too?

crates/client/rpc/src/cache/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Represents the metadata of the block we are looking for.
+struct BlockMetadata<B: BlockT> {
+    hash: B::Hash,
+    schema: StarknetStorageSchemaVersion,
+}
+
+/// Represent a pair of a cache and a data waitlist, used to communicate information between our
+/// internal methods.
+struct CacheWaitlist<B: BlockT, T> {
+    cache: LRUCache<<B as BlockT>::Hash, T>,
```
EvolveArt, line 43: not sure we need to use this hasher here tho, we should probably optimize for performance here w/ blake2

EvolveArt (changes requested):
left few comments

tdelabro:

PR #1108, opened 2023-09-18
ci: add coverage

d-roak:
Testing #1122 while re adding coverage


tdelabro:
Do we really want to add it now? 
Don't it take insanely long to run?

d-roak:
> Do we really want to add it now? Don't it take insanely long to run?

This PR is a draft to test how we can do it. We need to re add the tests for sure, doesn't mean that it has to be now. We can still merge PR's (we need to be careful with the reviews though) @tdelabro

d-roak:
closed in favor of #1145

PR #1125, opened 2023-09-20
chore: add simulate tx version offset tests

### LINKED ISSUE #1115: test: make sure simulated tx uses the correct offsetted version number
in `crates/primitives/starknet/src/transaction/execution.rs` add more tests here:

```rust
#[cfg(test)]
mod simulate_tx_offset {
    use starknet_ff::FieldElement;

    use super::SIMULATE_TX_VERSION_OFFSET;
    #[test]
    fn offset_is_correct() {
        assert_eq!(
            SIMULATE_TX_VERSION_OFFSET,
            FieldElement::from_hex_be("0x100000000000000000000000000000000").unwrap()
        );
    }

    // TODO: add test that check that each get_account_transaction_context impl correctly uses this
    // offset
}
```

petscheit:
Adds tests for checking the correct usage of the simulated tx version offset. 

Additionally, a minor refactoring, adds a trait used for applying the version offset for all different program paths.

These changes close #1115 

# Pull Request type

- Testing
- Refactoring

No

### DIFF
--- a/crates/primitives/transactions/src/execution.rs
+++ b/crates/primitives/transactions/src/execution.rs
@@ -55,15 +55,11 @@ pub trait GetAccountTransactionContext {
 }
 
 pub trait SimulateTxVersionOffset {
-    fn apply_simulate_tx_version_offset(
-        &self,
-    ) -> TransactionVersion;
+    fn apply_simulate_tx_version_offset(&self) -> TransactionVersion;
 }
 
 impl SimulateTxVersionOffset for TransactionVersion {
-    fn apply_simulate_tx_version_offset(
-        &self,
-    ) -> TransactionVersion {
+    fn apply_simulate_tx_version_offset(&self) -> TransactionVersion {
         Felt252Wrapper(Felt252Wrapper::from(self.0).0 + SIMULATE_TX_VERSION_OFFSET).into()
     }
 }
@@ -578,9 +574,25 @@ impl Execute for L1HandlerTransaction {
 
 #[cfg(test)]
 mod simulate_tx_offset {
+    use blockifier::execution::contract_class::ContractClass;
     use starknet_ff::FieldElement;
 
-    use super::SIMULATE_TX_VERSION_OFFSET;
+    use super::*;
+
+    // Applies the tx offset to a given version. Used only for testing
+    fn apply_simulate_tx_version_offset(version: TransactionVersion, is_query: bool) -> TransactionVersion {
+        if is_query {
+            return version.apply_simulate_tx_version_offset();
+        }
+
+        version
+    }
+
+    // Asserts that the offset is applied correctly to a given version
+    fn eval_simulate_tx_version_offset(original: TransactionVersion, received: TransactionVersion, is_query: bool) {
+        assert_eq!(received, apply_simulate_tx_version_offset(original, is_query));
+    }
+
     #[test]
     fn offset_is_correct() {
         assert_eq!(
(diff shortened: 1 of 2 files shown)


crates/primitives/starknet/src/transaction/execution.rs
```diff
@@ -575,6 +601,82 @@ mod simulate_tx_offset {
         );
     }
 
-    // TODO: add test that check that each get_account_transaction_context impl correctly uses this
-    // offset
+    #[test]
+    fn transaction_version_trait_correctly_applies_simulate_tx_version_offset() {
+        let v0 = TransactionVersion(StarkFelt::from(0u8));
+        assert_eq!(v0.clone().apply_simulate_tx_version_offset(), apply_simulate_tx_version_offset(v0.clone(), true));
```
tdelabro, line 607: This just test that the rust syntax `variable.function()` does the same this as the `function(variable)` one. 

This is not a test we want to keep

crates/primitives/starknet/src/transaction/execution.rs
```diff
@@ -575,6 +601,82 @@ mod simulate_tx_offset {
...
+            tx_hash: Default::default(),
+        };
+
+        let original_version = declare_tx.tx.version;
+
+        let queried_version = declare_tx.get_account_transaction_context(true).version;
+        eval_simulate_tx_version_offset(original_version.clone(), queried_version, true);
+
+        let non_queried_version = declare_tx.get_account_transaction_context(false).version;
+        eval_simulate_tx_version_offset(original_version.clone(), non_queried_version, false);
```
tdelabro, line 627: This is hard to understand.
`eval_simulate_tx_version_offset` make it hard to know that is actually tested.

```rust
let queried_version = declare_tx.get_account_transaction_context(true).version;
let non_queried_version = declare_tx.get_account_transaction_context(false).version;
assert_eq!(queried_version, non_queried_version + SIMULATE_TX_VERSION_OFFSET)
```

is much more straightforward and give the same guarantee

crates/primitives/starknet/src/transaction/execution.rs
```diff
@@ -575,6 +601,82 @@ mod simulate_tx_offset {
...
+    }
+
+    #[test]
+    fn deploy_account_transaction_correctly_applies_simulate_tx_version_offset() {
+        let declare_tx = DeployAccountTransaction {
```
tdelabro, line 632: wrong variable name

crates/primitives/starknet/src/transaction/execution.rs
tdelabro: same for the other tests :)

tdelabro (changes requested):

crates/primitives/starknet/src/transaction/execution.rs
petscheit: The above doesn't work, as we need to convert the version to a FieldElement first, before adding the offset. It got verbose and quite messy without impl a from trait, which I felt was overkill just for testing. I've worked on the function comment, making it clearer what happens. Let me know your thoughts on this.

tdelabro: `eval_simulate_tx_version_offset(original_version_v0, queried_version, true);`  is not expressive at all. 
 Reading this I have no idea what is done, what is asserted against what.
 
 I would much prefer a `assert_eq!(queried_version, Felt252Wrapper(Felt252::from(non_queried_version) + SIMULATE_TX_VERSION_OFFSET))`.
 It's a bit verbose I agree, but at least I undersand what is being tested (even if the newtypes shenanigans are a bit obsucre. And we can use intermediate value for even more readability.
 
 
 When writing test we don't aim for the most well abstracted code. We want to be able to understand in a fraction of second what is being tested. And often the best way to have this expressiveness is by writing things in a much verbose way than what we would have done for the actual code

petscheit: Yeah fair, impl as requested

crates/primitives/transactions/src/execution.rs
```diff
@@ -575,6 +587,85 @@ mod simulate_tx_offset {
...
+            Felt252Wrapper(Felt252Wrapper::from(original_version_v0.0).0 + SIMULATE_TX_VERSION_OFFSET).into()
+        );
+
+        let non_queried_version = declare_tx_v0.get_account_transaction_context(false).version;
+        assert_eq!(non_queried_version, original_version_v0);
+    }
+
+    #[test]
+    fn invoke_transaction_correctly_applies_simulate_tx_version_offset() {
+        let declare_tx_v0 = InvokeTransaction {
```
tdelabro, line 654: var `declare_tx_v0` is of type `InvokeTransaction`

tdelabro (changes requested):

crates/primitives/transactions/src/execution.rs
petscheit: Slipped through somehow. Fixed

Issue #1119, opened 2023-09-19
bug: Bad input from starknet js declare transaction


lana-shanghai:
# Bug Report

**Madara version:**

As of [commit](https://github.com/keep-starknet-strange/madara/pull/1118/commits/e800a950559e619ce7f32153e32a1b5f6985e6da) 

**Current behavior:**

Using starknet js, trying to use `account.declare` to deploy a contract. It requires to add the input `maxFee` of type BigInt, but the Madara runtime panics with 

```
2023-09-19 14:06:15 panicked at 'Bad input data provided to convert_transaction: Codec error', /Users/lana/github/madara/crates/runtime/src/lib.rs:115:1    
2023-09-19 14:06:15 Failed to convert transaction: Application(Execution(AbortedDueToTrap(MessageWithBacktrace { message: "wasm trap: wasm `unreachable` instruction executed", backtrace: Some(Backtrace { backtrace_string: "error while executing at wasm backtrace:\n    0: 0x25c919 - <unknown>!rust_begin_unwind\n    1: 0x15611e - <unknown>!core::panicking::panic_fmt::h68b51813cb7c9649\n    2: 0x1f1bf1 - <unknown>!ConvertTransactionRuntimeApi_convert_transaction" }) }))) 
```

snippet:

```
const account = new Account(provider, "0x00c1cf1490de1352865301bb8705143f3ef938f97fdf892f1090dcb5ac7bcd1d", "0x0000000000000000000000000000000000000000000000000000000000000002", "1");

const declareResult = await account.declare(
    {
        contract: sierra,
        casm: casm,
    },
    {
        nonce: 1,
        version: 2,
        maxFee: 100000,
    }
);
```

**Expected behavior:**

Create new scripts with starknet js to declare and deploy contracts on Madara. Last month it was working well. 
Find out if it is something on the Madara side of `convert_transaction` or on starknet js side.

**Steps to reproduce:**

Example [repo](https://github.com/lana-shanghai/madara_contract_scripts)
Run `npm install` and `node index.js`. This should show the last block finalized on Madara.
Test deployment of pre-built contracts:

`node declare_hello_starknet.js`

PR #1128, opened 2023-09-20
🐛 fix rpc tests and docs

EvolveArt:
# Pull Request type

X Bugfix
- Feature
- Build-related changes
X Documentation content changes
X Testing
- Other (please describe):

Resolves: #1119 

- Not enough tests
- Ignored tests for Cairo 1 contracts declaration 
- Outdated `rpc-contribution.md` docs

## What is the new behavior?

- Updates contribution doc
- Add more tests based on current issues and js scripts
- Remove `#[ignore]` on ignored tests

No


starknet-rpc-test/src/lib.rs
```diff
@@ -140,29 +140,46 @@ fn get_free_port() -> Result<u16, TestError> {
     Err(TestError::NoFreePorts)
 }
 
+fn get_free_p2p_port() -> Result<u16, TestError> {
```
d-roak, line 143: why do we need to set the p2p port? it's setted as a random one if the default is already occupied and we are not querying it

starknet-rpc-test/src/lib.rs
```diff
@@ -140,29 +140,46 @@ fn get_free_port() -> Result<u16, TestError> {
...
         let child_handle = Command::new("cargo")
 		// Silence Madara stdout and stderr
 		.stdout(Stdio::null())
-		.stderr(Stdio::null())
+		.stderr(if madara_log == "true" { Stdio::inherit() } else { Stdio::null() })
 		.args([
 			"run",
 			"--release",
 			"--",
 			"--alice",
```
d-roak, line 175: we can remove this, is already setted with `--dev`

starknet-rpc-test/src/lib.rs
```diff
@@ -140,29 +140,46 @@ fn get_free_port() -> Result<u16, TestError> {
...
 			"--sealing=manual",
 			&format!("--execution={execution}"),
-			"--chain=dev",
+			"--dev",
 			"--tmp",
-			&format!("--rpc-port={free_port}")
+			&format!("--port={free_p2p_port}"),
+			&format!("--rpc-port={free_port}"),
+            &format!("--madara-path=/tmp/{}", free_p2p_port)
```
d-roak, line 182: if we use this, we should remove the `--tmp` flag

.github/workflows/linters.yml
```diff
@@ -6,14 +6,6 @@ on:
   workflow_call:
 
 jobs:
-  prettier:
```
d-roak, line 9: why are we removing this?

starknet-rpc-test/src/lib.rs
```diff
@@ -129,45 +134,60 @@ impl Drop for MadaraClient {
...
 			"--",
-			"--alice",
 			"--sealing=manual",
 			&format!("--execution={execution}"),
-			"--chain=dev",
-			"--tmp",
-			&format!("--rpc-port={free_port}")
+			"--dev",
+            "--tmp",
+			&format!("--port={p2p_port}"),
```
d-roak, line 184: why are we setting the p2p port? do we need reproducibility on those?

EvolveArt: Yes and also in case the default is not available

d-roak (approved):

Issue #1077, opened 2023-09-07
dev: impl a braavos tx signer


tdelabro:
`given_contract_run_deploy_account_braavos_tx_works` is a test in `crates/pallets/starknet/src/tests/deploy_account_tx.rs`

It uses a Braavos account to sign and execute a tx.
Braavos has an unusual way of signing tx. It does not just sign the tx hash, it adds a bunch of stuff around it and then hash it all together and signs the output.

Here is how the Braavos wallet verify the signature: https://github.com/myBraavos/braavos-account-cairo/blob/develop/src/account/library.cairo#L328

Right now we hardcoded the correct value that must be signed for our tx. It would be better if we have a function that would compute this value on the fly.

Maybe some lib already exist that does that. Maybe you will have to code it all yourselve

tonypony220:
Can I take this?

tdelabro:
sure ! Wish you best luck !

tdelabro:
I think #1131 did. We will be merging it soon (waiting for another PR to be merged in main first)
@EvolveArt do you think we should keep waiting for Code Coverage?

PR #1131, opened 2023-09-20
added sign_message_hash_braavos function

tonypony220:
# Pull Request type

- Feature

Resolves: #1077 
added  function sign deloy tx for braavos

-
-
-

No


crates/pallets/starknet/src/tests/deploy_account_tx.rs
```diff
@@ -262,14 +262,12 @@ fn given_contract_run_deploy_account_braavos_tx_works() {
...
-        // Braavos has a complicated signature mecanism, they add stuffs around the tx_hash and then sign
-        // the whole thing. This hardcoded value is the expected "thing" that bravos expect you to
-        // sign for this transaction. Roll with it for now
-        let value_to_sign =
-            Felt252Wrapper::from_hex_be("0x06a8bb3d81c2ad23db93f01f72f987feac5210a95bc530eabb6abfaa5a769944").unwrap();
-        let mut signatures = sign_message_hash(value_to_sign);
-        signatures.extend_from_slice(&[Felt252Wrapper::ZERO; 8]);
-        deploy_tx.signature = signatures;
+        let tx_hash = deploy_tx.compute_hash::<<MockRuntime as Config>::SystemHash>(Starknet::chain_id(), false);
+        deploy_tx.signature = sign_message_hash_braavos::<<MockRuntime as Config>::SystemHash>(
```
EvolveArt, line 266: 2 small details:
1. We should probably not use the `SystemHash` here but directly the right hash for this braavos contract.
2. I prefer that we move the signature computation above the `deploy_tx` declaration with something like `let signature = sign...` then just fill it in the `deploy_tx` struct.

crates/pallets/starknet/src/tests/deploy_account_tx.rs
```diff
@@ -262,14 +262,12 @@ fn given_contract_run_deploy_account_braavos_tx_works() {
...
+            tx_hash,
+            Felt252Wrapper::ZERO,
+            &[Felt252Wrapper::ZERO; 7],
```
EvolveArt, line 269: could we add a test with real values if possible ?

EvolveArt (changes requested):
will merge once we have coverage back in CI

crates/pallets/starknet/src/tests/deploy_account_tx.rs
tonypony220: Could you please clarify your second suggestion? It seems like signing a hash before obtaining the hash itself might not be possible.

crates/pallets/starknet/src/tests/deploy_account_tx.rs
tonypony220: @EvolveArt Added hardcoded real signature in separate test.
I'm not sure if it's easy to add the secp256r1 curve, so we may do it in another issue if needed.

tdelabro (approved):

EvolveArt (approved):
lgtm, would be great to add more rustdoc/links to understand better the implementation

Issue #1117, opened 2023-09-18
feat: use actual vm ressources cost


tdelabro:
in ` crates/primitives/starknet/src/fees/mod.rs`, in ` calculate_l1_gas_by_vm_usage`

```rust
    let vm_resource_fee_costs = BTreeMap::from([
        (String::from("n_steps"), 1_f64),
        (String::from("pedersen_builtin"), 1_f64),
        (String::from("range_check_builtin"), 1_f64),
        (String::from("ecdsa_builtin"), 1_f64),
        (String::from("bitwise_builtin"), 1_f64),
        (String::from("poseidon_builtin"), 1_f64),
        (String::from("output_builtin"), 1_f64),
        (String::from("ec_op_builtin"), 1_f64),
    ]);
```

all values are set to `1`

I think we can use the values defined here instead: https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/fee-mechanism/#general_case

tdelabro:
@EvolveArt or am I missing something?

trbutler4:
i can take this

tdelabro:
Go ahead

trbutler4:
would the "output_builtin" amount be equivalent to the keccak amount here?

https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/fee-mechanism/#general_case

trbutler4:
the PR I opened assumes output_builtin to be equivalent to the keccak cost. It does cause some tests to fail, which are included in the screenshot on the pr

PR #1139, opened 2023-09-25
Feat/use actual resource cost

trbutler4:
updated to use actual vm resource costs 

# Pull Request type

- Feature

resource costs hardcoded with 1 

Resolves: #1117 

## What is the new behavior?

updated resource costs to be as defined here: https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/fee-mechanism/#general_case

Yes, failing tests: 

image

### DIFF
--- a/crates/primitives/fee/src/lib.rs
+++ b/crates/primitives/fee/src/lib.rs
@@ -33,17 +33,16 @@ pub const FEE_TRANSFER_N_STORAGE_CHANGES: u8 = 2; // Sender and sequencer balanc
 /// Number of storage updates to actually charge for the fee transfer tx.
 pub const FEE_TRANSFER_N_STORAGE_CHANGES_TO_CHARGE: u8 = FEE_TRANSFER_N_STORAGE_CHANGES - 1; // Exclude the sequencer balance update, since it's charged once throughout the batch.
 
-// TODO: add real values here.
 // FIXME: https://github.com/keep-starknet-strange/madara/issues/330
 static VM_RESOURCE_FEE_COSTS: phf::Map<&'static str, f64> = phf_map! {
-    "n_steps" => 1_f64,
-    "pedersen_builtin" => 1_f64,
-    "range_check_builtin" => 1_f64,
-    "ecdsa_builtin" => 1_f64,
-    "bitwise_builtin" => 1_f64,
-    "poseidon_builtin" => 1_f64,
-    "output_builtin" => 1_f64,
-    "ec_op_builtin" => 1_f64,
+    "n_steps" => 0.01_f64,
+    "pedersen_builtin" => 0.32_f64,
+    "range_check_builtin" => 0.16_f64,
+    "ecdsa_builtin" => 20.48_f64,
+    "bitwise_builtin" => 0.64_f64,
+    "poseidon_builtin" => 0.32_f64,
+    "output_builtin" => 20.48_f64,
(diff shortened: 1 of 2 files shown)


crates/primitives/fee/src/lib.rs
```diff
@@ -33,17 +33,16 @@ pub const FEE_TRANSFER_N_STORAGE_CHANGES: u8 = 2; // Sender and sequencer balanc
 /// Number of storage updates to actually charge for the fee transfer tx.
 pub const FEE_TRANSFER_N_STORAGE_CHANGES_TO_CHARGE: u8 = FEE_TRANSFER_N_STORAGE_CHANGES - 1; // Exclude the sequencer balance update, since it's charged once throughout the batch.
 
-// TODO: add real values here.
 // FIXME: https://github.com/keep-starknet-strange/madara/issues/330
```
d-roak, line 36: does it make sense to keep this comment here? cc @tdelabro

d-roak:
@trbutler4 thank you so much! Fix the failing test in this PR also, please

crates/primitives/fee/src/lib.rs
```diff
@@ -33,17 +33,16 @@ pub const FEE_TRANSFER_N_STORAGE_CHANGES: u8 = 2; // Sender and sequencer balanc
...
-    "poseidon_builtin" => 1_f64,
-    "output_builtin" => 1_f64,
-    "ec_op_builtin" => 1_f64,
+    "n_steps" => 0.01_f64,
+    "pedersen_builtin" => 0.32_f64,
+    "range_check_builtin" => 0.16_f64,
+    "ecdsa_builtin" => 20.48_f64,
+    "bitwise_builtin" => 0.64_f64,
+    "poseidon_builtin" => 0.32_f64,
+    "output_builtin" => 20.48_f64,
```
tdelabro, line 44: I asked internaly for the value of the `output_builtin`. Leave it with whatever you want for now.

crates/primitives/fee/src/lib.rs
tdelabro: We can remove it

tdelabro: And the next line too I belive

tdelabro (changes requested):

crates/primitives/fee/src/lib.rs
tdelabro: I got confirmation that `output_builtin` has no meaning for starknet contracts.
You can safely remove it from the entries

trbutler4:
> @trbutler4 can you rebase and fix the last conversations? You have all the information now :)

sure thing!

trbutler4:
i have updated the tests, but I dont entirely understand them. Most seemed to pass after adjusting the "low" transaction cost(?) Please let me know if anything doesnt make sense!

crates/primitives/fee/src/lib.rs
trbutler4: removed it

tdelabro:
> i have updated the tests, but I dont entirely understand them. Most seemed to pass after adjusting the "low" transaction cost(?) Please let me know if anything doesnt make sense!

Because you updated the price of the resources, the expected fee is changed. Most of them went down a lot because step price went from 1 to 0.01, and so on.

So it seems to add up to me

tdelabro (approved):

PR #1143, opened 2023-09-27
added SN_MAIN mont representation in pallet constants

antiyro:
Added SN_MAIN mont representation in pallet constants:

```
❯ starkli mont 0x534e5f4d41494e
[
    17696389056366564951,
    18446744073709551615,
    18446744073709551615,
    502562008147966918,
]
```


crates/runtime/src/pallets.rs
```diff
@@ -31,6 +31,14 @@ pub const SN_GOERLI_CHAIN_ID: Felt252Wrapper = Felt252Wrapper(starknet_ff::Field
     398700013197595345,
 ]));
 
+/// ChainId for Starknet Mainnet
```
tdelabro, line 34: Now that we have multiple values. I would add this under a new primitive crate `mp-chain-id` with both values being hardcoded there.

And can then import it where ever it's used

tdelabro (changes requested):
Also update the changelog

crates/runtime/src/pallets.rs
antiyro: Done, tell me if any changes are needed. I did not called the constant in starknet-rpc-test since it does not match the crate three.

crates/primitives/chain-id/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+[dependencies]
+
+# Madara Primitives
+mp-felt = { workspace = true }
+mp-hashers = { workspace = true }
+mp-transactions = { workspace = true }
+# Starknet dependencies
+blockifier = { workspace = true }
+starknet-ff = { workspace = true }
+starknet_api = { workspace = true }
```
tdelabro, line 22: I think you don't need all of those dependencies.
Only `starknet-ff` and `mp-felt`

crates/primitives/chain-id/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+    398700013197595345,
+]));
+
+/// ChainId for Starknet Mainnet
+pub const SN_MAIN_CHAIN_ID: Felt252Wrapper = Felt252Wrapper(starknet_ff::FieldElement::from_mont([
+    17696389056366564951,
+    18446744073709551615,
+    18446744073709551615,
+    502562008147966918,
+]));
```
tdelabro, line 20: Can you add two tests where you assert that the const are equal to the value build with `from_hex_be(b"SN_GOERLI")` and `from_hex_be(b"SN_MAINET")`.

This way we are sure those numbers are really what they claim to be

starknet-rpc-test/src/constants.rs
```diff
@@ -32,4 +32,4 @@ pub const MAX_FEE_OVERRIDE: &str = "0x100000";
 // Need to use `from_mont` because this needs to be a constant function call
 /// ChainId for Starknet Goerli testnet
 pub const SN_GOERLI_CHAIN_ID: FieldElement =
-    FieldElement::from_mont([3753493103916128178, 18446744073709548950, 18446744073709551615, 398700013197595345]);
+    FieldElement::from_mont([3753493103916128178, 18446744073709548950, 18446744073709551615, 398700013197595345]);
```
tdelabro, line 35: Can't you import it? Rather than defining it again?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - fix: fix sharingan chain spec
 - fix: update madara infra to main branch
 - fix: update `Cargo.lock`
+- fix: exported chain id constant in mp-chain-id crate and added one for SN_MAIN
```
tdelabro, line 11: Not a `fix`, rather a `refactor` 
https://www.conventionalcommits.org/en/v1.0.0-beta.4/#summary

tdelabro (changes requested):

starknet-rpc-test/src/constants.rs
antiyro: That was in a previous commit so I think I'll fuck it up if try to recover changes on it

crates/primitives/chain-id/Cargo.toml
```diff
@@ -14,9 +14,11 @@ targets = ["x86_64-unknown-linux-gnu"]
...
 # Starknet dependencies
-blockifier = { workspace = true }
 starknet-ff = { workspace = true }
-starknet_api = { workspace = true }
\ No newline at end of file
+
+[features]
+default = ["std"]
+std = [
+  "mp-felt/std",
```
tdelabro, line 23: I think starknet-ff also has a `std` feature

crates/primitives/chain-id/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! Traits for chain ids.
 #![cfg_attr(not(feature = "std"), no_std)]
 
+
```
tdelabro, line 4: empty line.
Rut cargo fmt maybe

crates/primitives/chain-id/src/tests.rs
```diff
@@ -0,0 +1,27 @@
+use starknet_ff::FieldElement;
+
+use crate::*;
+
+#[test]
+fn test_sn_goerli_chain_id() {
+    let sn_goerli_chain_id = FieldElement::from_byte_slice_be(b"SN_GOERLI").unwrap();
+    let mont_sn_goerli_chain_id = FieldElement::into_mont(sn_goerli_chain_id);
+    let expected_value: Felt252Wrapper = Felt252Wrapper(starknet_ff::FieldElement::from_mont(mont_sn_goerli_chain_id)).into();
```
tdelabro, line 9: Why are you doing this? It looks useless

Can't you do this instead:
```rust
let expected_value = Felt252Wrapper(FieldElement::from_byte_slice_be(b"SN_GOERLI").unwrap());
assert_eq!(
        SN_GOERLI_CHAIN_ID, 
        expected_value, 
        "SN_GOERLI_CHAIN_ID does not match the expected value."
    );
```

tdelabro (changes requested):

starknet-rpc-test/src/constants.rs
tdelabro: Right now it looks like the change is not on your branch. But it should be.

Or am I missing something?

crates/primitives/chain-id/src/tests.rs
antiyro: Right, I forgot the from_mont in my declaration

tdelabro (approved):

PR #1146, opened 2023-09-28
Validate L1 Messages

### LINKED ISSUE #1156: add the validate L1Handler logic in `fn validate_unsigned` in pallet starknet

bidzyyys:
closes #1156

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -340,7 +340,7 @@ pub mod pallet {
     #[pallet::storage]
     #[pallet::unbounded]
     #[pallet::getter(fn l1_messages)]
-    pub(super) type L1Messages<T: Config> = StorageMap<_, Blake2_128Concat, Nonce, (), OptionQuery>;
+    pub(super) type L1Messages<T: Config> = StorageMap<_, Twox64Concat, Nonce, (), OptionQuery>;
 
     /// Starknet genesis configuration.
     #[pallet::genesis_config]
@@ -763,15 +763,17 @@ pub mod pallet {
             let mut valid_transaction_builder = ValidTransaction::with_tag_prefix("starknet")
                 .priority(u64::MAX - nonce_for_priority)
                 .and_provides((sender_address, transaction_nonce))
-                .longevity(T::TransactionLongevity::get())
+				.longevity(T::TransactionLongevity::get())
                 .propagate(true);
 
-            // Enforce waiting for the tx with the previous nonce,
-            // to be either executed or ordered before in the block
-            if transaction_nonce > sender_nonce {
-                valid_transaction_builder = valid_transaction_builder
-                    .and_requires((sender_address, Felt252Wrapper(transaction_nonce.0 - FieldElement::ONE)));
-            }
+            if let Some(sender_nonce) = sender_nonce {
+					// Enforce waiting for the tx with the previous nonce,
+                    // to be either executed or ordered before in the block
+                    if transaction_nonce > sender_nonce {
+                        valid_transaction_builder = valid_transaction_builder
+                            .and_requires((sender_address, Felt252Wrapper(transaction_nonce.0 - FieldElement::ONE)));
+                    }
+                }
 
             valid_transaction_builder.build()
         }
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -732,63 +732,12 @@ pub mod pallet {
             // otherwise we have a nonce error and everything fails.
             // Once we have a real fee market this is where we'll chose the most profitable transaction.
 
-            let chain_id = Self::chain_id();
-            let block_context = Self::get_block_context();
-            let mut state: BlockifierStateAdapter<T> = BlockifierStateAdapter::<T>::default();
-            let mut execution_resources = ExecutionResources::default();
-            let mut initial_gas = blockifier::abi::constants::INITIAL_GAS_COST;
-
             let transaction = Self::get_call_transaction(call.clone()).map_err(|_| InvalidTransaction::Call)?;
 
             // Check the nonce is correct
-            let (sender_address, sender_nonce, transaction_nonce) =
-                if let UserAndL1HandlerTransaction::User(ref transaction) = transaction {
-                    let sender_address: ContractAddress = transaction.sender_address().into();
-                    let sender_nonce: Felt252Wrapper = Pallet::<T>::nonce(sender_address).into();
-                    let transaction_nonce = transaction.nonce();
+            let (sender_address, sender_nonce, transaction_nonce) = Self::validate_usigned_tx_nonce(&transaction)?;
 
-                    // Reject transaction with an already used Nonce
-                    if sender_nonce > *transaction_nonce {
-                        Err(InvalidTransaction::Stale)?;
-                    }
-
-                    // A transaction with a nonce higher than the expected nonce is placed in
-                    // the future queue of the transaction pool.
-                    if sender_nonce < *transaction_nonce {
-                        log!(
-                            info,
-                            "Nonce is too high. Expected: {:?}, got: {:?}. This transaction will be placed in the \
-                             transaction pool and executed in the future when the nonce is reached.",
-                            sender_nonce,
-                            transaction_nonce
-                        );
-                    }
-
-                    (transaction.sender_address(), sender_nonce, *transaction_nonce)
-                } else {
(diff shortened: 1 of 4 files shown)


tdelabro (changes requested):

crates/primitives/transactions/src/execution.rs
```diff
@@ -517,6 +517,10 @@ impl Execute for DeployAccountTransaction {
     }
 }
 
+impl Validate for L1HandlerTransaction {
+    const VALIDATE_TX_ENTRY_POINT_NAME: &'static str = VALIDATE_ENTRY_POINT_NAME;
+}
```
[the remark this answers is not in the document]
bidzyyys: @tdelabro 💯 agree on tests, I am not sure if it works as expected. I wanted at first to check if I've been going in the right direction.

crates/pallets/starknet/src/lib.rs
```diff
@@ -826,6 +775,81 @@ pub mod pallet {
...
+                 pool and executed in the future when the nonce is reached.",
+                sender_nonce,
+                transaction_nonce
+            );
+        }
+
+        Ok((sender_address.into(), sender_nonce, transaction_nonce))
+    }
+
+    fn validate_unsigned_tx(transaction: &UserAndL1HandlerTransaction) -> Result<(), InvalidTransaction> {
```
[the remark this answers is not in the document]
bidzyyys: sure!

crates/pallets/starknet/src/lib.rs
```diff
@@ -826,6 +775,81 @@ pub mod pallet {
 
 /// The Starknet pallet internal functions.
 impl<T: Config> Pallet<T> {
```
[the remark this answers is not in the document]
bidzyyys: I think it will be nice refactor!

crates/primitives/transactions/src/execution.rs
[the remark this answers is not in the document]
tdelabro: And I don't think we have test either for the other tx.
So you can:
- have the test done in a separated issue/PR, along with the other tx types
or
- add them there, for l1Handler only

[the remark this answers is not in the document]
bidzyyys: @tdelabro I will add tests in separated issue for all tx types.

crates/pallets/starknet/src/transaction_validation.rs
```diff
@@ -50,3 +53,67 @@ impl<OuterOrigin: Into<Result<RawOrigin, OuterOrigin>> + From<RawOrigin>> Ensure
...
     }
 }
+
+impl<T: Config> Pallet<T> {
+    pub fn validate_usigned_tx_nonce(
+        transaction: &UserAndL1HandlerTransaction,
+    ) -> Result<(Felt252Wrapper, Felt252Wrapper, Felt252Wrapper), InvalidTransaction> {
+        let (sender_address, transaction_nonce) = match transaction {
+            UserAndL1HandlerTransaction::User(tx) => (tx.sender_address().into(), *tx.nonce()),
+            UserAndL1HandlerTransaction::L1Handler(tx, _fee) => (ContractAddress::default(), tx.nonce.into()),
```
bidzyyys, line 63: @tdelabro I need to confirm if `sender_address` for `L1Handler` should be `ContractAddress::default()` or sth else

bidzyyys: Like here -> https://github.com/keep-starknet-strange/madara/blob/main/crates/pallets/starknet/src/tests/l1_message.rs#L20-L22

tdelabro: Good catch. The starknet specs don't enforce ordering for messaging. The nonce only exists for deduplication, not for ordering.

So there is no need for an account to be used. But it also means we have to keep multiple nonces in memory.

I think we should use a persistent hashset in the storage for now.

Check the nonce is not in during validation, put the nonce in during execution.

tdelabro: UserTx and messages have to be treated differently there

crates/primitives/transactions/src/execution.rs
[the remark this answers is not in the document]
tdelabro: See #1155

tdelabro (changes requested):

crates/pallets/starknet/src/transaction_validation.rs
bidzyyys: @tdelabro do I understand correctly that:
- I should add HashSet to `Starknet Pallet` for tracking nonces
- I should move nonce check to the execution part
- I should allow execution of messages only with unused nonce (no need for ordering)

Besides that, could you elaborate more on 
> UserTx and messages have to be treated differently there

crates/pallets/starknet/src/lib.rs
```diff
@@ -330,6 +330,14 @@ pub mod pallet {
     #[pallet::getter(fn seq_addr_update)]
     pub type SeqAddrUpdate<T: Config> = StorageValue<_, bool, ValueQuery>;
 
+    /// Mapping from Starknet class hash to contract class.
+    /// Safe to use `Identity` as the key is already a hash.
+    #[pallet::storage]
+    #[pallet::unbounded]
+    #[pallet::getter(fn l1_messages)]
+    pub(super) type L1Messages<T: Config> =
+        StorageMap<_, Blake2_128Concat, Felt252Wrapper, TransactionHash, OptionQuery>;
```
tdelabro, line 339: Why are we storing the hash of the transaction?
Is it used for something?

crates/pallets/starknet/src/lib.rs
```diff
@@ -330,13 +330,12 @@ pub mod pallet {
...
-    /// Mapping from Starknet class hash to contract class.
-    /// Safe to use `Identity` as the key is already a hash.
+    /// Information about processed L1 Messages
+    /// Based on Nonce value.
     #[pallet::storage]
     #[pallet::unbounded]
     #[pallet::getter(fn l1_messages)]
-    pub(super) type L1Messages<T: Config> =
-        StorageMap<_, Blake2_128Concat, Felt252Wrapper, TransactionHash, OptionQuery>;
+    pub(super) type L1Messages<T: Config> = StorageMap<_, Blake2_128Concat, Nonce, (), OptionQuery>;
```
tdelabro, line 338: https://docs.substrate.io/build/runtime-storage/#hashing-algorithms

Since the caller cannot forge the key values himself (they have to be nonce), we can use https://paritytech.github.io/polkadot-sdk/master/frame_support/struct.Twox64Concat.html instead

tdelabro (changes requested):

crates/pallets/starknet/src/transaction_validation.rs
```diff
@@ -50,3 +53,84 @@ impl<OuterOrigin: Into<Result<RawOrigin, OuterOrigin>> + From<RawOrigin>> Ensure
...
+                        &mut initial_gas,
+                        false,
+                    ),
+                }
+            }
+            UserAndL1HandlerTransaction::L1Handler(transaction, fee) => transaction
+                .into_executable::<T::SystemHash>(chain_id, *fee, false)
+                .validate_tx(&mut state, &block_context, &mut execution_resources, &mut initial_gas, false),
+        }
+        .map_or_else(|_error| Err(InvalidTransaction::BadProof), |_res| Ok(()))
```
tdelabro, line 127: I'm nitpicking but


```rust
.map_err(|_| Err(InvalidTransaction::BadProof))?;

Ok(())
```

Is a more standard and therefore easily readable way of structuring this part of the code (IMO, maybe it's just me)

crates/pallets/starknet/src/transaction_validation.rs
```diff
@@ -50,3 +53,84 @@ impl<OuterOrigin: Into<Result<RawOrigin, OuterOrigin>> + From<RawOrigin>> Ensure
...
+    }
+
+    pub fn ensure_l1_message_not_executed(nonce: &Nonce) -> Result<(), InvalidTransaction> {
+        match L1Messages::<T>::contains_key(nonce) {
+            true => Err(InvalidTransaction::Stale),
+            false => Ok(()),
+        }
```
tdelabro, line 134: The more idiomatic way to handle booleans is by using if/else.
`match` feels a bit overkill there ;)

tdelabro (changes requested):

tdelabro (changes requested):

tdelabro (approved):

Issue #1151, opened 2023-09-28
test: add starknet-rpc-test for estimate_fee


tdelabro:

kasteph:
@tdelabro I can work on this!

tdelabro:
How is it going?

kasteph:
@tdelabro I have a PR open already at #1153

tdelabro:
Oh my bad. Is it ready for review?

kasteph:
@tdelabro yes, ser!

PR #1153, opened 2023-09-28
test: add e2e test for rpc's estimate_fee

kasteph:
# Pull Request type

- Testing

Currently, no test for the rpc method: `estimate_fee`.

Resolves: #1151

## What is the new behavior?

- Adding testing for estimate_fee

No

### DIFF
--- a/starknet-rpc-test/estimate_fee.rs
+++ b/starknet-rpc-test/estimate_fee.rs
@@ -4,10 +4,12 @@ use assert_matches::assert_matches;
 use rstest::rstest;
 use starknet_core::types::{BlockId, BlockTag, BroadcastedInvokeTransaction, BroadcastedTransaction, StarknetError};
 use starknet_ff::FieldElement;
-use starknet_providers::ProviderError::StarknetError as StarknetProviderError;
 use starknet_providers::{MaybeUnknownErrorCode, Provider, StarknetErrorWithMessage};
-use starknet_rpc_test::fixtures::{broadcasted_declare_txn_v1, madara};
+use starknet_providers::ProviderError::StarknetError as StarknetProviderError;
+use starknet_rpc_test::constants::ACCOUNT_CONTRACT;
+
 use starknet_rpc_test::MadaraClient;
+use starknet_rpc_test::fixtures::{broadcasted_declare_txn_v1, madara};
 
 #[rstest]
 #[tokio::test]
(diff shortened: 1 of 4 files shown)


starknet-rpc-test/estimate_fee.rs
```diff
@@ -0,0 +1,92 @@
...
+#[tokio::test]
+async fn fail_if_one_txn_cannot_be_executed(#[future] madara: MadaraClient) -> Result<(), anyhow::Error> {
+    let madara = madara.await;
+    let rpc = madara.get_starknet_client();
+
+    madara.create_empty_block().await?;
+
+    assert_matches!(
+        rpc.estimate_fee(&vec![
+            BroadcastedTransaction::Invoke(BroadcastedInvokeTransaction {
```
tdelabro, line 41: Add at least a Tx that can be executed to the call

starknet-rpc-test/estimate_fee.rs
```diff
@@ -0,0 +1,92 @@
...
+        ], BlockId::Tag(BlockTag::Latest)).await,
+        Err(StarknetProviderError(StarknetErrorWithMessage { code: MaybeUnknownErrorCode::Known(code), .. })) if code == StarknetError::ContractError
+    );
+
+    Ok(())
+}
+
+#[rstest]
+#[tokio::test]
+async fn returns_same_vec_length_as_txns(
```
tdelabro, line 58: You can merge this one with `work_ok`

starknet-rpc-test/estimate_fee.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[future] madara: MadaraClient,
+    broadcasted_declare_txn_v1: BroadcastedTransaction,
```
tdelabro, line 60: Use invoke tx instead of declaring one, please.
The declare will start failing when we will move to executing all tests on the same single node (they would have been already declared)

starknet-rpc-test/src/fixtures.rs
```diff
@@ -1,8 +1,36 @@
...
+pub fn broadcasted_declare_txn_v1(compressed_contract_class: CompressedLegacyContractClass) -> BroadcastedTransaction {
+    BroadcastedTransaction::Declare(BroadcastedDeclareTransaction::V1(BroadcastedDeclareTransactionV1 {
+        max_fee: FieldElement::ZERO,
+        signature: vec![],
+        nonce: FieldElement::ZERO,
+        sender_address: FieldElement::from_hex_be(ACCOUNT_CONTRACT).unwrap(),
+        is_query: true,
+        contract_class: compressed_contract_class.into(),
+    }))
+}
```
tdelabro, line 36: I don't think you need a fixture for this. Just declare it in the body of your function

starknet-rpc-test/estimate_fee.rs
```diff
@@ -0,0 +1,92 @@
...
+) -> Result<(), anyhow::Error> {
+    let madara = madara.await;
+    let rpc = madara.get_starknet_client();
+
+    let estimate = rpc.estimate_fee(&vec![broadcasted_declare_txn_v1], BlockId::Tag(BlockTag::Latest)).await?;
+
+    assert!(estimate[0].overall_fee > 0);
+    assert!(estimate[0].gas_consumed > 0);
```
tdelabro, line 89: I think you should then execute them for real and compare the fee of the estimate with the one of the real execution.

It the state of the chain is exactly the same, it should be the same

tdelabro (changes requested):

starknet-rpc-test/src/fixtures.rs
tdelabro: move the fixture to the file you use it

tdelabro (changes requested):

tdelabro (approved):

Issue #1126, opened 2023-09-20
bug: setup logs not displaying


apoorvsadana:
Before the node starts we log some [information](https://github.com/keep-starknet-strange/madara/blob/e7513f22417198e3e3bb5860fae4a2f5ef6bed87/crates/pallets/starknet/src/utils.rs#L55) to tell the user that we are downloading/setting up files. However, the logger is only initialised once the node has started because of which these logs aren't showed up at all. You can find more about the reasons in the discussion here - https://github.com/keep-starknet-strange/madara/pull/1011

We need to evaluate if it's possible to start the logger before the node starts and if yes, we need to implement that.

d-roak:
i assigned it to you. Look carefully to the discussion, it doesn't seem to be a trivial thing to resolve

dbejarano820:
well that was painful, i'm moving on 😆

justinfrevert:
@d-roak I'd like to take a look at this one

d-roak:
All the info is logged in the setup subcommand

PR #1164, opened 2023-10-03
Fix logging during setup

justinfrevert:
# Pull Request type

Current behavior is that the setup cli command produces no logs to output, despite invoking functions which contain logs.

Resolves: https://github.com/keep-starknet-strange/madara/issues/1126

## What is the new behavior?

- Setup function produces logs as expected
- The logger is initialized 

No

## Other information

Screenshot of new output:
image
Are we okay with this output? I am not sure what this is all doing, but if all the information is useful, I think it makes sense.

Noting one other thing: if we're not happy with initializing the whole runner for whatever reason, a third option is to just run it as a script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - fix: fix sharingan chain spec
 - fix: update madara infra to main branch
 - fix: update `Cargo.lock`
-- fix: ensure logger is initialized for `setup` cli command to produce logging output
+- fix: ensure logger is initialized for `setup` cli command
 
 ## v0.3.0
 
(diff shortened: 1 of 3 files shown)


justinfrevert:
Part of this work touches utils which live inside of pallet code, but I think they'd be better suited living in shared `primitives`, or some other shared utils area, as I think they could cause confusion being in this area. If all good, I'd be happy to move those.

tdelabro:
Hey, I have been working on a refacto of this part of the code.
I also explored the solution you are implementing but sadly it doesn't work ;)
You get the error in the CI while trying to run `madara setup`.

Let me finish my work on this and then we will try to explore other ways. 

In the meantime you can pick an issue here: https://github.com/orgs/keep-starknet-strange/projects/10/views/11

PR #1165, opened 2023-10-04
refacto: improve node cli

tdelabro:

Yes, some arguments and behaviors in the CLI changed

## Sumup

- feat(cli): `run` is the by default command when running the `madara` bin
- refacto(cli): `run` and `setup` commands are defined in their own files
- refacto(cli): `run.testnet` argument removed in favor of the substrate native `chain` arg
- feat(cli): `run.fetch_chain_spec` argument removed in favor of the substrate native `chain` arg
- feat(cli): `setup` require a source file, either from an url or a path on the local filesystem
- chore(cli): use `Url`, `Path` and `PathBuf` types rather than `String`
- refacto(cli): moved the pallet/chain_spec/utils methods to the node crate
- feat(cli): `madara_path` arg has been remove, we use the substrate native `base_path` arg instead
- feat(cli): sharingan chain specs are loaded during the compilation, not downloaded from github
- refacto(pallet/starknet): `GenesisLoader` refactored as `GenesisData` + a `base_path` field

## How to use?

You use the same `chain` arg in both `setup` and `run`, and you will be fine.
It will setup the chain spec for this specific chain, and then use those to run.


kalaninja (commented):
Looks very good to me.

crates/node/src/command.rs
```diff
@@ -36,17 +36,33 @@ impl SubstrateCli for Cli {
...
-                    enable_manual_seal,
-                    self.madara_path.clone().expect("`madara_path` expected to be set with clap default value"),
-                )?)
+            DEV_CHAIN_ID => {
+                let enable_manual_seal = self.run.sealing.map(|_| true);
+                let base_path = self
+                    .run
+                    .base
+                    .shared_params
+                    .base_path
```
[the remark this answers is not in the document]
tdelabro: No, you are right!

crates/node/src/command.rs
```diff
@@ -36,17 +36,33 @@ impl SubstrateCli for Cli {
...
+                    .as_ref()
+                    .map(BasePath::new)
+                    .unwrap_or_else(|| BasePath::from_project("", "", &<Cli as SubstrateCli>::executable_name()));
+                Box::new(chain_spec::development_config(enable_manual_seal, base_path)?)
             }
-            "" | "local" | "madara-local" => Box::new(chain_spec::local_testnet_config(
-                self.madara_path.clone().expect("`madara_path` expected to be set with clap default value"),
+            SHARINGAN_CHAIN_ID => Box::new(chain_spec::ChainSpec::from_json_bytes(
+                &include_bytes!("../../../configs/chain-specs/testnet-sharingan-raw.json")[..],
```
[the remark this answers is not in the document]
tdelabro: @d-roak raised the same concern. I think I will do the same. Download it during `setup`, and then just read it from the file.

crates/node/src/command.rs
```diff
@@ -36,17 +36,33 @@ impl SubstrateCli for Cli {
...
-            path => Box::new(chain_spec::ChainSpec::from_json_file(std::path::PathBuf::from(path))?),
+            "" | "local" | "madara-local" => {
+                let base_path = self
+                    .run
+                    .base
+                    .shared_params
+                    .base_path
+                    .as_ref()
+                    .map(BasePath::new)
+                    .unwrap_or_else(|| BasePath::from_project("", "", &<Cli as SubstrateCli>::executable_name()));
```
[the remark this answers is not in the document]
tdelabro: In order to avoid code duplication?

Just the `self
                    .run
                    .base
                    .shared_params
                    .base_path
                    .as_ref()
                    .map(BasePath::new)` part then. Not the `.unwrap_or_else(|| BasePath::from_project("", "", &<Cli as SubstrateCli>::executable_name()))` which is aware of `<Cli as SubstrateCli>`.

crates/node/src/commands/setup.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+}
+
+impl SetupCmd {
+    pub fn run(&self) -> Result<()> {
+        log::info!("setup cmd: {:?}", self);
+        let dest_config_dir_path = {
+            let is_dev = self.shared_params().is_dev();
+            let chain_id = self.shared_params().chain_id(is_dev);
+            let base_path = self
```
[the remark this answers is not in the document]
tdelabro: I will try it!

crates/node/src/commands/run.rs
```diff
@@ -0,0 +1,74 @@
...
+    let sealing = cli.run.sealing;
+
+    runner.run_node_until_exit(|config| async move {
+        service::new_full(config, sealing, da_config).map_err(sc_cli::Error::Service)
+    })
+}
+
+fn override_dev_environment(cmd: &mut ExtendedRunCmd) {
+    // create a reproducible dev environment
+    cmd.base.shared_params.dev = false;
```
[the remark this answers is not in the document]
kalaninja: ok, got it. Thanks for the explanation. It might be useful to put it in the comment though. "create a reproducible dev environment" alone doesn't tell much.

crates/node/src/command.rs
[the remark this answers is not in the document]
kalaninja: Oh, yes, you are right `ExtendedRunCmd` is unaware of the executable name. So, not the best place for the method. Should go to `Cli` then, or left as it is. But, I personally don't like it, 8 duplicated lines of code one after another.

kalaninja:
I also can propose to decompose the big `if`-clause in `setup` module. For example the following way:

```
struct SetupRunner;
trait RunSetup<S> {
    fn run(&self, setup: S) -> Result<()>;
}

struct LocalSetupParams {
    src_dir: PathBuf,
    dest_dir: PathBuf,
}
impl RunSetup<LocalSetupParams> for SetupRunner {
    fn run(&self, setup: LocalSetupParams) -> Result<()> {
        let src_index_path = setup.src_dir.join(CONFIG_INDEX_FILE_NAME);
        let index_json = std::fs::read_to_string(src_index_path).map_err(|e| Error::Application(Box::new(e)))?;
        let index_config: Configs = serde_json::from_str(&index_json)
            .map_err(|e| Error::Input(format!("Invalid `{CONFIG_INDEX_FILE_NAME}` content: {e}")))?;
        write_to_file(index_json, &setup.dest_dir.join(CONFIG_INDEX_FILE_NAME))?;

        let src_genesis_assets_dir = setup.src_dir.join(GENESIS_ASSETS_DIR_NAME);
        let dest_genesis_assets_dir = setup.dest_dir.join(GENESIS_ASSETS_DIR_NAME);
        for asset in index_config.genesis_assets {
            force_copy_file(&src_genesis_assets_dir.join(asset.name), &dest_genesis_assets_dir)?;
        }

        Ok(())
    }
}

struct RemoteSetupParams {
    index_url: Url,
    dest_dir: PathBuf,
}
impl RunSetup<RemoteSetupParams> for SetupRunner {
    fn run(&self, setup: RemoteSetupParams) -> Result<()> {
        let index_bytes = reqwest::blocking::get(setup.index_url)
            .and_then(|response| response.bytes())
            .map_err(|e| Error::Application(Box::new(e)))?;
        let index_config: Configs =
            serde_json::from_slice(&index_bytes[..]).map_err(|e| Error::Application(Box::new(e)))?;
        write_to_file(index_bytes, &setup.dest_dir.join(CONFIG_INDEX_FILE_NAME))?;

        let base_url = Url::parse(&index_config.remote_base_path)
            .and_then(|url| url.join(format!("{GENESIS_ASSETS_DIR_NAME}/").as_str()))
            .map_err(|e| Error::Application(Box::new(e)))?;
        let dest_genesis_assets_dir = setup.dest_dir.join(GENESIS_ASSETS_DIR_NAME);
        for asset in index_config.genesis_assets {
            safe_download_genesis_asset(&base_url, asset, &dest_genesis_assets_dir)?;
        }

        Ok(())
    }
}

enum Setup {
    Local(LocalSetupParams),
    Remote(RemoteSetupParams),
}

impl RunSetup<Setup> for SetupRunner {
    fn run(&self, setup: Setup) -> Result<()> {
        match setup {
            Setup::Local(params) => self.run(params),
            Setup::Remote(params) => self.run(params),
        }
    }
}

fn write_to_file<T: AsRef<[u8]>>(content: T, dest_file: &Path) -> Result<()> {
    let parent = dest_file.parent().ok_or("Failed to get parent directory")?;
    std::fs::create_dir_all(parent)?;
    std::fs::File::create(dest_file)?.write_all(content.as_ref())?;

    Ok(())
}

fn force_copy_file(src_file: &Path, dest_path: &Path) -> Result<()> {
    let file_name = src_file.file_name().ok_or("File name not found")?;
    if !src_file.exists() {
        return Err(format!("Source file '{}' does not exist", src_file.display()).into());
    }

    std::fs::create_dir_all(dest_path)?;
    std::fs::copy(src_file, dest_path.join(file_name))?;

    Ok(())
}

fn safe_download_genesis_asset(base_url: &Url, file: FileInfos, dest_path: &Path) -> Result<()> {
    let full_url = base_url.join(&file.name).map_err(|e| Error::Application(Box::new(e)))?;
    println!("Fetching '{}'", &full_url);

    let file_content = reqwest::blocking::get(full_url.clone())
        .and_then(|response| response.bytes())
        .map_err(|e| Error::Application(Box::new(e)))
        .and_then(|bytes| write_to_file(&bytes, &dest_path.join(file.name)).map(|_| bytes))?;

    if let Some(expected_hash) = file.md5 {
        let digest = md5::compute(file_content);
        let actual_hash = format!("{:x}", digest);
        if actual_hash != expected_hash {
            return Err(format!("MD5 mismatch for file '{full_url}': {actual_hash} != {expected_hash}").into());
        }
    }

    Ok(())
}
```

So `SetupCmd` can be implemented the following way:

```
impl SetupCmd {
    pub fn run(&self) -> Result<()> {
        log::info!("setup cmd: {:?}", self);

        let is_dev = self.shared_params().is_dev();
        let chain_id = self.shared_params().chain_id(is_dev);
        let dest_dir = self
            .shared_params()
            .base_path()?
            .unwrap_or_else(|| BasePath::from_project("", "", &Cli::executable_name()))
            .config_dir(&chain_id);
        log::info!("Setting up madara config at '{}'", dest_dir.display());

        let setup = if let Some(path) = &self.source.from_local {
            Setup::Local(LocalSetupParams { src_dir: PathBuf::from(path), dest_dir })
        } else if let Some(url) = &self.source.from_remote {
            let index_url =
                Url::parse(url).map_err(|e| Error::Input(format!("Invalid url for `{CONFIG_INDEX_FILE_NAME}`:{e}")))?;
            println!("Fetching chain config from '{}'", &index_url);

            Setup::Remote(RemoteSetupParams { index_url, dest_dir })
        } else {
            unreachable!(
                "clap::Args is derived upon `SetupSource` in a way that guarantee that either `from_remote` or \
                 `from_local` is present"
            );
        };

        SetupRunner.run(setup)
    }
}
```

crates/node/src/command.rs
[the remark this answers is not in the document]
kalaninja: btw, `executable_name` is not a method. you can still use it
```
impl ExtendedRunCmd {
    fn base_path(&self) -> Result<BasePath> {
        self.base.base_path().map(|maybe_path| {
            maybe_path.unwrap_or_else(|| BasePath::from_project("", "", &<Cli as SubstrateCli>::executable_name()))
        })
    }
}
```

crates/node/src/command.rs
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/command.rs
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/chain_spec.rs
```diff
@@ -143,14 +148,13 @@ pub fn local_testnet_config(madara_path: PathBuf) -> Result<ChainSpec, String> {
     ))
 }
 
-fn load_genesis(madara_path: PathBuf) -> GenesisLoader {
-    let madara_path = madara_path.to_str().unwrap().to_string();
-    let genesis_path = madara_path.clone() + "/configs/genesis-assets/genesis.json";
-    let genesis = utils::read_file_to_string(genesis_path)
+fn load_genesis(data_path: PathBuf) -> GenesisLoader {
+    let genesis_path = data_path.join("genesis-assets/genesis.json");
```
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/commands/run.rs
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/commands/setup.rs
```diff
@@ -0,0 +1,163 @@
...
+/// Where the `md5` and `url` fields are optional
+#[derive(Debug, clap::Args)]
+#[group(required = true, multiple = false)]
+pub struct SetupSource {
+    /// Download an index.json file for an url
+    #[clap(
+        long,
+        conflicts_with="from_local",
+        value_hint=clap::ValueHint::Url,
+        // This combination of properties allow us to use a default value only if the arg is passed witout value.
```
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/commands/setup.rs
```diff
@@ -0,0 +1,163 @@
...
+                configs_content
+            };
+
+            let base_url = Url::parse(&madara_configs.remote_base_path).map_err(|e| Error::Application(Box::new(e)))?;
+            for asset in madara_configs.genesis_assets {
+                fetch_and_validate_genesis_assets(&base_url, asset, &dest_config_dir_path)?;
+            }
+        } else {
+            unreachable!(
+                "clap::Args is derived upon `SetupSource` in a way that guarante that either `from_remote` or \
```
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/commands/setup.rs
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/command.rs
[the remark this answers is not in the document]
kalaninja: good idea

kalaninja:
> I understand its appeal. But I prefer to keep it as it is for now.

I agree. The proposed change makes more sense if you are going to add other ways to setup.

kalaninja (approved):

crates/node/src/commands/setup.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Copy an index.json file for an url
+    #[clap(long, conflicts_with = "from_remote")]
+    pub from_local: Option<String>,
+}
+
+#[derive(Debug, clap::Args)]
+pub struct SetupCmd {
+    #[allow(missing_docs)]
+    #[clap(flatten)]
+    pub shared_params: SharedParams,
```
d-roak, line 41: why? we have `--dev` flag, that doesn't make sense on the `SetupCmd` scenario

crates/node/src/commands/setup.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn shared_params(&self) -> &SharedParams {
+        &self.shared_params
+    }
+}
+
+impl SetupCmd {
+    pub fn run(&self) -> Result<()> {
+        log::info!("setup cmd: {:?}", self);
+        let dest_config_dir_path = {
+            let is_dev = self.shared_params().is_dev();
```
d-roak, line 57: shouldn't it be verified by the `--chain dev`?

crates/node/src/commands/setup.rs
```diff
@@ -0,0 +1,176 @@
...
+            let chain_id = self.shared_params().chain_id(is_dev);
```
d-roak, line 58: `chain_id(true)` ?

crates/node/src/commands/setup.rs
tdelabro: You are right. Because we overwrite the `is_dev` setting.

crates/node/src/commands/setup.rs
tdelabro: why that?

tdelabro: you can setup for non-dev chains. Local but non dev

d-roak: that is what u are setting there atm, i think
`let is_dev = self.shared_params().is_dev();` = `let is_dev = true;`

crates/node/src/commands/setup.rs
tdelabro: well actually no. We overwrite it only for run.
Not for setup

d-roak: yes, the argument is that `--dev` shouldn't exist in this context at all, and we should just used the `--chain` flag

tdelabro: `--chain dev` and `--chain local --dev` are two different things

the first one give you the `dev` chain
the second one the `local` with `force_authorign` and `alice` and rpc_exteranal unsafe

tdelabro: it should be
```
    if cmd.base.shared_params.chain.is_none() {
        cmd.base.shared_params.chain = Some("dev".to_string())
    };
```
in the override method

tdelabro: oh no you are right

```rust
	#[arg(long, conflicts_with_all = &["chain"])]
	pub dev: bool,
```

tdelabro: but still, you can run the following `--chain=dev` without `--alice` and so on

And the impl of `.chain_id(is_dev)` is the following:

```rust
	/// Get the chain spec for the parameters provided
	pub fn chain_id(&self, is_dev: bool) -> String {
		match self.chain {
			Some(ref chain) => chain.clone(),
			None =>
				if is_dev {
					"dev".into()
				} else {
					"".into()
				},
		}
	}
```

tdelabro: so no. It should stay as it is

crates/node/src/commands/setup.rs
tdelabro: no, is dev just check if `--dev` was passed
so `madara setup --dev` with have it true and `madara setup --chain=local` will have it false

crates/node/src/commands/setup.rs
tdelabro: > yes, the argument is that --dev shouldn't exist in this context at all, and we should just used the --chain flag

`--dev` still exists. It's only overwritten for the `run` cmd.

kalaninja: You can implement it the way you want, but default behavior for substrate is that `--dev` is and alias for ` --chain dev`

kalaninja: oh no, it has a bigger impact.

d-roak (approved):

PR #1186, opened 2023-10-12
docs: fix setup and run commands in `getting-started.md`

kalaninja:
This PR updates `getting-started.md` to follow the changes introduced in #1165:
- the behavior of node's `setup` command was changed:
  - using one of the switches became [mandatory](https://github.com/keep-starknet-strange/madara/blob/29a8b112f5a09dc2ad9a29989b95b7e3a15623ec/crates/node/src/commands/setup.rs#L19): either `--from_remote` or `--from_local`
  - `--chain` should be specified if you intend to run the node in dev mode, because `--dev` sets `--chain=dev` implicitly
- `run` command was uplifted (its arguments apply to the node directly)

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -16,17 +16,17 @@ rustup show
 Use Rust's native `cargo` command to build and launch the template node:
 
 ```sh
-cargo run --release -- setup
-cargo run --release -- run --dev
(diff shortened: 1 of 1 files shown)


kalaninja:
cc: @tdelabro

tdelabro (approved):

PR #1189, opened 2023-10-15
Refactor non-default sealing

### LINKED ISSUE #1144: feat: allow `dev` nodes to use instant seal
# Feature Request

Right now in `load_spec`
```rust
Ok(match id {
            "dev" => {
                let enable_manual_seal = self.run.sealing.map(|_| true);
```
Meaning event if ` self.run.sealing` was of value `Some(Sealing::Instant)` we would still use manual sealing 

@EvolveArt pointed out that enableing it may involve solving this TODO: https://github.com/keep-starknet-strange/madara/blob/main/crates/node/src/service.rs#L319

kalaninja:
- fix(service): confusing message when node starts (output the actual sealing method being used)
- refactor: how the sealing mode is passed into runtime
- feat: finalization for instant sealing

New sealing mode `instant-finality` is added to the cli (kudos to @tdelabro ) for command parsing:
```rust
enum Sealing {
  Manual,
  Instant,
  InstantFinality,
}
``` 

Internally sealing mode is represented with the following domain model:
```rust
enum SealingMode {
    #[default]
    Default,
    Manual,
    Instant {
        finalize: bool,
    },
}
``` 

`let (command_sink, commands_stream) = mpsc::channel(1000);` is adjusted to create for `SealingMode::Manual` only.

Closes #1144

### DIFF
--- a/crates/node/src/commands/run.rs
+++ b/crates/node/src/commands/run.rs
@@ -12,20 +12,11 @@ use crate::service;
 /// Available Sealing methods.
 #[derive(Debug, Copy, Clone, clap::ValueEnum, Default, Serialize, Deserialize)]
 pub enum Sealing {
-    // Seal using rpc method.
+    /// Seal using rpc method.
     #[default]
     Manual,
-    // Seal when transaction is executed.
+    /// Seal when transaction is executed.
     Instant,
-}
-
-impl From<Sealing> for SealingMode {
-    fn from(value: Sealing) -> Self {
-        match value {
-            Sealing::Manual => SealingMode::Manual,
-            Sealing::Instant => SealingMode::Instant,
-        }
-    }
 }
 
 #[derive(Clone, Debug, clap::Args)]
@@ -36,6 +27,11 @@ pub struct ExtendedRunCmd {
     /// Choose sealing method.
     #[clap(long, value_enum, ignore_case = true)]
     pub sealing: Option<Sealing>,
+
+    // TODO. Move this into `Sealing::Instant` when https://github.com/clap-rs/clap/issues/2621 is resolved.
+    /// Finalize sealed blocks. This flag is ignored if sealing is set to `Manual`.
+    #[clap(long, default_value = "false", default_missing_value = "true")]
+    pub finalize: bool,
 
     /// Choose a supported DA Layer
     #[clap(long)]
@@ -49,6 +45,19 @@ impl ExtendedRunCmd {
             .shared_params
             .base_path()?
             .unwrap_or_else(|| BasePath::from_project("", "", &<Cli as SubstrateCli>::executable_name())))
+    }
+}
+
+impl From<ExtendedRunCmd> for SealingMode {
+    fn from(value: ExtendedRunCmd) -> Self {
+        if let Some(sealing) = value.sealing {
+            match sealing {
+                Sealing::Manual => SealingMode::Manual,
+                Sealing::Instant => SealingMode::Instant { finalize: value.finalize },
+            }
+        } else {
+            SealingMode::Default
+        }
     }
 }
 
@@ -75,9 +84,8 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
         }
     };
 
-    let sealing = cli.run.sealing;
     runner.run_node_until_exit(|config| async move {
-        service::new_full(config, sealing, da_config).map_err(sc_cli::Error::Service)
+        service::new_full(config, cli.run.into(), da_config).map_err(sc_cli::Error::Service)
     })
 }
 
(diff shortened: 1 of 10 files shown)


d-roak:
Hey @kalaninja, thanks ser !
Can u just had an entry to the changelog pls

kalaninja:
NOTE: By default substrate doesn't finalize blocks with instant sealing. It is [hardcoded](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/client/consensus/manual-seal/src/lib.rs#L240)
I can implement a workaround if needed, but I don't know what behavior is needed (if need):
- always finalize
- add cli switch, e.g. `--sealing=instant --finalize=true`

crates/node/src/service.rs
```diff
@@ -423,7 +421,8 @@ pub fn new_full(
 
             network_starter.start_network();
 
-            log::info!("Manual Seal Ready");
+            log::info!("{:?} Seal Enabled", sealing);
```
tdelabro, line 424: Can you impl `Display` on `SealingMode` so it goes out in a human-readable way?

tdelabro (changes requested):

kalaninja:
As discussed with @d-roak  I've added support for block finalization when instant sealing is used.

crates/node/src/commands/run.rs
```diff
@@ -26,6 +28,11 @@ pub struct ExtendedRunCmd {
     #[clap(long, value_enum, ignore_case = true)]
     pub sealing: Option<Sealing>,
 
+    // TODO. Move this into `Sealing::Instant` when https://github.com/clap-rs/clap/issues/2621 is resolved.
+    /// Finalize sealed blocks. This flag is ignored if sealing is set to `Manual`.
+    #[clap(long, default_value = "false", default_missing_value = "true")]
```
tdelabro, line 33:
```
        // This combination of properties allow us to use a default value only if the arg is passed without value.
        // If it is not passed at all, no default value is used.
        // See: https://docs.rs/clap/latest/clap/struct.Arg.html#method.default_missing_value
        num_args = 0..=1, require_equals=true, default_missing_value = Some(constants::DEFAULT_CONFIGS_URL)
```
you have to provide all three of the arguments if you want to use `default_missing_value` correctly

crates/node/src/commands/run.rs
```diff
@@ -26,6 +28,11 @@ pub struct ExtendedRunCmd {
...
+    pub finalize: bool,
```
tdelabro, line 34: this argument only makes sense when sealing is set to `Instant`
https://docs.rs/clap/latest/clap/struct.Arg.html#method.required_if_eq allow you to express this relationship

tdelabro: Overall this is a lot of complexity when you could just extend the `Sealing` enum
```rust
enum Sealing {
  Manual,
  Instant,
  InstantFinality,
}

**Review by tdelabro — CHANGES_REQUESTED** [review]

File: crates/node/src/commands/run.rs (hunk 2 above)
Comment (kalaninja) [reply crates/node/src/commands/run.rs:33]: yes, you are right. It is not needed for `bool` at all

File: crates/node/src/commands/run.rs (hunk 3 above)
Comment (kalaninja) [reply crates/node/src/commands/run.rs:34]: yes, makes a lot of sense. I like this approach more.

File: crates/node/src/service.rs [hunk 4]
```diff
@@ -586,30 +591,42 @@ where
...
+            Box::pin(sc_consensus_manual_seal::run_instant_seal(sc_consensus_manual_seal::InstantSealParams {
+                block_import,
+                env: proposer_factory,
+                client,
+                pool: transaction_pool,
+                select_chain,
+                consensus_data_provider: None,
+                create_inherent_data_providers,
+            }))
+        }
```
d-roak, line 628: For readability, it seems better to
```
SealingMode::Instant { finalize } => {
   let instant_seal_params = sc_consensus_manual_seal::InstantSealParams {
         block_import,
         env: proposer_factory,
         client,
         pool: transaction_pool,
         select_chain,
         consensus_data_provider: None,
         create_inherent_data_providers,
   })
   if (finalize) {
      Box::pin(sc_consensus_manual_seal::run_instant_seal_and_finalize(instant_seal_params)
   } else {
      Box::pin(sc_consensus_manual_seal::run_instant_seal(instant_seal_params)

   }
}
```

wdyt?

kalaninja: I don't like nesting `if` inside `match`, but removing code duplication is worth it.

d-roak: usually agree, but this case seems that the alternative is worse

Issue #1180, opened 2023-10-11
dev: RPC calls optimization


antiyro:
It appears that Madara recompute hashes for some rpc calls while it could (as the other full node clients for mainnet) fetch them from the mapping db to avoid recomputing them since pedersen hashing is some of the most expensive operations

As an example with Deoxys, for mainnet block 3000 it takes over 5.76 seconds to do a getBlockWithTransactionHashes via Madara (trough a vps) and 0.64 seconds via Pathfinder trough alchemy

`pathfinder (alchemy): 0,02s user 0,01s system 4% cpu 0,646 total seconds`
`madara (remote vps): 0,01s user 0,01s system 0% cpu 5,768 total seconds`

antiyro:
An example of what can happen on Starknet mainnet: https://voyager.online/tx/0x10b5ce52c65c4793763d773758bd83f9ed127f46f52d928e8aaf5a5eea35001

PR #1197, opened 2023-10-18
Add the `--cache` option, allows caching more information into the database

nils-mathieu:
This allows trading a bit of the memory footprint of Madara for better response time (specifically in the RPC client implementation).

# Pull Request type

This is a new feature.

Nothing is cached and transaction hashes are re-computed every time they are needed.

Resolves: #1180 

## What is the new behavior?

Transaction hashes are not stored along with block data in the database under the column `STARKNET_TRANSACTION_HASHES_CACHE`.

No.

## Other information

Currently, the flag is introduce as a simple boolean flag.

```Rust
    #[clap(long, action)]
    cache: bool,
```

In the future, it might be good to make this a vector of flags, which could optionally enable caching for only *some* parts of the database. Right now it's only for transaction hashes.

### DIFF
--- a/crates/client/db/src/lib.rs
+++ b/crates/client/db/src/lib.rs
@@ -19,7 +19,7 @@ mod meta_db;
 
 use std::marker::PhantomData;
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 
 use da_db::DaDb;
 use mapping_db::MappingDb;
@@ -38,14 +38,17 @@ struct DatabaseSettings {
 }
 
 pub(crate) mod columns {
-    pub const NUM_COLUMNS: u32 = 5;
+    /// Total number of columns.
+    // ===== /!\ ===================================================================================
+    // MUST BE INCREMENTED WHEN A NEW COLUMN IN ADDED
+    // ===== /!\ ===================================================================================
+    pub const NUM_COLUMNS: u32 = 6;
 
     pub const META: u32 = 0;
     pub const BLOCK_MAPPING: u32 = 1;
     pub const TRANSACTION_MAPPING: u32 = 2;
     pub const SYNCED_MAPPING: u32 = 3;
     pub const DA: u32 = 4;
-
     /// This column is used to map starknet block hashes to a list of transaction hashes that are
     /// contained in the block.
     ///
@@ -78,30 +81,33 @@ impl<B: BlockT> Backend<B> {
     /// Open the database
     ///
     /// The database will be created at db_config_dir.join(<db_type_name>)
-    pub fn open(database: &DatabaseSource, db_config_dir: &Path) -> Result<Self, String> {
-        Self::new(&DatabaseSettings {
-            source: match database {
-                DatabaseSource::RocksDb { .. } => {
-                    DatabaseSource::RocksDb { path: starknet_database_dir(db_config_dir, "rockdb"), cache_size: 0 }
-                }
-                DatabaseSource::ParityDb { .. } => {
-                    DatabaseSource::ParityDb { path: starknet_database_dir(db_config_dir, "paritydb") }
-                }
-                DatabaseSource::Auto { .. } => DatabaseSource::Auto {
-                    rocksdb_path: starknet_database_dir(db_config_dir, "rockdb"),
-                    paritydb_path: starknet_database_dir(db_config_dir, "paritydb"),
-                    cache_size: 0,
+    pub fn open(database: &DatabaseSource, db_config_dir: &Path, cache_more_things: bool) -> Result<Self, String> {
+        Self::new(
+            &DatabaseSettings {
+                source: match database {
+                    DatabaseSource::RocksDb { .. } => {
+                        DatabaseSource::RocksDb { path: starknet_database_dir(db_config_dir, "rockdb"), cache_size: 0 }
+                    }
+                    DatabaseSource::ParityDb { .. } => {
+                        DatabaseSource::ParityDb { path: starknet_database_dir(db_config_dir, "paritydb") }
+                    }
+                    DatabaseSource::Auto { .. } => DatabaseSource::Auto {
+                        rocksdb_path: starknet_database_dir(db_config_dir, "rockdb"),
+                        paritydb_path: starknet_database_dir(db_config_dir, "paritydb"),
+                        cache_size: 0,
+                    },
+                    _ => return Err("Supported db sources: `rocksdb` | `paritydb` | `auto`".to_string()),
                 },
-                _ => return Err("Supported db sources: `rocksdb` | `paritydb` | `auto`".to_string()),
             },
-        })
+            cache_more_things,
+        )
     }
 
-    fn new(config: &DatabaseSettings) -> Result<Self, String> {
+    fn new(config: &DatabaseSettings, cache_more_things: bool) -> Result<Self, String> {
         let db = db_opening_utils::open_database(config)?;
 
         Ok(Self {
-            mapping: Arc::new(MappingDb { db: db.clone(), write_lock: Arc::new(Mutex::new(())), _marker: PhantomData }),
+            mapping: Arc::new(MappingDb::new(db.clone(), cache_more_things)),
             meta: Arc::new(MetaDb { db: db.clone(), _marker: PhantomData }),
             da: Arc::new(DaDb { db: db.clone(), _marker: PhantomData }),
         })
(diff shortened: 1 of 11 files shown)


nils-mathieu:
I had to pass the `cache` boolean around, not sure if there's a better way to pass this kind of configuration around.

If the way I did it is the intended way, we probably need an intermediate structure named something like an `Arc<RuntimeConfig>` to pass around instead of a `bool`. That would be more clear, and way easier to extend later if more options are added.

crates/client/db/src/lib.rs
```diff
@@ -45,6 +45,12 @@ pub(crate) mod columns {
     pub const TRANSACTION_MAPPING: u32 = 2;
     pub const SYNCED_MAPPING: u32 = 3;
     pub const DA: u32 = 4;
+
+    /// This column is used to map starknet block hashes to a list of transaction hashes that are
+    /// contained in the block.
+    ///
+    /// This column should only be accessed if the `--cache` flag is enabled.
+    pub const STARKNET_TRANSACTION_HASHES_CACHE: u32 = 5;
```
tdelabro, line 53: You have to increment `NUM_COLUMNS` to 6 if you add a new column

crates/client/db/src/mapping_db.rs
```diff
@@ -85,14 +88,22 @@ impl<B: BlockT> MappingDb<B> {
 
         transaction.set(crate::columns::SYNCED_MAPPING, &commitment.block_hash.encode(), &true.encode());
 
-        for transaction_hash in commitment.starknet_transaction_hashes.into_iter() {
+        for transaction_hash in commitment.starknet_transaction_hashes.iter() {
```
tdelabro, line 91: moves
```rust
if commitment.cache {
            transaction.set(
                crate::columns::STARKNET_TRANSACTION_HASHES_CACHE,
                &commitment.starknet_block_hash.encode(),
                &commitment.starknet_transaction_hashes.encode(),
            );
        }
```

before your loop and you will retain the ability to iter by value rather than ref

crates/client/mapping-sync/src/lib.rs
```diff
@@ -70,6 +77,8 @@ impl<B: BlockT, C, BE, H> MappingSyncWorker<B, C, BE, H> {
             have_next: true,
             retry_times,
             sync_from,
+
```
tdelabro, line 80: empty line

crates/client/mapping-sync/src/sync_blocks.rs
```diff
@@ -88,14 +101,19 @@ where
...
     backend.mapping().write_hashes(mapping_commitment)?;
 
     Ok(())
 }
 
+/// # Arguments
+///
+/// - `cache`: whether more information should be cached when storing the block in the database.
 fn sync_one_block<B: BlockT, C, BE, H>(
+    cache: bool,
```
tdelabro, line 116: Rather than passing this arg in every intermediate func.

pass it to `MadaraBlackend::open` in `services.rs`
```rust
    let madara_backend = Arc::new(MadaraBackend::open(&config.database, &db_config_dir(config))?);
```

add it to this struct:
```rust
pub struct MappingDb<B: BlockT> {
    pub(crate) db: Arc<dyn Database<DbHash>>,
    pub(crate) write_lock: Arc<Mutex<()>>,
    pub(crate) _marker: PhantomData<B>,
}
```

So now you can access it any time you want

crates/client/db/src/mapping_db.rs
```diff
@@ -112,4 +123,25 @@ impl<B: BlockT> MappingDb<B> {
...
+    ///
+    /// * `starknet_hash` - the hash of the starknet block to search for.
+    ///
+    /// # Returns
+    ///
+    /// The list of transaction hashes.
+    ///
+    /// This function may return `None` for two separate reasons:
+    ///
+    /// - The cache is disabled.
```
tdelabro, line 139: By adding the `cache` bool to the mappingdb struct directly, you can now perform a check here, and return an error with message if the cache is disable, rather than `None`

crates/client/rpc/src/lib.rs
```diff
@@ -401,10 +413,20 @@ where
...
         let block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).unwrap_or_default();
         let chain_id = self.chain_id()?;
-
-        let transactions =
-            block.transactions_hashes::<H>(Felt252Wrapper(chain_id.0)).into_iter().map(FieldElement::from).collect();
         let blockhash = block.header().hash::<H>();
+
+        let a: H256 = Felt252Wrapper(FieldElement::MAX).into();
+
+        let transactions = self
```
tdelabro, line 420: transaction_hashes

crates/client/rpc/src/lib.rs
```diff
@@ -401,10 +413,20 @@ where
...
+            .backend
+            .mapping()
+            .cached_transaction_hashes_from_block_hash(blockhash.into())
+            .unwrap_or_else(|err| {
+                error!("{err}");
+                None
+            })
```
tdelabro, line 427: use your `get_cached_transaction_hashes` fn

crates/client/rpc/src/lib.rs
```diff
@@ -401,10 +413,20 @@ where
...
+            .map(|vec| vec.into_iter().map(|val| Felt252Wrapper::try_from(val).unwrap().0).collect())
```
tdelabro, line 428: use your `h256_to_felt` fn

crates/client/rpc/src/lib.rs
```diff
@@ -649,6 +676,16 @@ where
         let chain_id = self.chain_id()?;
         let chain_id = Felt252Wrapper(chain_id.0);
 
+        let transaction_hashes = self.get_cached_transaction_hashes(block.header().hash::<H>().into());
+        let mut transactions = Vec::with_capacity(block.transactions().len());
+        for (index, tx) in block.transactions().iter().enumerate() {
+            let hash = transaction_hashes
+                .as_ref()
+                .map(|vec| h256_to_felt(*vec.get(index).ok_or(StarknetRpcApiError::InternalServerError)?))
```
tdelabro, line 684: avoid calling a Vec variable "vec".
I'm fine with `v` but something more expressive like `tx_hashes` is better

crates/node/src/commands/run.rs
```diff
@@ -29,6 +29,14 @@ pub struct ExtendedRunCmd {
     /// Choose a supported DA Layer
     #[clap(long)]
     pub da_layer: Option<DaLayer>,
+
+    /// When enabled, more information about the blocks and their transaction is cached and stored
+    /// in the database.
+    ///
+    /// This may improve response times for RPCs that require that information, but it also
+    /// increases the memory footprint of the node.
+    #[clap(long, action)]
```
tdelabro, line 38: what does `action` do ?

tdelabro: We want it to be false if not passed through CLI, is it what it does?

tdelabro (changes requested):

crates/client/mapping-sync/src/sync_blocks.rs
nils-mathieu: That's what I had asked you; you told me to pass it through the `MappingCommitment`. I think we misunderstood ourselves. I'll change this.

crates/client/db/src/mapping_db.rs
nils-mathieu: I don't think that's a good idea to return an error if the cache is disabled. The idea was to abstract the cache as much as possible.

If for some reason the value is not in the cache, we probably want to recompute it rather than returning an error an refusing to respond. And if the cache is disabled, the value will
simply not be in the cache at all. I like those two cases having the same apparent behavior (`None` being returned)

crates/node/src/commands/run.rs
nils-mathieu: `action` on `bool` makes it so that when the flag is automatically `false` by default and then replaced by `true` when the flag is set. I just checked the doc and apparently it's automatically on for `bool` now. I don't think that was the case before.

Do I remove this?

crates/client/db/src/mapping_db.rs
nils-mathieu: Why do you want to iter by value? The value gets encoded anyway and `encode` takes a shared reference. I'm pretty sure that iterating by value is slower anyway because of the `H256` copies. (to be fair, both are probably optimized the same but heh)

tdelabro: no specific reason. And your are right that it won't change the perf

crates/client/db/src/mapping_db.rs
tdelabro: Okay to return `None` in both ways. But you should still check the bool, it will avoid you the whole IO reading work

nils-mathieu: oh right. Didn't think about that.
Good catch.

tdelabro (approved):

PR #1205, opened 2023-10-20
L1 Event Listener

### LINKED ISSUE #1177: create a l1_event_listener lib crate (or some other name)

bidzyyys:
Closes #1177


crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Clone, PartialEq, serde::Deserialize, Debug)]
+pub struct L1MessagesWorkerConfig {
+    #[serde(default = "default_http")]
+    pub http_provider: String,
+    #[serde(default = "default_contract_address")]
+    pub contract_address: String,
+}
+
+impl TryFrom<&PathBuf> for L1MessagesWorkerConfig {
+    type Error = String;
```
tdelabro, line 16: For errors: string < anyhow < dedicated struct

I think for this we should at least use anyhow. But maybe having a dedicated struct will be better. Maybe it's not worth it, depending on the way this error is later handled

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,36 @@
+use std::fs::File;
+use std::path::PathBuf;
+
+pub const DEFAULT_ETHEREUM_NODE: &str = "127.0.0.1:8545";
+pub const DEFAULT_CONTRACT_ADDRESS: &str = "0x5fbdb2315678afecb367f032d93f642f64180aa3";
+
+#[derive(Clone, PartialEq, serde::Deserialize, Debug)]
+pub struct L1MessagesWorkerConfig {
+    #[serde(default = "default_http")]
+    pub http_provider: String,
```
tdelabro, line 10: You should probably use the `Url` type rather than a string there. Unless there is some reason not to downstream.

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+    #[serde(default = "default_contract_address")]
+    pub contract_address: String,
```
tdelabro, line 12: the `ether` libs is already imported and expose an `Address` type.
https://docs.rs/ethers/latest/ethers/abi/type.Address.html

tdelabro: For more security when people will be using non-default addresses it's probably better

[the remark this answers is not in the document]
tdelabro: Also add a `new` method in order to allow people to create it without reading from disk

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+impl TryFrom<&PathBuf> for L1MessagesWorkerConfig {
```
tdelabro, line 15: I'm not a big fan of using `TryFrom` for this because it includes io reading.
Nothing explicitly prohibits it in the Rust language, but having a dedicated `new_from_file` method is more explicit imo

crates/client/l1-messages/src/contract.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn try_from(event: &LogMessageToL2Filter) -> Result<Self, Self::Error> {
+        // Check against panic
+        // https://docs.rs/ethers/latest/ethers/types/struct.U256.html#method.as_u128
+        if event.fee > U256::from_big_endian(&(u128::MAX.to_be_bytes())) {
+            Err(L1MessagesWorkerError::ToFeeError)
+        } else {
+            Ok(Fee(event.fee.as_u128()))
+        }
+    }
+}
```
tdelabro, line 28: This look more like a getter than a real Into.

I would either:
- impl it as a getter on LogMessageToL2Filter
or
- imp `TryFrom` once: `TryFrom<LogMessageToL2Filter> for (Fee, HandleL1MessageTransaction)`

You call both one after the other anyway.

crates/client/l1-messages/src/contract.rs
```diff
@@ -0,0 +1,53 @@
...
+        let contract_address = Felt252Wrapper::try_from(sp_core::U256(event.to_address.0))?;
+
+        // Function of the contract to call.
+        let entry_point_selector = Felt252Wrapper::try_from(sp_core::U256(event.selector.0))?;
+
+        // L1 message nonce.
+        let nonce: u64 = Felt252Wrapper::try_from(sp_core::U256(event.nonce.0))?.try_into()?;
+
+        // Add the from address here so it's directly in the calldata.
+        let mut calldata: Vec<Felt252Wrapper> = Vec::from([Felt252Wrapper::try_from(event.from_address.as_bytes())?]);
```
tdelabro, line 44: Why do we need to do that? I know it was in the previous implementation. But I can't tell why

tdelabro: It may have been a custom addition that is not required. Wdyt @EvolveArt?

crates/client/l1-messages/src/contract.rs
```diff
@@ -0,0 +1,53 @@
...
+
+        for x in &event.payload {
+            calldata.push(Felt252Wrapper::try_from(sp_core::U256(x.0))?);
+        }
+
+        let tx = HandleL1MessageTransaction { nonce, contract_address, entry_point_selector, calldata };
+        Ok(tx)
```
tdelabro, line 51: You can one-line this

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,20 @@
+use mp_felt::Felt252WrapperError;
+
+#[warn(dead_code)]
+#[derive(Debug, Eq, PartialEq)]
+pub enum L1MessagesWorkerError {
```
tdelabro, line 5: It should impl `Error`. Looks like a perfect place to use `thiserror`

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,20 @@
...
+    SubmitTxError,
+    ToFeeError,
+    ToTransactionError,
+}
+
+impl From<Felt252WrapperError> for L1MessagesWorkerError {
+    fn from(_e: Felt252WrapperError) -> Self {
+        L1MessagesWorkerError::ToTransactionError
+    }
+}
```
tdelabro, line 20: Isn't this a bit weird? Lot of information is lost

crates/client/l1-messages/src/worker.rs
```diff
@@ -0,0 +1,192 @@
...
+    let l1_contract = match connect_to_l1_contract(&config) {
+        Ok(l1_contract) => l1_contract,
+        Err(e) => {
+            log::error!("⟠ Unexpected error while connecting to L1: {:?}", e);
+            return;
+        }
+    };
+
+    let events = l1_contract.events().from_block(get_block_number::<B, S>(&offchain_storage).unwrap_or_default());
+    let mut stream = match events.stream().await {
```
tdelabro, line 66: `events` is an iterator, why do you make it a `Stream`? It will never be not ready, it will always be either ready or exhausted

crates/client/l1-messages/src/worker.rs
```diff
@@ -0,0 +1,192 @@
...
+where
+    B: BlockT,
+    C: ProvideRuntimeApi<B> + HeaderBackend<B>,
+    C::Api: StarknetRuntimeApi<B> + ConvertTransactionRuntimeApi<B>,
+    P: TransactionPool<Block = B> + 'static,
+    S: Backend<B> + Send + 'static,
+{
+    log::info!("⟠ Starting L1 Messages Worker with settings: {:?}", config);
+
+    let mut offchain_storage = match backend.offchain_storage() {
```
tdelabro, line 49: Why did you choose to use substrate `offchain_storage` rather than our `MadaraBackend` storage? 

The main upside of this is that this storage is also accessible from inside the runtime through `offchain_workers`, but you removed our worker in this PR.

So I don't think it should be used there. We should go with a client-only db there

crates/node/src/service.rs
```diff
@@ -527,6 +533,14 @@ pub fn new_full(
         );
     }
 
+    if let Some(ref l1_messages_config) = l1_messages {
+        let config = L1MessagesWorkerConfig::try_from(l1_messages_config)?;
+        task_manager.spawn_essential_handle().spawn(
+            "L1 Messages",
+            Some(MADARA_TASK_GROUP),
+            mc_l1_messages::worker::run_worker(config, client, transaction_pool, db_backend),
```
tdelabro, line 541: I think there is a small misunderstood there.
`task_manager.spawn_essential_handle().spawn` will spawn a task, once. The task will only be spawn once. When the function you are calling, here `mc_l1_messages::worker::run_worker` returns, the thread will die.

The design we want here is that this function you call should never return. It should contain an infinite loop, asynchronously pulling blocks from Ethereum. 

The way I read your impl of `run_worker` it will return after processing all the events since the last time it was called. But it's only called once at the node launch.

Instead you should call `ether.events()` inside an infinite loop

tdelabro (changes requested):

crates/node/src/service.rs
bidzyyys: @tdelabro this implementation provides an infinite loop. The way an event stream works is that it listens to specific events and does not end. So you do not need to worry here :)

bidzyyys: @tdelabro the only way this function can return earlier is potential issues with connection to L1 Contract or storage.

crates/client/l1-messages/src/worker.rs
bidzyyys: @tdelabro as I understand correctly, an iterator may end after processing all existing events, but `Stream` provides you with a way to handle possible future events. I tested `Stream`, and it does not break the listening loop and always reacts to an incoming event.

crates/client/l1-messages/src/config.rs
bidzyyys: I applied attitude from `data-availability` lib but will rewrite into `anyhow`

crates/client/l1-messages/src/worker.rs
tdelabro: So it will end up yielding `None` forever

crates/node/src/service.rs
tdelabro: But your infinite loop will end up yielding `None` forever once the first set of events is processed. It will never catch up to later l1 blocks

crates/client/l1-messages/src/worker.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    let l1_contract = match connect_to_l1_contract(&config) {
+        Ok(l1_contract) => l1_contract,
+        Err(e) => {
+            log::error!("⟠ Unexpected error while connecting to L1: {:?}", e);
+            return;
+        }
+    };
+
+    let events = l1_contract.events().from_block(get_block_number::<B, S>(&offchain_storage).unwrap_or_default());
```
tdelabro, line 65: here `events` is a finite iterator with a finite length. You make it a stream, so once we went through this finite number of events, it will just yield `None` forever.

That is not the behavior we want. We want it to keep looking for new l1 blocks and process their events

crates/node/src/service.rs
bidzyyys: I disagree, it catches later (future) blocks

crates/client/l1-messages/src/worker.rs
bidzyyys: Not forever, once a new block with proper event comes, it handles event. Tested manually.

bidzyyys: And once you want to stop the node with Ctrl+C, it stops as expected

crates/client/l1-messages/src/worker.rs
bidzyyys: @tdelabro but it keeps looking for new blocks...

crates/client/l1-messages/src/contract.rs
bidzyyys: Ok

crates/client/l1-messages/src/contract.rs
bidzyyys: Removed now, does not have sense to me too.

crates/client/l1-messages/src/config.rs
bidzyyys: Sure!

crates/client/l1-messages/src/config.rs
bidzyyys: `ethers` api prefers String.

crates/client/l1-messages/src/error.rs
bidzyyys: Done

crates/client/db/src/lib.rs
```diff
@@ -45,11 +47,13 @@ pub(crate) mod columns {
     pub const TRANSACTION_MAPPING: u32 = 2;
     pub const SYNCED_MAPPING: u32 = 3;
     pub const DA: u32 = 4;
+    pub const MESSAGING: u32 = 5;
```
tdelabro, line 50: If you add a column you have to increment `pub const NUM_COLUMNS: u32 = 5;`

crates/client/db/src/messaging_db.rs
```diff
@@ -0,0 +1,34 @@
...
+use sp_runtime::traits::Block as BlockT;
+
+use crate::DbHash;
+
+pub struct MessagingDb<B: BlockT> {
+    pub(crate) db: Arc<dyn Database<DbHash>>,
+    pub(crate) _marker: PhantomData<B>,
+}
+
+// TODO: purge old cairo job keys
```
tdelabro, line 16: What does that mean?

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,35 @@
+use std::fs::File;
+use std::path::PathBuf;
+
+use ethers::types::Address;
+
+use crate::error::L1MessagesConfigError;
+
+pub const DEFAULT_ETHEREUM_NODE: &str = "127.0.0.1:8545";
+pub const DEFAULT_CONTRACT_ADDRESS: &str = "0x5fbdb2315678afecb367f032d93f642f64180aa3";
```
bidzyyys, line 9: To be removed with another commit with CR changes

crates/client/l1-messages/src/contract.rs
```diff
@@ -40,14 +38,12 @@ impl TryFrom<&LogMessageToL2Filter> for HandleL1MessageTransaction {
         // L1 message nonce.
         let nonce: u64 = Felt252Wrapper::try_from(sp_core::U256(event.nonce.0))?.try_into()?;
 
-        // Add the from address here so it's directly in the calldata.
-        let mut calldata: Vec<Felt252Wrapper> = Vec::from([Felt252Wrapper::try_from(event.from_address.as_bytes())?]);
+        let mut calldata = Vec::<Felt252Wrapper>::new();
```
tdelabro, line 41: You can use the iterator and map syntax. More readable here

crates/client/l1-messages/src/error.rs
```diff
@@ -1,20 +1,37 @@
 use mp_felt::Felt252WrapperError;
 
-#[warn(dead_code)]
-#[derive(Debug, Eq, PartialEq)]
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
+pub enum L1MessagesConfigError {
+    #[error("File with L1 Messages Worker config not found")]
+    FileNotFound(String),
```
tdelabro, line 6: The content of the error (the String) is not used by display. Why?

crates/client/l1-messages/src/error.rs
```diff
@@ -1,20 +1,37 @@
...
+    #[error("Failed to deserialize L1 Messages Worker Config from config file")]
```
tdelabro, line 7: The content of the error (the String) is not used by display. Why?

crates/client/l1-messages/src/error.rs
```diff
@@ -1,20 +1,37 @@
...
+}
+
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
 pub enum L1MessagesWorkerError {
+    #[error("Failed to initialize L1 Messages Worker based on provided Config")]
     ConfigError,
+    #[error("Failed to convert transaction via Runtime API")]
     ConvertTransactionRuntimeApiError,
-    MessageAlreadyProcessed,
+    #[error("Madara Messaging DB Error")]
```
tdelabro, line 17: Same

crates/client/l1-messages/src/error.rs
```diff
@@ -1,20 +1,37 @@
...
+    #[error("Failed to read/write into Offchain Storage")]
     OffchainStorageError,
+    #[error("Failed to use Runtime API")]
     RuntimeApiError,
+    #[error("Failed to submit transaction into Transaction Pool")]
     SubmitTxError,
+    #[error("Failed to convert L1 Message into Fee")]
     ToFeeError,
-    ToTransactionError,
+    #[error("Failed to convert L1 Message into L2 Transaction")]
```
tdelabro, line 29: Same

crates/client/l1-messages/src/error.rs
```diff
@@ -1,20 +1,37 @@
...
+    ToTransactionError(String),
 }
 
 impl From<Felt252WrapperError> for L1MessagesWorkerError {
-    fn from(_e: Felt252WrapperError) -> Self {
-        L1MessagesWorkerError::ToTransactionError
+    fn from(e: Felt252WrapperError) -> Self {
+        L1MessagesWorkerError::ToTransactionError(format!("{:?}", e))
```
tdelabro, line 35: You can just store `e` directly rather than formatting it. And relly on your Error impl on `L1MessagesWorkerError` to do the formating

crates/client/l1-messages/src/worker.rs
```diff
@@ -18,51 +15,50 @@ use crate::contract::{L1Contract, LogMessageToL2Filter};
...
+        Ok(blknum) => blknum,
         Err(e) => {
-            log::error!("⟠ Unexpected error while connecting to L1: {:?}", e);
+            log::error!("⟠ Madara Messaging DB unavailable: {:?}", e);
             return;
         }
     };
 
-    let events = l1_contract.events().from_block(get_block_number::<B, S>(&offchain_storage).unwrap_or_default());
+    let events = l1_contract.events().from_block(last_synced_l1_block);
```
tdelabro, line 61:
```
let events_stream = l1_contract.events().from_block(last_synced_l1_block).steam();
```

It will be more expressive than this one that I struggled to understand.

tdelabro: you can even use https://docs.rs/ethers/latest/ethers/middleware/contract/struct.Event.html#method.stream_with_meta directly

crates/client/l1-messages/src/worker.rs
```diff
@@ -107,16 +102,18 @@ where
 
 async fn process_l1_message<C, P, B>(
     event: &LogMessageToL2Filter,
-    client: Arc<C>,
-    pool: Arc<P>,
+    client: &Arc<C>,
+    pool: &Arc<P>,
+    backend: &Arc<mc_db::Backend<B>>,
+    l1_blknum: &u64,
```
tdelabro, line 108: `l1_blknum` spell the whole works. This abreviation is not easy to figure out

crates/node/src/service.rs
```diff
@@ -534,11 +532,12 @@ pub fn new_full(
     }
 
     if let Some(ref l1_messages_config) = l1_messages {
-        let config = L1MessagesWorkerConfig::try_from(l1_messages_config)?;
+        let config = L1MessagesWorkerConfig::new_from_file(l1_messages_config)
+            .map_err(|e| sc_service::error::Error::Other(format!("{:?}", e)))?;
         task_manager.spawn_essential_handle().spawn(
             "L1 Messages",
```
tdelabro, line 538: try to respect the same kind onf name that is used by the other tasks. uncapitalized `-` separated

tdelabro (changes requested):

crates/client/db/src/messaging_db.rs
bidzyyys: Copy-paste artefact...

crates/client/l1-messages/src/worker.rs
bidzyyys: Unfortunately, the compiler says `creates a temporary value which is freed while still in use` for `l1_contract.events().from_block(last_synced_l1_block).stream().await` . We need to have `events` assigned just like in this example - https://docs.rs/ethers/latest/ethers/middleware/contract/struct.Event.html#method.stream

crates/client/l1-messages/src/contract.rs
bidzyyys: It would be more readable but `map()` requires `FnMut` that is not fallible... 
We need here error handling for ->  `Felt252Wrapper::try_from(sp_core::U256(x.0))?` 
It would be great if you know better solution...

bidzyyys: Found interesting workaround -- https://www.justanotherdot.com/posts/error-handling-with-closures-in-iterators.html

tdelabro: Yep that is the way to go. You collect your `Iterator<Result<V, E>>` into a `Result<Vec<V>, E>` it will early exit at the first error encountered

crates/client/l1-messages/src/config.rs
tdelabro: https://docs.rs/ethers/2.0.11/ethers/providers/struct.Http.html#method.new
No they seems to be encouraging the use of `Url`

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,35 @@
...
+        serde_json::from_reader(file).map_err(|_e| L1MessagesConfigError::InvalidFile(format!("{:?}", path)))
+    }
+
+    pub fn get_provider(&self) -> &String {
+        &self.http_provider
+    }
+
+    pub fn get_contract_address(&self) -> &Address {
+        &self.contract_address
+    }
```
tdelabro, line 34: for this kind of getters, you can remove `get` from the name and just go with the name of the value you are reading

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,37 @@
+use mp_felt::Felt252WrapperError;
+
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
+pub enum L1MessagesConfigError {
+    #[error("File with L1 Messages Worker config not found: `{0}`")]
+    FileNotFound(String),
```
tdelabro, line 6: Keep the error in the variant. It's an std::fs error I think. And relly on display for the conversion to string

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[error("Failed to deserialize L1 Messages Worker Config from config file: `{0}`")]
+    InvalidFile(String),
```
tdelabro, line 8: Same. This is a `serde_json::Error`. Keep it as it is given to you.
Use `#[from]` to do conversion

tdelabro: I think you can even make those two transparent.

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[error("Failed to read/write into Offchain Storage")]
+    OffchainStorageError,
+    #[error("Failed to use Runtime API")]
+    RuntimeApiError,
+    #[error("Failed to submit transaction into Transaction Pool")]
+    SubmitTxError,
+    #[error("Failed to convert L1 Message into Fee")]
+    ToFeeError,
+    #[error("Failed to convert L1 Message into L2 Transaction: `{0}`")]
+    ToTransactionError(String),
```
tdelabro, line 30: Keep the internal value an error rather than a string

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
+pub enum L1MessagesWorkerError {
+    #[error("Failed to initialize L1 Messages Worker based on provided Config")]
+    ConfigError,
+    #[error("Failed to convert transaction via Runtime API")]
+    ConvertTransactionRuntimeApiError,
+    #[error("Madara Messaging DB Error: `{0}`")]
+    DatabaseError(String),
```
tdelabro, line 18: keep it a error rather than a String

crates/client/l1-messages/src/error.rs
tdelabro: Store the error in the error strut, and let the caller decide what he wants to do with it. Maybe he wants to match it and not print it. By converting to string here you make his job more difficult

crates/node/src/service.rs
```diff
@@ -527,6 +531,15 @@ pub fn new_full(
         );
     }
 
+    if let Some(ref l1_messages_config) = l1_messages {
+        let config = L1MessagesWorkerConfig::new_from_file(l1_messages_config)
+            .map_err(|e| sc_service::error::Error::Other(e.to_string()))?;
+        task_manager.spawn_essential_handle().spawn(
+            "l1-messaging",
```
tdelabro, line 538: "l1-messaging` can be understood in both way l1 to l2 or l2 to l1. We want to be more explicit "ethereum-core-contract-events-listener"

tdelabro: or something along those lines

examples/rpc/starknet/starknet_l1NonceUnused.hurl
```diff
@@ -0,0 +1,13 @@
+POST {{RPC_URL}}
```
tdelabro, line 1: What is this for? I'm not familiar with `.hurl` format.
`starknet_l1NonceUnused` neither with this rpc route. Where does it come from?

examples/messaging/messaging.json
```diff
@@ -0,0 +1,4 @@
+{
```
tdelabro, line 1: Can you add comments to this?
I have no idea what it is an example of

crates/client/rpc-core/src/lib.rs
```diff
@@ -135,4 +135,8 @@ pub trait StarknetRpcApi {
     /// Returns the receipt of a transaction by transaction hash.
     #[method(name = "getTransactionReceipt")]
     fn get_transaction_receipt(&self, transaction_hash: FieldElement) -> RpcResult<MaybePendingTransactionReceipt>;
+
+    /// Checks if L1 Message Nonce has not been used
+    #[method(name = "l1NonceUnused")]
+    fn l1_nonce_unused(&self, nonce: u64) -> RpcResult<bool>;
```
tdelabro, line 141: Why do we need to expose this to external users as an RPC route?

crates/client/rpc/src/lib.rs
```diff
@@ -454,6 +454,17 @@ where
...
     }
 
+    /// Checks if L1 Nonce has been used.
+    fn l1_nonce_unused(&self, nonce: u64) -> RpcResult<bool> {
+        let best_block_hash = self.client.info().best_hash;
+        let unused = self
+            .client
+            .runtime_api()
+            .l1_nonce_unused(best_block_hash, nonce)
+            .map_err(|_| StarknetRpcApiError::InternalServerError)?;
```
tdelabro, line 464: log the error with `log::error` before returning the internal one

crates/client/rpc/src/lib.rs
```diff
@@ -454,6 +454,17 @@ where
         Ok(Felt(chain_id.0))
     }
 
+    /// Checks if L1 Nonce has been used.
+    fn l1_nonce_unused(&self, nonce: u64) -> RpcResult<bool> {
+        let best_block_hash = self.client.info().best_hash;
+        let unused = self
```
tdelabro, line 460: `is_unused`

crates/node/src/cli.rs
```diff
@@ -51,6 +51,9 @@ pub struct ExtendedRunCmd {
 
     #[clap(long)]
     pub da_layer: Option<DaLayer>,
+
+    #[clap(long, short = 'm')]
+    pub l1_messages: Option<PathBuf>,
```
tdelabro, line 56: `l1_messages_worker_config_file` isn't it?
Remove the `short` also

tdelabro: You can use https://docs.rs/clap/latest/clap/enum.ValueHint.html with ValueHint::FilePath

tdelabro: Also, I would rather have the possibility to either:
- give a path to a config file
- give two arguments `l1_contract_address` and `provider_url`

Both should be mutually exclusive. You can achieve this with clap args groups, as I did in  crates/node/src/commands/setup.rs with `struct SetupSource` using `conflicts_with`

crates/node/src/command.rs
```diff
@@ -292,9 +292,22 @@ pub fn run() -> sc_cli::Result<()> {
...
 
+            let l1_messages_config = match cli.run.l1_messages {
+                Some(ref config_path) => {
+                    if !config_path.exists() {
+                        return Err("L1 Messages Worker config file not available".into());
+                    }
+                    Some(config_path.clone())
+                }
+                None => {
+                    log::info!("madara initialized w/o L1 Messages Worker");
```
tdelabro, line 303: I think this should be a `warning` for any madara node which is not running in `dev` mode

crates/node/src/service.rs
```diff
@@ -527,6 +531,15 @@ pub fn new_full(
         );
     }
 
+    if let Some(ref l1_messages_config) = l1_messages {
```
tdelabro, line 534: Should it be spawned before or after aura?

crates/node/src/service.rs
```diff
@@ -527,6 +531,15 @@ pub fn new_full(
...
+        let config = L1MessagesWorkerConfig::new_from_file(l1_messages_config)
+            .map_err(|e| sc_service::error::Error::Other(e.to_string()))?;
+        task_manager.spawn_essential_handle().spawn(
```
tdelabro, line 537: `spawn_essential_handle` make it so that if this thread returns or panic, the whole node is shut down.
Imo this should not be the case. If we allow nodes to run without l1-worker we can allow the  service to go down without shuting down everything

tdelabro: There is another method for non-essential services

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -56,17 +55,23 @@ fn given_hardcoded_contract_run_invoke_tx_then_it_works() {
 
         let transaction: InvokeTransaction = get_invoke_dummy(Felt252Wrapper::ZERO).into();
 
-        let tx = Message {
-            topics: vec![
-                "0xdb80dd488acf86d17c747445b0eabb5d57c541d3bd7b6b87af987858e5066b2b".to_owned(),
-                "0x0000000000000000000000000000000000000000000000000000000000000001".to_owned(),
-                "0x0000000000000000000000000000000000000000000000000000000000000001".to_owned(),
-                "0x01310e2c127c3b511c5ac0fd7949d544bb4d75b8bc83aaeb357e712ecf582771".to_owned(),
+        let tx = HandleL1MessageTransaction {
```
tdelabro, line 58: Maybe the name of this tests is not as explicit as it should be. Why does it involve consuming a l1 message when it's about `invoke` txs?

crates/runtime/src/lib.rs
```diff
@@ -397,6 +397,16 @@ impl_runtime_apis! {
 
             (messages_to_l1, messages_to_l2)
         }
+
+        fn l1_nonce_unused(nonce: u64) -> bool {
```
tdelabro, line 401: I would rather have the conversion being done outside of the runtime.
Your signature should be `fn l1_nonce_unused(nonce: Nonce) -> bool`

crates/runtime/src/lib.rs
```diff
@@ -397,6 +397,16 @@ impl_runtime_apis! {
...
+            StarkFelt::try_from(nonce).map_or_else(
```
tdelabro, line 402: Also, StarkFelt from u64 is not something that can fail. Any u64 can be stored into a Felt. I think there is another way to achieve this conversion, without the possibility of erroring

crates/client/l1-messages/src/error.rs
tdelabro: use `#[from]`

crates/client/l1-messages/src/contract.rs
```diff
@@ -0,0 +1,49 @@
...
+        // https://docs.rs/ethers/latest/ethers/types/struct.U256.html#method.as_u128
+        if self.fee > U256::from_big_endian(&(u128::MAX.to_be_bytes())) {
+            Err(L1MessagesWorkerError::ToFeeError)
+        } else {
+            Ok(Fee(self.fee.as_u128()))
+        }
+    }
+}
+
+impl TryFrom<&LogMessageToL2Filter> for HandleL1MessageTransaction {
```
tdelabro, line 28: Why a ref rather than taking ownership?

crates/client/l1-messages/src/contract.rs
```diff
@@ -0,0 +1,49 @@
...
+    type Error = L1MessagesWorkerError;
```
tdelabro, line 29: I would create an error enum dedicated to this conversion, with different variants:
- InvalidContractAddress
- InvalidEntryPointSelector
- InvalidNonce
- InvalidCalldata

Each one containing the type of error that could have cause it

crates/client/l1-messages/src/contract.rs
```diff
@@ -0,0 +1,49 @@
...
+    pub fn try_get_fee(&self) -> Result<Fee, L1MessagesWorkerError> {
+        // Check against panic
+        // https://docs.rs/ethers/latest/ethers/types/struct.U256.html#method.as_u128
+        if self.fee > U256::from_big_endian(&(u128::MAX.to_be_bytes())) {
+            Err(L1MessagesWorkerError::ToFeeError)
+        } else {
+            Ok(Fee(self.fee.as_u128()))
+        }
+    }
+}
```
tdelabro, line 26: I would not impl this on `LogMessageToL2Filter`. Just let the caller read the `pub` field and decide to do the conversion itself.

This is too specific. It only fulfills one specific use case you have (and you only have once) but it's not something that should be implemented on the type directly

crates/client/db/src/lib.rs
```diff
@@ -38,18 +40,20 @@ struct DatabaseSettings {
...
 }
 
 pub mod static_keys {
     pub const CURRENT_SYNCING_TIPS: &[u8] = b"CURRENT_SYNCING_TIPS";
     pub const LAST_PROVED_BLOCK: &[u8] = b"LAST_PROVED_BLOCK";
+    pub const L1_MESSAGES: &[u8] = b"L1_MESSAGES";
```
tdelabro, line 56: Shouldn't it be named `LAST_SYNCED_L1_BLOCK`

tdelabro (changes requested):

crates/client/l1-messages/src/error.rs
bidzyyys: Fair point, but in this case errors in most of `db` implementations are `String` type.

crates/client/l1-messages/src/error.rs
bidzyyys: Fair point, but in this case errors in most of `db` implementations are `String` type.

crates/client/rpc-core/src/lib.rs
bidzyyys: Not needed in this implementation, will remove it.

examples/rpc/starknet/starknet_l1NonceUnused.hurl
bidzyyys: Will remove the file as this is test for https://github.com/keep-starknet-strange/madara/pull/1205#discussion_r1403780855

crates/client/rpc/src/lib.rs
bidzyyys: To be removed as https://github.com/keep-starknet-strange/madara/pull/1205#discussion_r1403780855

crates/client/rpc/src/lib.rs
bidzyyys: To be removed as https://github.com/keep-starknet-strange/madara/pull/1205#discussion_r1403780855

crates/node/src/command.rs
bidzyyys: It can be a `warning` anytime.

crates/pallets/starknet/src/tests/invoke_tx.rs
bidzyyys: I have no idea why it was implemented this way but I can remove consuming part.

crates/node/src/service.rs
bidzyyys: I do not see any valid point for spawning this service before `aura` that is responsible for block production. In my opinion it should start after `aura` while this thread submits transactions to be proceeded.

crates/node/src/service.rs
bidzyyys: Followed instructions from https://github.com/keep-starknet-strange/madara/issues/1177 but will refactor into `spawn_handle`

crates/runtime/src/lib.rs
bidzyyys: Regular `StarkFelt::from` works.

crates/runtime/src/lib.rs
bidzyyys: Sure!

crates/client/l1-messages/src/contract.rs
bidzyyys: Fair point!

examples/messaging/messaging.json
bidzyyys: Correct me if I am wrong but I think that comments are not available in JSON format.
This file is an example of l1 messaging worker configuration (renamed file).

crates/client/l1-messages/src/error.rs
tdelabro: This is a bad design. Try changing those method signatures so we can use the actual errors

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[error("Failed to convert Entrypoint Selector from L1 Event: `{0}`")]
+    InvalidEntryPointSelector(Felt252WrapperError),
+    #[error("Failed to convert Nonce param from L1 Event: `{0}`")]
+    InvalidNonce(Felt252WrapperError),
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum L1MessagesConfigError {
+    #[error("File with L1 Messages Worker config not found")]
+    FileNotFound(#[from] std::io::Error),
```
tdelabro, line 18: `"File with L1 Messages Worker config not found: {0}"`

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[error("Failed to deserialize L1 Messages Worker Config from config file")]
```
tdelabro, line 19: `"File with L1 Messages Worker config not found: {0}"`

we can use a bit more information on why it failed

crates/runtime/src/lib.rs
```diff
@@ -356,9 +357,51 @@ impl_runtime_apis! {
...
+        fn get_tx_messages_to_l1(tx_hash: TransactionHash) -> Vec<MessageToL1> {
+            Starknet::tx_messages(tx_hash)
+        }
+
         fn get_tx_execution_outcome(tx_hash: TransactionHash) -> Option<Vec<u8>> {
-           Starknet::tx_revert_error(tx_hash).map(|s| s.into_bytes())
+            Starknet::tx_revert_error(tx_hash).map(|s| s.into_bytes())
+        }
+
+        fn get_starknet_messages(block_extrinsics: Vec<<Block as BlockT>::Extrinsic>, chain_id: Felt252Wrapper) -> (Vec<MessageL2ToL1>, Vec<MessageL1ToL2>) {
```
tdelabro, line 368: `extract_starknet_messages` rather than `get`?

crates/pallets/starknet/src/lib.rs
```diff
@@ -238,6 +215,11 @@ pub mod pallet {
     #[pallet::getter(fn tx_events)]
     pub(super) type TxEvents<T: Config> = StorageMap<_, Identity, TransactionHash, Vec<StarknetEvent>, ValueQuery>;
 
+    #[pallet::storage]
+    #[pallet::unbounded]
+    #[pallet::getter(fn tx_messages)]
+    pub(super) type TxMessages<T: Config> = StorageMap<_, Identity, TransactionHash, Vec<MessageToL1>, ValueQuery>;
+
```
tdelabro, line 222: Where does this come from? Your PR is about l1->l2 messages, why do you have to bother with the l2->l1 messages?

crates/pallets/starknet/src/lib.rs
```diff
@@ -1033,6 +980,53 @@ impl<T: Config> Pallet<T> {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `call_info` — A ref to the call info structure.
+    /// * `next_order` — Next expected message order, has to be 0 for a top level invocation
+    ///
+    /// # Returns
+    ///
+    /// Next expected message order
+    fn aggregate_messages_in_call_info(tx_hash: TransactionHash, call_info: &CallInfo, next_order: usize) -> usize {
```
tdelabro, line 993: I don't get why is this part of this PR

crates/pallets/starknet/src/transaction_validation.rs
```diff
@@ -50,3 +53,77 @@ impl<OuterOrigin: Into<Result<RawOrigin, OuterOrigin>> + From<RawOrigin>> Ensure
...
+            UserAndL1HandlerTransaction::L1Handler(transaction, fee) => transaction
+                .into_executable::<T::SystemHash>(chain_id, *fee, false)
+                .validate_tx(&mut state, &block_context, &mut execution_resources, &mut initial_gas, false),
+        }
+        .map_err(|_| InvalidTransaction::BadProof)?;
+
+        Ok(())
+    }
+
+    pub fn ensure_l1_message_not_executed(nonce: &Nonce) -> Result<(), InvalidTransaction> {
```
tdelabro, line 126: When this function is called, its error is either mapped to another error or converted to a bool. 

rename as `is_l1_message_already_executed` return a bool

crates/client/l1-messages/src/error.rs
```diff
@@ -0,0 +1,43 @@
...
+    L1MessageAlreadyProcessed,
+    #[error("Failed to read/write into Offchain Storage")]
+    OffchainStorageError,
+    #[error("Failed to use Runtime API")]
+    RuntimeApiError,
+    #[error("Failed to submit transaction into Transaction Pool")]
+    SubmitTxError,
+    #[error("Failed to convert L1 Message into Fee")]
+    ToFeeError,
+    #[error("Failed to convert L1 Message into L2 Transaction")]
```
tdelabro, line 41: `"Failed to convert L1 Message into L2 Transaction: {0}"`

tdelabro (changes requested):

tdelabro:
@bidzyyys also you will have to rebase or cherry-pick your commits

crates/pallets/starknet/src/lib.rs
bidzyyys: I have no idea what happened here while it is not my code...

crates/runtime/src/lib.rs
bidzyyys: This is also not my code...

crates/pallets/starknet/src/lib.rs
bidzyyys: This is also not my code...

crates/pallets/starknet/src/lib.rs
bidzyyys: Me too, it is not my code...

bidzyyys:
@tdelabro totally agree on rebase/cherry-pick while I can see diff not produced by me and IDK what happened here...

crates/pallets/starknet/src/lib.rs
tdelabro: This code is present on the l1-l2-messaging branch

bidzyyys:
Superseded by #1282

PR #1210, opened 2023-10-22
L2 -> L1 synchronization thread

m-kus:


crates/client/settlement/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct StarknetState {
+    /// The state commitment after last settled block.
+    pub state_root: StarkFelt,
+    /// The number (height) of last settled block.
+    pub block_number: StarkFelt,
+    pub config_hash: StarkFelt,
+    pub chain_id: StarkFelt,
```
tdelabro, line 45: Does every field have to be public?

crates/client/settlement/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl Default for StarknetState {
+    fn default() -> Self {
+        Self { state_root: 0u64.into(), block_number: 0u64.into(), config_hash: 0u64.into(), chain_id: 0u64.into() }
+    }
+}
```
tdelabro, line 52: I believe StarkFelt impl default. So you can probably derive Default

crates/client/settlement/src/mockup/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    async fn update_state(
+        &self,
+        program_output: StarknetOsOutput,
+        _da_blob_hash: StarkFelt,
+        _da_blob_size: StarkFelt,
+    ) -> Result<()> {
+        println!("{:#?}", program_output);
```
tdelabro, line 37: forgotten println

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use async_trait::async_trait;
+use futures_timer::Delay;
+use thiserror::Error;
+
+use crate::RetryStrategy;
+
+#[derive(Debug, Error)]
+pub enum UnrecoverableError {
+    #[error("State root mismatch for level {0}")]
```
tdelabro, line 11: what does `level` means in this context?

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,30 @@
...
+    async fn can_retry(&self, error: &anyhow::Error) -> bool {
+        match error.downcast_ref::<UnrecoverableError>() {
+            Some(_) => false,
+            None => {
+                Delay::new(self.delay).await;
+                true
+            }
+        }
+    }
+}
```
tdelabro, line 30: Where are the recoverable errors defined?

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// That means we will need to publish state diffs and STARK proof before state update.
+    pub async fn sync_state(
+        substrate_client: Arc<SC>,
+        settlement_provider: Box<dyn SettlementProvider>,
+        retry_strategy: Box<dyn RetryStrategy>,
+    ) {
+        loop {
+            match Self::sync_state_proc(substrate_client.as_ref(), settlement_provider.as_ref()).await {
+                Ok(()) => {
+                    return;
```
tdelabro, line 39: DO you really want to `return` here? Isn't it more a place where we want to `continue`?

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+                Err(err) => {
+                    log::error!("[settlement] {err}");
+                    match retry_strategy.can_retry(&err).await {
+                        true => {
+                            log::info!("[settlement] Retrying after error");
+                            continue;
+                        }
+                        false => {
+                            log::error!("[settlement] Unrecoverable error, exiting");
+                            return;
```
tdelabro, line 50: Maybe the signature should be `-> !` and this should be a `panic!`. Wdyt?

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    async fn sync_state_proc<SP>(substrate_client: &SC, settlement_provider: &SP) -> anyhow::Result<()>
```
tdelabro, line 58: what is the meaning of `_proc` here?

tdelabro, line 58: Can you use a more expressive error type here? An enum.

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+                log::debug!("[settlement] State transitioned to {:?}", new_state);
+                last_settled_state = new_state;
+                sync_from += 1;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn get_starknet_block(substrate_client: &SC, block_number: u64) -> anyhow::Result<(StarknetBlock, B::Hash)> {
```
tdelabro, line 115: Can we move this in some sort of client/utils crate?

tdelabro: There is probably code to share with the RPCs

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        let program_output = StarknetOsOutput {
+            prev_state_root: prev_state.state_root,
+            new_state_root: next_state.state_root,
+            block_number: next_state.block_number,
+            block_hash: next_block.header().hash::<H>().into(),
+            config_hash: next_state.config_hash,
+            messages_to_l1,
+            messages_to_l2,
+        };
```
tdelabro, line 202: We are short-circuiting the SNOS run here, isn't it?

Can you add a comment about it please somewhere

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+                // TODO: we skip this check for now, to simplify the setup (allow non-initialized state)
+                // We use zero state as a starting point (note that genesis state root is not zero actually)
+                return Ok(StarknetState::default());
+            }
+        };
+
+        Ok(last_settled_state)
+    }
+
+    async fn update_starknet_state<SP>(
```
tdelabro, line 162: doc

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+        let substrate_block_header = substrate_client
+            .header(substrate_block_hash)?
+            .ok_or_else(|| anyhow::anyhow!("Failed to get substrate block header for {}", substrate_block_hash))?;
+
+        let starknet_block = find_starknet_block(substrate_block_header.digest())
+            .map_err(|e| anyhow::anyhow!("Failed to find SN block: {}", e))?;
+        Ok((starknet_block, substrate_block_hash))
+    }
+
+    fn verify_starknet_state(substrate_client: &SC, state: &StarknetState) -> anyhow::Result<()> {
```
tdelabro, line 133: doc

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+
+        // Verify that current onchain state is consistent with corresponding Madara block
+        if state.state_root != block.header().global_state_root {
+            return Err(UnrecoverableError::StateRootMismatch(level).into());
+        }
+
+        Ok(())
+    }
+
+    async fn get_starknet_state<SP>(settlement_provider: &SP) -> anyhow::Result<StarknetState>
```
tdelabro, line 145: doc

crates/client/settlement/src/sync_state.rs
tdelabro, line 58: doc

starknet-rpc-test/src/lib.rs
```diff
@@ -141,28 +148,58 @@ fn get_free_port() -> Result<u16, TestError> {
...
-			"--",
-			"--alice",
-			"--sealing=manual",
-			&format!("--execution={execution}"),
-			"--chain=dev",
-			"--tmp",
-			&format!("--rpc-port={free_port}")
+            // Silence Madara stdout and stderr
+            .stdout(stdout)
+            .stderr(stderr)
```
tdelabro, line 190: Silence them back before merging

tdelabro (changes requested):

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// For now we use a simplified setup without validity proofs & DA, but in the future
+    /// Starknet state contract will also validate state transition against the fact registry.
+    /// That means we will need to publish state diffs and STARK proof before state update.
+    pub async fn sync_state(
+        substrate_client: Arc<SC>,
+        settlement_provider: Box<dyn SettlementProvider>,
+        retry_strategy: Box<dyn RetryStrategy>,
+    ) {
+        loop {
+            match Self::sync_state_proc(substrate_client.as_ref(), settlement_provider.as_ref()).await {
```
bidzyyys, line 37: Is type coercion for `Arc` not working here?

.gitignore
```diff
@@ -19,6 +19,7 @@
 **/node_modules/
 
 data
+logs
```
bidzyyys, line 22: Why do you need logs here? Is it possible to log somewhere else?

starknet-rpc-test/src/lib.rs
m-kus: I've added a flag to `MadaraClient::init` to enable/disable logging, I suppose this might be useful for other scenarios as well. I will set it to false everywhere, also can make it feature enabled if you'd like

crates/client/settlement/src/ethereum/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+    r#"[
+        function programHash() public view returns (uint256)
+        function stateBlockNumber() external view returns (int256)
+        function stateRoot() external view returns (uint256)
+        function configHash() public view returns (uint256)
+        function updateState(uint256[] calldata programOutput, uint256 onchainDataHash, uint256 onchainDataSize) external
+    ]"#,
+);
+
+pub fn convert_u256_to_felt(word: U256) -> Result<StarkFelt> {
```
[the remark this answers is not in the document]
m-kus: But they are both foreign?

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub struct RetryOnRecoverableErrors {
+    pub delay: Duration,
+}
+
+#[async_trait]
+impl RetryStrategy for RetryOnRecoverableErrors {
+    async fn can_retry(&self, error: &anyhow::Error) -> bool {
+        match error.downcast_ref::<UnrecoverableError>() {
```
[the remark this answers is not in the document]
m-kus: Yeah, I think I'll move to a dedicated error type instead of anyhow, it will make things simpler

crates/client/settlement/src/ethereum/mod.rs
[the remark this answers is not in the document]
m-kus: @tdelabro do you think we could add this cast to starknet-api?

crates/client/data-availability/src/ethereum/mod.rs
```diff
@@ -14,12 +14,23 @@ use crate::{DaClient, DaMode};
 
 #[derive(Clone, Debug)]
 pub struct EthereumClient {
-    http_provider: Provider<Http>,
-    signer: Arc<SignerMiddleware<Provider<Http>, LocalWallet>>,
-    cc_address: Address,
+    pub http_provider: Provider<Http>,
+    pub signer: Arc<SignerMiddleware<Provider<Http>, LocalWallet>>,
+    pub cc_address: Address,
```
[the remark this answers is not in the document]
m-kus: After some thoughts I created a separate client for Starknet state contract

starknet-rpc-test/src/lib.rs
m-kus: I made logs optional, now they can be enabled by providing MADARA_LOG env variable which should contain a relative log folder, is that ok? Generally I think there might be other cases when you'd want to check Madara logs for troubleshooting (at least I couldn't debug efficiently without them)

m-kus: Added readme for integration tests

crates/client/settlement/src/sync_state.rs
m-kus: Inside sync_state_proc we have another loop which listens for the block finalization events, it will exit when the finality_notifications_stream will be closed/terminated -- I assume upon the node termination?

crates/client/settlement/src/sync_state.rs
m-kus: Wow, thanks, I didn't know about this type!

crates/client/settlement/src/errors.rs
m-kus: The Starkness level where state root differs from what we got from the settlement contract

crates/client/settlement/src/errors.rs
m-kus: Now it's more explicitly defined

crates/client/settlement/src/sync_state.rs
m-kus: It's inner loop, renamed

crates/client/settlement/src/ethereum/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+        let contract = StarknetContract::new(self.cc_address, self.signer.clone());
+        let tx = contract
+            .update_state(program_output, onchain_data_hash, onchain_data_size)
+            .send()
+            .await
+            .map_err(|e| anyhow::anyhow!("Failed to update state: {}", e))?
+            .await
+            .map_err(|e| anyhow::anyhow!("Ethereum poll update error: {}", e))?;
+
+        log::debug!("[settlement] State update: {:#?}", tx);
```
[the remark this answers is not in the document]
m-kus: It will fail and outer loop will retry unless the error is not recoverable

crates/client/settlement/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Use Starknet state contract in Ethereum
+    Ethereum,
+}
+
+#[async_trait]
+pub trait SettlementProvider: Send + Sync {
+    async fn get_state(&self) -> Result<Option<StarknetState>>;
+    async fn update_state(
+        &self,
+        program_output: StarknetOsOutput,
```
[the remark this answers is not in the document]
m-kus: Typically I'd agree, but in this very case we get the SNOS output as a result of a previous conversion and we don't need it for any other purpose.

crates/client/settlement/src/mockup/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        let state = StarknetState {
```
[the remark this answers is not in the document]
m-kus: We could but there's no use for it, at least in the current scope

crates/client/settlement/src/sync_state.rs
m-kus: I tried to reuse as much as possible, also here we rely on crate-level error type (now)

crates/client/settlement/src/ethereum/mod.rs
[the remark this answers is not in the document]
m-kus: Let's refactor afterwards, I guess we will have more things to reuse, e.g. Starknet contract bindings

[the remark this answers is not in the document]
tdelabro: Yep. Keep it like this for now. In the future we will have the common felt type with hopefully conversion methods to u256

crates/node/src/service.rs
```diff
@@ -416,16 +416,34 @@ pub fn new_full(
...
+                    LocalStarknetContractClient::try_from(ethereum_conf)
+                        .map_err(|e| ServiceError::Other(e.to_string()))?,
+                )
+            }
+            #[cfg(feature = "mockup-settlement-provider")]
+            SettlementLayer::Mockup => {
+                Box::new(mc_settlement::mockup::MockupSettlementProvider { file_path: config_path })
             }
-            SettlementLayer::Mockup => Box::new(MockupSettlementProvider { file_path: config_path }),
+            _ => unreachable!("Unexpected settlement layer: {:?}", layer_kind),
```
tdelabro, line 428: This souldn't be required as you are doing  match? You get an ettor?

tdelabro (changes requested):

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,63 @@
...
+    StarknetApi(#[from] starknet_api::StarknetApiError),
+
+    #[error("Failed to find Madara log: {0}")]
+    DigestLog(#[from] mp_digest_log::FindLogError),
+
+    #[error("Failed to find Substrate block hash for Starknet level: {0}")]
+    MappingSync(u64),
+
+    #[error("Failed to find Substrate block header for hash: {0}")]
+    UnknownBlock(String),
```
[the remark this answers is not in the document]
m-kus: It's `<B as Block>::Hash`, but we would need to make Error generic then

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,63 @@
...
+    #[error("Blockchain error: {0}")]
+    Blockchain(#[from] sp_blockchain::Error),
+
+    #[error("Starknet API error: {0}")]
+    StarknetApi(#[from] starknet_api::StarknetApiError),
+
+    #[error("Failed to find Madara log: {0}")]
+    DigestLog(#[from] mp_digest_log::FindLogError),
+
+    #[error("Failed to find Substrate block hash for Starknet level: {0}")]
```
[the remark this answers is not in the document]
m-kus: Yes, chain height, I'll rename

crates/client/settlement/src/errors.rs
m-kus: Yes, I'll rename

crates/client/settlement/src/mockup/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+        &self,
+        program_output: StarknetOsOutput,
+        _da_blob_hash: StarkFelt,
+        _da_blob_size: StarkFelt,
+    ) -> Result<()> {
+        let state = StarknetState {
+            block_number: program_output.block_number,
+            state_root: program_output.new_state_root,
+            config_hash: program_output.config_hash,
+            chain_id: 0u64.into(),
```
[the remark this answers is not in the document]
m-kus: It's fixed in the latest revision, chain_id is retrieved from the runtime

crates/client/settlement/src/ethereum/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+        if program_hash == U256::zero() {
+            // Contract is not initialized yet
+            return Ok(None);
+        }
+
+        let block_number: I256 = contract
+            .state_block_number()
+            .call()
+            .await
+            .map_err(|e| anyhow::anyhow!("Failed to fetch state block number: {}", e))?;
```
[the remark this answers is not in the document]
m-kus: It's likely an outdated version, anyhow is no longer used

crates/node/src/service.rs
m-kus: Since mockup is not feature-enabled, I marked the enum as #[non_exhaustive] (afaik this is considered as good practice in such cases). Thus we need to handle _

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    #[error("Runtime API error: {0}")]
+    RuntimeApi(#[from] sp_api::ApiError),
+
+    #[error("{0}")]
+    // not using #[from] because of template vars and this way we can also have more context
+    EthereumClientError(String),
+
+    #[error("{0}")]
+    EthereumConfigError(String),
```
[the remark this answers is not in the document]
m-kus: It's a generic type, which means we either have to make our error generic as well, or lock to a specific ethers middleware (provider + wallet type)

starknet-rpc-test/settlement_worker.rs
```diff
@@ -0,0 +1,199 @@
+extern crate starknet_rpc_test;
```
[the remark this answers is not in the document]
m-kus: Sounds good!

crates/client/settlement/src/errors.rs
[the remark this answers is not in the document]
tdelabro: Sound ok for me to make it generic. As this error is clearly linked to the substrate framework

crates/client/settlement/src/errors.rs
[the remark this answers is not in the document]
tdelabro: Okay. But use `thiserror::Error` instead. And let the display impl do the conversion to string

crates/node/src/service.rs
tdelabro: https://doc.rust-lang.org/reference/attributes/type_system.html

It seems that it is used to guarantee backward compatibility downstream. I don't think this is something we care about.

crates/client/settlement/src/errors.rs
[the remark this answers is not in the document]
m-kus: Not for the particular settlement provider implementations, e.g. for Ethereum contract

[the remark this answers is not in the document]
tdelabro: The error message do mention Substrate: `    #[error("Failed to find Substrate block header for hash: {0}")]`
So I assume it will always be Substrate related

tdelabro (changes requested):

crates/client/settlement/src/errors.rs
```diff
@@ -0,0 +1,75 @@
...
+    UnknownStarknetBlock(u64),
+
+    #[error("Failed to find Substrate block header for hash: {0}")]
+    UnknownSubstrateBlock(String),
+
+    #[error("Unexpected global state root for block #{0}")]
+    StateRootMismatch(u64),
+
+    #[error("{0}")]
+    EthereumClientError(String),
```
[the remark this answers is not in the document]
m-kus: This seems outdated, pls check the latest commit

crates/client/settlement/src/ethereum/client.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn new(address: Address, client: Arc<M>) -> Self {
+        Self { contract: StarknetContract::new(address, client) }
+    }
+
+    pub async fn state_block_number(&self) -> Result<I256> {
+        self.contract
+            .state_block_number()
+            .call()
+            .await
+            .map_err(|e| Error::EthereumClientError(format!("Failed to fetch state block number: {}", e)))
```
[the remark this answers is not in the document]
m-kus: Also outdated

crates/client/settlement/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+homepage = "https://github.com/keep-starknet-strange/madara"
+edition = "2021"
+license = "MIT"
+publish = false
+repository = "https://github.com/keep-starknet-strange/madara"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
```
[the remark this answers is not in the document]
m-kus: For which crates?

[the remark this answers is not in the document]
tdelabro: As much as possible. All those from madara primitive. Probably also those from `sp`. And a bunch of other.
It's rare that a crate does not have default features

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,266 @@
...
+    /// For now we use a simplified setup without validity proofs & DA, but in the future
+    /// Starknet state contract will also validate state transition against the fact registry.
+    /// That means we will need to publish state diffs and STARK proof before state update.
+    ///
+    /// This is an external loop that is responsible for handling temporary (recoverable) errors.
+    pub async fn sync_state(
+        substrate_client: Arc<SC>,
+        settlement_provider: Box<dyn SettlementProvider<B>>,
+        retry_strategy: Box<dyn RetryStrategy<B>>,
+        starknet_spec: StarknetSpec,
```
tdelabro, line 38: Do you have to move the value? You are only using it by reference

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,266 @@
...
+    {
+        let mut last_settled_state = settlement_provider.get_state().await?.ok_or(Error::StateNotInitialized)?;
+        log::debug!("[settlement] Last settled state {:?}", last_settled_state);
+
+        Self::verify_starknet_spec(settlement_provider, starknet_spec).await?;
+        // If we haven't reached the settled level yet (e.g. syncing from scratch) this check will pass.
+        // But we need to run it again once we are up to speed.
+        Self::verify_starknet_state(substrate_client, &last_settled_state)?;
+
+        let mut finality_notifications = substrate_client.finality_notification_stream();
```
tdelabro, line 94: I don't like that we are using this in order to sync l2.
Different reason:
1) https://crates.parity.io/sc_client_api/client/trait.BlockchainEvents.html#tymethod.finality_notification_stream the doc state their is no guarantee an event is fired
2) If this thread goes down all the network node has to go done, otherwise we may miss some events

I would rather use a combination of https://crates.parity.io/sp_blockchain/trait.Backend.html#tymethod.last_finalized (in order to get the last finalized one) and https://crates.parity.io/sp_blockchain/trait.HeaderBackend.html#tymethod.header (to get every block from last_settled to last_finalized). This way this thread can be down, we can reboot the node and it will still correctly sync

tdelabro:
```rust
fn sync() -> !  {
let last_settled = get_last_settled();
let last_finalized = get_last_finalized();

  if last_settled.number == last_finalized.number {
     sleep.await;
     continue; 
  }
  if last_settled.number > last_finalized.number {
     panic!();
  }

  for block_number in last_settled.number..last_finalized.number {
     let block = get_block(block_number);
     <do our thing>
  }

}
```

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,266 @@
...
+        Ok((starknet_block, substrate_block_hash))
+    }
+
+    /// Checks that settlement contract is initialized with the same program & config hash as
+    /// Madara.
+    async fn verify_starknet_spec<SP>(settlement_provider: &SP, starknet_spec: &StarknetSpec) -> Result<(), B>
+    where
+        SP: ?Sized + SettlementProvider<B>,
+    {
+        let spec = settlement_provider.get_chain_spec().await?;
```
tdelabro, line 170: Is this value going to change once the settlement contract is deployed? If no, we can query it once at the beginning of the thread and use it forever after

tdelabro: Maybe it's an assumption we don't want to do?

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,266 @@
...
+        Ok(())
+    }
+
+    /// Tries to verify that the state root for the specified block in Madara storage
+    /// is the same as in the given state.
+    ///
+    /// If Madara chain haven't reached the given block level yet, it returns OK, assuming that as
+    /// soon as it catches up - this check will be done again.
+    fn verify_starknet_state(substrate_client: &SC, state: &StarknetState) -> Result<(), B> {
+        let level: u64 = state.block_number.try_into()?;
```
tdelabro, line 189: `height` rather than `level`

tdelabro (changes requested):
I think you should review the design of the worker

crates/client/settlement/src/ethereum/client.rs
```diff
@@ -0,0 +1,138 @@
...
+        _onchain_data_hash: U256,
+        _onchain_data_size: U256,
+    ) -> Result<TransactionReceipt> {
+        // TODO: change ABI and use DA part once we move on to the rollup mode (currently validium)
+        self.contract
+            .update_state(program_output)
+            .send()
+            .await
+            .map_err(|e| Error::EthereumClientError(format!("Failed to send transaction: {}", e)))?
+            .log_msg("[ethereum client] pending transaction")
```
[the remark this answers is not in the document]
m-kus: It can potentially be useful for troubleshooting in the future

crates/client/settlement/src/sync_state.rs
m-kus: Regarding the second item - this is handled in the current implementation, the thread can catch up with the head.
It also handles the situations where Madara state is behind the settled one (e.g. you are bootstrapping a new node) - in your snipped you panic, which is prob not what you'd want.

Overall, listening the finality notification stream is basically the same as I would poll last_finalized() method. Given that we cannot miss events, it's probably ok even if it's not guaranteed that event is not fired, as long as there will be another one next.

crates/client/settlement/src/ethereum/client.rs
```diff
@@ -0,0 +1,138 @@
...
+            .await
+            .map_err(|e| Error::EthereumClientError(format!("Ethereum poll update error: {}", e)))?
+            .ok_or_else(|| Error::EthereumClientError(format!("Failed to get transaction receipt")))
+    }
```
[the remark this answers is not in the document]
m-kus: I'm now using anyhow to bypass original error, as we discussed, is it ok?

crates/client/settlement/src/ethereum/client.rs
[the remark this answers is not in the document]
m-kus: Reduced to debug level

crates/client/settlement/src/sync_state.rs
tdelabro: Okay, I missed that part. So you fallback yo the solution I described if the chain is not up to date.
Seems okay

m-kus:
Superseded by #1279

Issue #1168, opened 2023-10-05
feat: Usability: Starknet Foundry with Madara 


lana-shanghai:
## Using starknet foundry with Madara 

Investigate how to add compatibility with Madara. Currently trying to create an Open Zeppelin account (default in Starknet Foundry) results in the following error:

```
sncast --url http://127.0.0.1:9944 account create --name my-account
command: account create
error: JSON-RPC error: code=61, message="The transaction version is not supported"
```

Account creation [code](https://github.com/foundry-rs/starknet-foundry/blob/33bd7a7c8247ad360fae1b47e69b1cd744199d10/crates/cast/src/starknet_commands/account/create.rs#L40)

The point where Madara fails with `UnsupportedTxVersion` is [here](https://github.com/keep-starknet-strange/madara/blob/6207f1c3a1376ac9e26d3a18a9685328422ec580/crates/client/rpc/src/lib.rs#L591) 

Open a PR in the Starknet Foundry [repo](https://github.com/foundry-rs/starknet-foundry) and test creating an account, fund it from the predeployed account, declare, deploy, invoke some test contracts. Reach out to the creators if necessary. Use the [starknet js scripts](https://github.com/lana-shanghai/madara_contract_scripts/tree/main) for inspiration.

justinfrevert:
This also impacts https://github.com/keep-starknet-strange/madara/issues/1149 which I had briefly looked into. I'm happy to look into this one

justinfrevert:
I did some digging: it seems the main cause for the errors is starknet-rs defaulting to `is_query: false` for `DeployAccount` actions(that happens [here)](https://github.com/xJonathanLEI/starknet-rs/blob/4ab2f36872435ce57b1d8f55856702a6a30f270a/starknet-accounts/src/factory/mod.rs#L346). I spoke to one of the devs in the starknet.rs channel, and [sounds like they're aware and planning to fix](https://t.me/starknet_rs/4880).

In the meantime, I can do a quick check if we can change the check on `is_query`, though it sounds like it might be good to just keep in place.

lana-shanghai:
@justinfrevert cool! Do you suggest we do a temporary check skip for now, until it is fixed in starknet rs?

justinfrevert:
We can do that for now, but it's critical that we undo this later because it affects transaction security(as I've heard). I'll do that and create an issue.

justinfrevert:
[...]

Issue #1149, opened 2023-09-28
test: add CI workflow for starknet-foundry compatiblity


tdelabro:
We want starknet-foundry to work 100% with madara, ideally just by specifying the correct rpc url

Add a workflow to the CI.
It should use a custom starknet-foundry project and run it against madara, making sure the tool is compatible

justinfrevert:
@tdelabro can I take this one?

tdelabro:
@justinfrevert do you think it should live in the madara repository? Or exist in another repo and just be used in the madara repo CI?

justinfrevert:
> @justinfrevert we don't want to overcharge the madara repo with the compatibility test. Do you think you could work on this on a repo of your own, that you will transfer to the @keep-starknet-strange organization when ready?
> 
> Then we will use it in Madara CI. Given the repo define a gh action
> 
> Does that sound ok to you?

Sure, it works for me. It seems like a fine approach. Will start on a separate repo focused on its own gh action.

justinfrevert:
@tdelabro how extensive do you think the testing of the initial github action should be?

tdelabro:
I think, `declare`, `deploy`, and `invoke` would be a good start. 
I think that is what 95% of foundry time is spent doing. But I may be missing something. What do you think?

tdelabro:
Hey, all good?

justinfrevert:
Hey @tdelabro I am looking into the incompatibilities themselves in https://github.com/keep-starknet-strange/madara/issues/1168 before I create the github action

tdelabro:
@justinfrevert sure! Thanks a lot!!

tdelabro:
@justinfrevert What is the state here?

justinfrevert:
Hi @tdelabro this was blocked because of https://github.com/keep-starknet-strange/madara/issues/1168. I just found the fix for that yesterday, so I will create the PR today and also begin working on this issue.

tdelabro:
@justinfrevert okay perfect. That's great to hear!

PR #1215, opened 2023-10-23
Alter transaction checks for starknet-rs/foundry compatibility

justinfrevert:
This PR fixes errors which occur when sending transactions through Starknet Foundry/starknet.rs.

Currently, transactions sent from starknet-rs fail, since starknet-rs sends `is_query: false` for their estimate calls.  

Resolves: #1168 

## What is the new behavior?

- Remove the `is_query` check in the estimate function. According to the chat in the starknet-rs, this will be changed later(see [the discussion](https://t.me/starknet_rs/4880)) so I added an error log indicating this. 
- Consequently, `is_query` can't be defaulted further along the code path since it is used in the construction of the hash used in signing/verifying. We accept the `is_query` value that was sent instead.

No, because the `is_query` value which was sent is used.

## Other information

This is also related to https://github.com/keep-starknet-strange/madara/issues/1149. That work will be done in a separate pr

### DIFF
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -608,7 +608,9 @@ where
             BroadcastedTransaction::DeployAccount(deploy_tx) => !deploy_tx.is_query,
         });
         if is_invalid_query_transaction {
-            log::error!("Got `is_query`: false. In a future version, this will fail fee estimation with UnsupportedTxVersion");
+            log::error!(
+                "Got `is_query`: false. In a future version, this will fail fee estimation with UnsupportedTxVersion"
+            );
         }
 
         let substrate_block_hash = self.substrate_block_hash_from_starknet_block(block_id).map_err(|e| {
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -608,7 +608,7 @@ where
             BroadcastedTransaction::DeployAccount(deploy_tx) => !deploy_tx.is_query,
         });
         if is_invalid_query_transaction {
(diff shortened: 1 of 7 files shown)


justinfrevert:
It seems like some of the actions(declare) still fail at the extrinsic-level. I was a bit confused because of the tool's output. I need to keep looking into it

tdelabro:
@justinfrevert is it ready for review?

crates/client/rpc/src/lib.rs
```diff
@@ -608,7 +608,9 @@ where
             BroadcastedTransaction::DeployAccount(deploy_tx) => !deploy_tx.is_query,
         });
         if is_invalid_query_transaction {
```
tdelabro, line 610: reverse the bool:
- remove all the `!` in the match
- rename `is_invalid_query_transaction` to `is_query`
- change `if is_invalid_query_transaction` to `if !is_query {}`
-  call `.estimate_fee(substrate_block_hash, tx, is_query)`

It's better than taking the inverse of the inverse

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
justinfrevert: That's nicer, thanks!

justinfrevert:
> @justinfrevert is it ready for review?

Yes, it was at the time. For some reason I wasn't able to put a label on it when I created it.

justinfrevert:
Also, fyi: I have a github action in the works to assure this functionality. I am wrapping that work up and will submit a separate PR soon.

tdelabro (approved):

Issue #1111, opened 2023-09-18
feat: return valid `actual_fee` in `get_transaction_receipt`


tdelabro:
Currently the `get_transaction_receipt` rpc return an incomplete object. The `actual_fee` field is set to `0`. 

Instead we should retrieve it from the events related to this TX and return it.

Also, uncomment the check for this in `starknet-rpc-test/get_transaction_receipt.rs` (signaled by a `TODO`)

antiyro:
Can I take this one?

tdelabro:
My DM are open if you have questions about the required impl

Issue #1095, opened 2023-09-12
dev: return deployed contract address in ` get_transaction_receipt` api


tdelabro:

a-moreira:
@tdelabro I'd like to take this one, although I imagine it needs #992 to be merged first

tdelabro:
Yes exactly. I will ping you when it is merged :)
Thanks a lot

tonypony220:
seems to be available now, isn't it?

tdelabro:
Yes indeed. You want to give it a try?

tonypony220:
@tdelabro Yes, I think to try. 
From first glance it seems that we could get transaction and calculate contract address, but maybe it's not right way.
Also this is related one [#1111](https://github.com/keep-starknet-strange/madara/issues/1111)

PR #868, opened 2023-07-13
No fee pallet option

apoorvsadana:
# Pull Request type

Need to merge #823 first

Please add the labels corresponding to the type of changes your PR introduces:

- Feature

## What is the current behavior?

Resolves: #847 

## What is the new behavior?

- A new config in pallet that makes charging of fees optional
- Removed hardcoded fees in transactions

## Does this introduce a breaking change?

No

PR #1216, opened 2023-10-23
dev: return deployed contract address in `get_transaction_receipt` api

tonypony220:
# Pull Request type

return zeros

Resolves: #1095
Resolves: #1111

## What is the new behavior?

- runtime api exposes new method to get fee disabled config  `is_transaction_fee_disabled`
- runtime api exposes new method to get transaction from block extrinsics `get_index_and_tx_for_tx_hash`
- runtime api refactor get_events just by id `fn get_events_for_tx_by_index(tx_index: u64) -> Option<Vec<StarknetEvent>>` 
---------
`get_transaction_receipt` changes: 
- get `actual_fee` from `Transfer` event if fee not `fee_disabled` 
- in case `DeployAccount` transaction calculates `contract_address` , cause searching it in events is errorprone

No


crates/client/rpc/src/lib.rs
```diff
@@ -943,48 +944,88 @@ where
...
+            let transfer_event_key = starknet_core::utils::get_selector_from_name("Transfer").unwrap();
+            let find_transfer_fee_event = events.iter().find_map(|event| {
+                event.keys.iter().find(|&&key| key == transfer_event_key)?;
+                event.data.get(1).and_then(|send_to| (*send_to == sequencer_address.0).then_some(()))?;
+                Some(event)
+            });
+
+            match find_transfer_fee_event {
+                Some(transfer_fee_event) => {
+                    let actual_fee = Felt252Wrapper(transfer_fee_event.data[2].clone()).0; // ony low bits, see execute_fee_transfer() why
```
tdelabro, line 960: Can you explain why here?

crates/client/rpc/src/lib.rs
```diff
@@ -943,48 +944,88 @@ where
...
+                        let address_in_other_event = events_converted.iter().find_map(|event| {
+                            event.keys.iter().find(|&&key| key != transfer_event_key)?;
+                            Some(Felt252Wrapper(event.from_address.clone()).0)
+                        });
+                        match address_in_other_event {
+                            Some(contract_address) => contract_address,
+                            None => Default::default(), // should we calculate it than?
+                        }
+                    }
+                };
```
tdelabro, line 998: Why is this one treated differently?

crates/client/rpc/src/lib.rs
```diff
@@ -943,48 +944,88 @@ where
...
+                    (actual_fee, Some(contract_address))
+                }
+                None => (Default::default(), None), // transaction without fee
+            }
+        }
+
+        let events_converted: Vec<starknet_core::types::Event> =
+            events.clone().into_iter().map(event_conversion).collect();
+
+        let (actual_fee, contract_address) = extract_fee_and_contract_address(&events_converted, &sequencer_address);
```
tdelabro, line 971: With your current impl of `extract_fee_and_contract_address` I an account does a transfer to the sequencer address. You will register this tx as the fee transfer, when there is another fee transfer after.

tdelabro: Maybe you should iterate in reverse. Or always assume transfer will be the last event of the tx and directly index it. Rather than searching for it

tdelabro: Write a test for this scenario

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
```diff
@@ -943,48 +944,88 @@ where
             }
         }
 
+        fn extract_fee_and_contract_address(
+            events: &Vec<starknet_core::types::Event>,
+            sequencer_address: &Felt252Wrapper,
+        ) -> (FieldElement, Option<FieldElement>) {
+            let transfer_event_key = starknet_core::utils::get_selector_from_name("Transfer").unwrap();
```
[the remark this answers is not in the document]
tonypony220: done

crates/client/rpc/src/lib.rs
```diff
@@ -943,48 +944,88 @@ where
...
+            let find_transfer_fee_event = events.iter().find_map(|event| {
+                event.keys.iter().find(|&&key| key == transfer_event_key)?;
```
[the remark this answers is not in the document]
tonypony220: Yes, good point.

crates/client/rpc/src/lib.rs
tonypony220: you mean why only low bits? I copied comment from `execute_fee_transfer()` here

crates/client/rpc/src/lib.rs
tonypony220: In case we have special transaction with disabled fee, i guess there would not be any transfer event, but we still need contract_address, so it's second search for that.

crates/client/rpc/src/lib.rs
tonypony220: Thank you for this idea! 
About test. Seems that [this test](https://github.com/keep-starknet-strange/madara/blob/10f6179101926067de4fb114b4510d1737f8f7bb/crates/pallets/starknet/src/tests/events.rs#L57) is about correct order of events already.

crates/client/rpc/src/lib.rs
tonypony220: Okay, I found PR #868, and it seems that the `disable-transaction-fee` feature is not quite ready to be used because most of the tests do not assume that it can be configured that way. 
So I deleted this additional search.

tonypony220:
@EvolveArt can you maybe take a look?

EvolveArt (approved):
Overall lgtm.

We should create an issue to deal with the `deploy_account` edge case mentionned in the PR's comments.
Maybe in the future we can also find a more efficient way than looking at the events to retrieve this information.

crates/client/rpc/src/lib.rs
tdelabro: I meant to write tests about transferring ETH to the sequencer address and make sure that it doesn't mess up with get_transaction_receipt

crates/client/rpc/src/lib.rs
```diff
@@ -1003,20 +1018,19 @@ where
...
-                        match address_in_other_event {
-                            Some(contract_address) => contract_address,
-                            None => Default::default(), // should we calculate it than?
-                        }
+                // In case of no transfer event we should search again for contract address
+                let contract_address = contract_address.unwrap_or_else(|| {
+                    let transfer_event_key =
+                        starknet_core::utils::get_selector_from_name(mp_fee::TRANSFER_SELECTOR_NAME).unwrap();
+                    let address_in_other_event = events_converted.iter().find_map(|event| {
+                        event.keys.iter().find(|&&key| key != transfer_event_key)?;
```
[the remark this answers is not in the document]
tdelabro: And I have the feeling the account paying for the deployment can be another account than the one being deployed. Depends on the specific of the contract implementation.

So I would say we don't try to get the account from the transfer event at the same time we get the fees. We just get it from there, from the ContractDeployed account

tdelabro (changes requested):

tdelabro:
@tonypony220 are you still making progresses on this one?

tonypony220:
> @tonypony220 are you still making progresses on this one?

Yeap!

crates/client/rpc/src/lib.rs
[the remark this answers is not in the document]
tonypony220: Yes, that's a really good point about the fee transfer mistake! As for 'ContractDeployed,' I want to believe that this event should be there, but when I look at the tests, it doesn't seem like it. Nevertheless, I have added it.

crates/client/rpc/src/lib.rs
```diff
@@ -965,31 +970,41 @@ where
...
+            //     keys: [selector("Transfer")],
+            //     data: [
+            //         send_from_address,       // account_contract_address
+            //         send_to_address,         // to (sequencer address)
+            //         expected_fee_value_low,
+            //         expected_fee_value_high,
+            //     ],
+            // },
+            let contract_address = Felt252Wrapper(*transfer_fee_event.data.get(0)?).0;
+            // TODO: This is what's done in the blockifier but this should be improved.
```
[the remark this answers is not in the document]
tonypony220: It was just reference to the same comment, I deleted this

crates/client/rpc/src/lib.rs
```diff
@@ -965,31 +970,41 @@ where
...
+            let lsb_amount = Felt252Wrapper(*transfer_fee_event.data.get(2)?).0;
+            Some((lsb_amount, Some(contract_address)))
         }
 
         let events_converted: Vec<starknet_core::types::Event> =
             events.clone().into_iter().map(event_conversion).collect();
 
-        let (actual_fee, contract_address) = extract_fee_and_contract_address(&events_converted, &sequencer_address);
+        let (actual_fee, contract_address) = if fee_disabled {
+            (FieldElement::ZERO, None)
```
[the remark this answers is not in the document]
tonypony220: did you mean fee disabled? This case handled below in case of `DeployAccount` recipe only

crates/client/rpc/src/lib.rs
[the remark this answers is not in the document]
tdelabro: Okay, I would be happy if we just compute it!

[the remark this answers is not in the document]
tonypony220: It's ready for review

crates/primitives/fee/src/lib.rs
```diff
@@ -48,6 +48,12 @@ pub static VM_RESOURCE_FEE_COSTS: [(&str, FixedU128); 7] = [
     ("ec_op_builtin", FixedU128::from_inner(10_240_000_000_000_000_000)),
 ];
 
+pub const TRANSFER_SELECTOR_NAME: &str = "Transfer";
```
tdelabro, line 51: Is this used anywhere in the code?

tdelabro (changes requested):

crates/primitives/fee/src/lib.rs
tonypony220: oh, yes. Forgot about this. Added to tests so that hash from it would be more readable

tdelabro (approved):

crates/runtime/src/lib.rs
```diff
@@ -348,7 +354,7 @@ impl_runtime_apis! {
                 tx_index as u32 == index
```
tonypony220, line 354: A little bit not sure about this covertion from `u64` -> `u32`, but it was `usize` -> `u32`

tdelabro: ```Some((tx_index as u64, transaction))``` it's u64 because of this cast from usize.
I think you can cast index as u32 from the get go. There will never be more than u32::MAX transaction in a block.
So having index being a u32 everywhere sound ok to me. 

You can also use `try_from().expect()` if you don't want the error to be silenced and the number to saturate

tdelabro (approved):

PR #1217, opened 2023-10-23
Build CommitmentStateDiff based on new block storage changes

tdelabro:


crates/client/commitment-state-diff/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+
+    let mut commitment_state_diff = CommitmentStateDiff {
+        address_to_class_hash: Default::default(),
+        address_to_nonce: Default::default(),
+        storage_updates: Default::default(),
+        class_hash_to_compiled_class_hash: Default::default(),
+    };
+
+    for (_prefix, full_storage_key, change) in storage_notification.changes.iter() {
+        // The storages we are interested in here all have longe keys
```
EvolveArt, line 149:
```suggestion
        // The storages we are interested in here all have long keys
```

Could you expand on this ?

crates/client/commitment-state-diff/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+                    contract_storage.insert(storage_key, value);
+
+                    commitment_state_diff.storage_updates.insert(contract_address, contract_storage);
+                }
+            }
+        } else if prefix == *SN_CONTRACT_CLASS_HASH_PREFIX {
+            let contract_address =
+                ContractAddress(PatriciaKey(StarkFelt(full_storage_key.0[32..].try_into().unwrap())));
+            // `change` is safe to unwrap as `ContractClassHashes` storage is `ValueQuery`
+            let class_hash = ClassHash(StarkFelt(change.unwrap().0.clone().try_into().unwrap()));
```
EvolveArt, line 185: we could declare `change.unwrap().0.clone().try_into().unwrap()` outside of the if/else branches

crates/client/commitment-state-diff/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+            commitment_state_diff.class_hash_to_compiled_class_hash.insert(class_hash, compiled_class_hash);
+        }
+    }
+
+    Ok((starknet_block_hash, commitment_state_diff))
+}
+
+pub async fn log_commitment_state_diff(mut rx: mpsc::Receiver<(BlockHash, CommitmentStateDiff)>) {
+    while let Some((block_hash, csd)) = rx.next().await {
+        log::info!("recieved state diff for block {block_hash}: {csd:?}");
```
EvolveArt, line 207:
```suggestion
        log::info!("received state diff for block {block_hash}: {csd:?}");
```

crates/client/commitment-state-diff/src/lib.rs
tdelabro: done

crates/client/commitment-state-diff/src/lib.rs
tdelabro: `change.unwrap()`
change is not always safe to unwrap.

It would be `None` if the storage was an `OptionValue` that has been set back to none.

Here I wrote:
`// `change` is safe to unwrap as `ContractClassHashes` storage is `ValueQuery``

later you will find:
```rust
            let compiled_class_hash = CompiledClassHash(match change {
                Some(data) => StarkFelt(data.0.clone().try_into().unwrap()),
                // This should not happen in the current state of starknet protocol, but there have been
                // an erase_contract_class mechanism live on the network during the regenesis migration.
                // Better safe than sorry
                None => StarkFelt::default(),
            });
```
where I match change rather than unwrapping it

EvolveArt (approved):

drspacemn:
lgtm

PR #1232, opened 2023-10-29
NEAR Protocol DA Support

encody:
Adds support for [NEAR Protocol](https://near.org/) as a data availability layer.

# Pull Request type

- Feature

No support for NEAR Protocol as a data availability layer.

## What is the new behavior?

- NEAR Protocol DA configuration file
- NEAR Protocol DA client

Yes

- Because `mc_data_availability::DaLayer` is not annotated with `#[non_exhaustive]`, the introduction of the `Near` variant will cause downstream crates that do not match against a wildcard to fail to compile.
- The `mc_data_availability::DaClient` trait has been altered to work with `bytes::Bytes` types instead of EVM-specific types. Conversion methods to- and from- `ethers::types::U256` have been provided: `mc_data_availability::utils::u256_to_bytes` and `mc_data_availability::utils::try_bytes_to_vec_u256`.
- Some dependency version requirements have been loosened (e.g. `tokio`, `ethers`) to major versions.

## Other information

This branch currently depends on a crate (`near-da-http-api-data`) available in [this repo](https://github.com/near/rollup-data-availability), but which is not published on crates.io, as well as custom forks of `nearcore` and `near-jsonrpc-client-rs`, which are necessary due to dependency versioning issues with starknet dependencies.


firatNEAR:
Could somebody from Madara side take a look at this PR? It is ready to be reviewed from our side

Cargo.toml
```diff
@@ -237,8 +238,19 @@ num-bigint = "0.4.4"
...
+[patch.crates-io]
+# custom forks that don't break starknet dependencies
+near-o11y = { git = "https://github.com/encody/nearcore", branch = "madara-compat" }
+near-chain-configs = { git = "https://github.com/encody/nearcore", branch = "madara-compat" }
+near-crypto = { git = "https://github.com/encody/nearcore", branch = "madara-compat" }
+near-primitives = { git = "https://github.com/encody/nearcore", branch = "madara-compat" }
+near-primitives-core = { git = "https://github.com/encody/nearcore", branch = "madara-compat" }
+near-jsonrpc-primitives = { git = "https://github.com/encody/nearcore", branch = "madara-compat" }
+
+near-jsonrpc-client = { git = "https://github.com/encody/near-jsonrpc-client-rs", branch = "madara-compat" }
```
apoorvsadana, line 256: let's rebase this and add all near libraries behind a feature flag like we do right now for Avail and Celestia

crates/client/data-availability/Cargo.toml
```diff
@@ -42,7 +43,7 @@ mc-db = { workspace = true, default-features = true }
 starknet_api = { workspace = true, default-features = true }
 
 # Ethereum
-ethers = "2.0.7"
+ethers = "2"
```
apoorvsadana, line 46: We're using a specific commit of ethers now in the workspace Cargo.toml. Should be fixed with a rebase.

crates/client/data-availability/Cargo.toml
```diff
@@ -55,3 +56,7 @@ celestia-types = { git = "https://github.com/eigerco/celestia-node-rs", rev = "b
 
 # Madara
 mp-storage = { workspace = true, default-features = true }
+
+# NEAR Protocol
+near-da-rpc = { git = "https://github.com/near/rollup-data-availability", branch = "madara-compat" }
+near-da-primitives = { git = "https://github.com/near/rollup-data-availability", branch = "madara-compat" }
```
apoorvsadana, line 62: Let's move this behind the feature flag as well.

crates/client/data-availability/src/near/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    async fn publish_state_diff(&self, state_diff: bytes::Bytes) -> Result<()> {
+        // setter
+
+        let res = self
+            .client
+            .submit(&[Blob::new_v0(self.client.config.namespace, state_diff.to_vec())])
+            .await
+            .map_err(|e| anyhow::anyhow!("failed to submit blobs: {e}"))?;
```
apoorvsadana, line 53: do we need to wait to check of the txn was finally inlcuded in the block?

apoorvsadana (commented):
Hey! Thanks a lot for the help and the effort. Can you please rebase the PR because we have made some changes since the last commit. Also, it would be nice if you could add some basic docs in `docs/data-availability/near.md` explaining how to set up Madara for NEAR DA. We can follow a similar process for other DAs as well then.

crates/client/data-availability/src/near/mod.rs
encody: no, `submit` submits and waits for finality

crates/node/src/service.rs
```diff
@@ -12,10 +12,6 @@ use futures::prelude::*;
 use madara_runtime::opaque::Block;
 use madara_runtime::{self, Hash, RuntimeApi, SealingMode, StarknetHasher};
 use mc_commitment_state_diff::CommitmentStateDiffWorker;
-#[cfg(feature = "avail")]
-use mc_data_availability::avail::{config::AvailConfig, AvailClient};
-#[cfg(feature = "celestia")]
-use mc_data_availability::celestia::{config::CelestiaConfig, CelestiaClient};
```
apoorvsadana, line 18: why are we removing this?

crates/client/data-availability/src/near/mod.rs
apoorvsadana: if possible, can you add a comment for this?

tdelabro:
not relevant after #1580

Issue #1202, opened 2023-10-19
bug: estimate_fee should keep state changes between two tx when doing bulk estimate


tdelabro:
# Bug Report

**Madara version:** 0.4.0

**Current behavior:**

The RPC route estimate_fee receives a vec of tx to estimate.
We spawn one instance of the runtime for each tx. 
This means that the state is reset between each estimated transaction.

**Expected behavior:**

The state should not reset. Tx number n should be executed on a state changed by tx number n-1

**Steps to reproduce:**

Call estimate fee with multiple tx from the same account, while increasing the nonce for each tx.
It will fail because the second will have nonce = 1 when the runtime expect nonce = 0.

**Related code:**

the loop in crates/client/rpc/src/lib.rs, fn estimate_fee: `        for tx in request {...}` should happen inside  crates/pallets/starknet/src/lib.rs, fn estimate_fee.

it's signature should be changed to:
```rust
    pub fn estimate_fee(transaction: Vec<UserTransaction>) -> Result<Vec<(u64, u64)>, DispatchError>;
```

**Other information:**

PR #1240, opened 2023-10-31
estimate fee for bulk transactions

lana-shanghai:
# Pull Request type

- Feature
Allow to pass a vector of transactions for fee estimation. Needed for instance for compatibility with the Argent wallet. 

Currently we accept only 1 transaction to estimate the fee and the state rolls back to initial state after each estimation. 
Resolves: [issue](https://github.com/keep-starknet-strange/madara/issues/1202)

## What is the new behavior?

We pass a vector of transaction to the `estimate_fee` function

### DIFF
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -641,8 +641,11 @@ where
                 StarknetRpcApiError::ContractError
             })?;
 
-        let estimates = fee_estimates.into_iter().map(|x| FeeEstimate { gas_price: 0, gas_consumed: x.0, overall_fee: x.1 }).collect();
-        
+        let estimates = fee_estimates
+            .into_iter()
+            .map(|x| FeeEstimate { gas_price: 0, gas_consumed: x.1, overall_fee: x.0 })
+            .collect();
+
         Ok(estimates)
     }
 
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -619,27 +619,30 @@ where
         let best_block_hash = self.client.info().best_hash;
         let chain_id = Felt252Wrapper(self.chain_id()?.0);
 
-        let mut estimates = vec![];
+        let mut transactions = vec![];
         for tx in request {
             let tx = tx.try_into().map_err(|e| {
                 error!("Failed to convert BroadcastedTransaction to UserTransaction: {e}");
                 StarknetRpcApiError::InternalServerError
             })?;
-            let (actual_fee, gas_usage) = self
-                .client
-                .runtime_api()
-                .estimate_fee(substrate_block_hash, tx, is_query)
-                .map_err(|e| {
-                    error!("Request parameters error: {e}");
-                    StarknetRpcApiError::InternalServerError
-                })?
-                .map_err(|e| {
-                    error!("Failed to call function: {:#?}", e);
-                    StarknetRpcApiError::ContractError
-                })?;
-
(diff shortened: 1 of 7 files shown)


EvolveArt (approved):

crates/client/rpc/src/lib.rs
```diff
@@ -619,27 +619,33 @@ where
...
-                    error!("Request parameters error: {e}");
-                    StarknetRpcApiError::InternalServerError
-                })?
-                .map_err(|e| {
-                    error!("Failed to call function: {:#?}", e);
-                    StarknetRpcApiError::ContractError
-                })?;
-
-            estimates.push(FeeEstimate { gas_price: 0, gas_consumed: gas_usage, overall_fee: actual_fee });
+            transactions.push(tx);
```
tdelabro, line 628: You can use `into_iter` and `map`

crates/pallets/starknet/src/lib.rs
```diff
@@ -1092,64 +1090,82 @@ impl<T: Config> Pallet<T> {
...
             disable_nonce_validation: bool,
-        ) -> TransactionExecutionResult<TransactionExecutionInfo> {
-            // TODO: initialization can probably be skiped by using mem::MaybeUninit
-            let mut execution_result = Ok(Default::default());
+            chain_id: Felt252Wrapper,
+        ) -> Vec<TransactionExecutionResult<TransactionExecutionInfo>> {
+            let mut execution_results = vec![];
             let _: Result<_, DispatchError> = storage::transactional::with_transaction(|| {
-                execution_result = tx.execute(state, block_context, true, disable_nonce_validation);
+                for tx in txs {
```
tdelabro, line 1104: You should use an iterator and map, and stop iterating at the first error encountered.
`collect::<Result<Value, E>>` allows you to do this.

Here you keep executing every transaction, even after a previous one failed. This is a waste of computation

tdelabro: And you just do `exectuction_results = txs.into_iter().map(|tx| {...})` to bring the result outside the closure

crates/pallets/starknet/src/lib.rs
```diff
@@ -1092,64 +1090,82 @@ impl<T: Config> Pallet<T> {
...
+        for res in execution_results {
+            match res {
+                Ok(tx_exec_info) => {
+                    log!(info, "Successfully estimated fee: {:?}", tx_exec_info);
+                    if let Some(l1_gas_usage) = tx_exec_info.actual_resources.0.get("l1_gas_usage") {
+                        results.push((tx_exec_info.actual_fee.0 as u64, *l1_gas_usage as u64));
+                    } else {
+                        return Err(Error::<T>::TransactionExecutionFailed.into());
+                    }
+                }
```
tdelabro, line 1161: This is not needed anymore if you stop executing the tx after the first that failed

crates/pallets/starknet/src/runtime_api.rs
```diff
@@ -46,7 +46,7 @@ sp_api::decl_runtime_apis! {
         /// Returns the chain id.
         fn chain_id() -> Felt252Wrapper;
         /// Returns fee estimate
```
tdelabro, line 48: Add in the doc, which `u64` is the `overall_fee` and which is the `gas_consumed`

tdelabro: But even better would be creating a dedicated struct so the API users are less prone to commit mistakes

Issue #1243, opened 2023-11-04
feat: add argent multicall contract to genesis


EvolveArt:
# Feature Request

**Describe the Feature Request**

Argent is relying on this [multicall](https://github.com/argentlabs/argent-contracts-starknet/blob/main/contracts/multicall/src/multicall.cairo) contract for calling contracts. We should add it to genesis along with the other contracts (ERC20, ERC721...)

jobez:
i can work on this

PR #1247, opened 2023-11-05
feat/add argent multicall contract to genesis

jobez:
# Pull Request type

- Bugfix
- Feature ✔️ 
- Build-related changes
- Testing
- Other (please describe):

Currently, madara is not compatible with Argent wallet usage. 

Resolves: #1243 

## What is the new behavior?

We add the argent [multicall ](https://github.com/argentlabs/argent-contracts-starknet/blob/main/contracts/multicall/src/multicall.cairo) contract to genesis to enable compatibility. 

Where the artifact `ArgentMulticall.casm.json` was generated via the following steps in the `argent-contracts-starknet` repo

- following their setup, which by convention sets the compiler version to 2.0.0 https://github.com/argentlabs/argent-contracts-starknet/blob/main/Makefile#L6
Compile to sierra
- starknet-compile ./contracts/multicall/ ArgentMulticall.sierra.json --allowed-libfuncs-list-name all --contract-path multicall::multicall::Multicall 
Further, into casm
- starknet-sierra-compile ArgentMulticall.sierra.json ArgentMulticall.casm.json
and I determined class hash via
- starkli class-hash ArgentMulticall.casm.json 

## Note
I synced with @lana-shanghai about some discrepancies I was seeing in the class-hash starkli was deriving for some of the assets in `genesis-assets`

I wrote this quick script to compare results with what was in genesis: https://gist.github.com/jobez/4213fc2ece3bd0c33e1c670e41a0e2ca#file-class_hash_with_starkli-sh

Using starkli version: 0.1.19 (3fd85cf), I was seeing differences in class hash in a couple cases, output can be seen here: https://gist.github.com/jobez/4213fc2ece3bd0c33e1c670e41a0e2ca#file-starkli_diff_result-txt

If someone can reproduce these differences, I'd be happy to do a follow up PR and update the class hashes in the genesis file.


EvolveArt:
Thanks for this PR! 

Regarding the class hashes discrepancy issue, thanks for pointing it out will investigate.

Would be great to have a follow up PR to fix these and maybe add a check in the CI so we make sure it's always correct.

EvolveArt (approved):

EvolveArt:
> Thanks for this PR! 
> 
> Regarding the class hashes discrepancy issue, thanks for pointing it out will investigate.
> 
> Would be great to have a follow up PR to fix these and maybe add a check in the CI so we make sure it's always correct.

Also if we can deploy this contract at genesis at the same address that Argent uses would be perfect @jobez

EvolveArt (changes requested):
.

lana-shanghai:
> > Thanks for this PR!
> > Regarding the class hashes discrepancy issue, thanks for pointing it out will investigate.
> > Would be great to have a follow up PR to fix these and maybe add a check in the CI so we make sure it's always correct.
> 
> Also if we can deploy this contract at genesis at the same address that Argent uses would be perfect @jobez

Sorry, totally forgot about this. Yes, it has to be at the same address for the wallet to be able to find it. If the hash is different, we will have the same problem as when we didn't this contract deployed. :(

jobez:
> > Thanks for this PR!
> > Regarding the class hashes discrepancy issue, thanks for pointing it out will investigate.
> > Would be great to have a follow up PR to fix these and maybe add a check in the CI so we make sure it's always correct.
> 
> Also if we can deploy this contract at genesis at the same address that Argent uses would be perfect @jobez

predeployed to address as sourced from:

https://github.com/argentlabs/argent-x/blob/0cb22f7bb5b0045d8f0aebb22ee245bd59f5db8c/packages/multicall/src/main.ts#L6

EvolveArt (approved):

PR #1250, opened 2023-11-06
update avail dependencies

vibhurajeev:
# Pull Request type

The current avail adapter uses avail-subxt version "0.3.0", and the default avail ws url was "wss://kate.avail.tools:443/ws".

Resolves: #NA

## What is the new behavior?
- The default ws url is now "ws://127.0.0.1:9945"
- Avail subxt version is "0.4.0"
- By default validate codegen for avail is set to false.

No


EvolveArt:
thanks for this PR @vibhurajeev if you can just update the changelog pls :)

EvolveArt (approved):

PR #1257, opened 2023-11-08
Improve stability with avail DA

vibhurajeev:
# Pull Request type

The current behaviour does not have ws restart implemented and da worker waits for transaction finalisation.

Resolves: #NA

## What is the new behavior?
- The new behaviour creates a new ws connection on connection closed.
- The da worker now fires the da transaction and does not wait for finalisation.

No


crates/client/data-availability/src/avail/mod.rs
```diff
@@ -19,22 +21,59 @@ type AvailPairSigner = subxt::tx::PairSigner<AvailConfig, Pair>;
...
     async fn publish_state_diff(&self, state_diff: Vec<U256>) -> Result<()> {
         let bytes = get_bytes_from_state_diff(&state_diff);
         let bytes = BoundedVec(bytes);
-
-        let submitted_block_hash = self.publish_data(&bytes).await?;
+        self.publish_data(&bytes).await?;
 
         // This theoritically do not have to be put here since we wait for finalization before
-        self.verify_bytes_inclusion(submitted_block_hash, &bytes).await?;
+        // self.verify_bytes_inclusion(submitted_block_hash, &bytes).await?;
```
EvolveArt, line 76: remove comments

crates/client/data-availability/src/avail/mod.rs
```diff
@@ -19,22 +21,59 @@ type AvailPairSigner = subxt::tx::PairSigner<AvailConfig, Pair>;
...
+
+        Ok(Self { client: ws_client, config: conf })
+    }
+}
+
 #[async_trait]
 impl DaClient for AvailClient {
     async fn publish_state_diff(&self, state_diff: Vec<U256>) -> Result<()> {
         let bytes = get_bytes_from_state_diff(&state_diff);
         let bytes = BoundedVec(bytes);
```
EvolveArt, line 72: let's put this on one line

crates/client/data-availability/src/avail/mod.rs
```diff
@@ -50,41 +89,47 @@ impl DaClient for AvailClient {
 }
 
 impl AvailClient {
-    async fn publish_data(&self, bytes: &BoundedVec<u8>) -> Result<H256> {
+    async fn publish_data(&self, bytes: &BoundedVec<u8>) -> Result<()> {
+        let mut ws_client = self.ws_client.lock().await;
```
EvolveArt, line 93: why do we need a lock ? afaik there is no concurrency possible

crates/client/data-availability/src/avail/mod.rs
```diff
@@ -50,41 +89,47 @@ impl DaClient for AvailClient {
...
-            .find(|call| match call {
-                Call::DataAvailability(DaCall::submit_data { data }) => data == bytes,
-                _ => false,
-            })
-            .ok_or(anyhow::anyhow!("Bytes not found in specified block"))?;
+        match ws_client.client().tx().sign_and_submit(&data_transfer, &self.signer, extrinsic_params).await {
+            Ok(i) => i,
+            Err(e) => {
+                if e.to_string().contains("restart required") {
+                    ws_client.restart().await;
```
EvolveArt, line 102:
```suggestion
                    ws_client.restart().await?;
```

crates/client/data-availability/src/avail/mod.rs
```diff
@@ -50,41 +89,47 @@ impl DaClient for AvailClient {
...
+                }
+
+                return Err(anyhow!("Da failed due to closed websocket connection: {e}"));
```
EvolveArt, line 105:
```suggestion
                return Err(anyhow!("DA Layer error : failed due to closed websocket connection {e}"));
```

EvolveArt (approved):
smol changes otherwise lgtm

crates/client/data-availability/src/avail/mod.rs
vibhurajeev: I left it this way, so the original error is propagated, and not the restart failing error. What do you think?

EvolveArt: Let's add a comment then until we move away from anyhow

crates/client/data-availability/src/avail/mod.rs
vibhurajeev: There is no concurrency possible, but we need to update the client with the new instance during which no one should access it. If its inside a method with non mutable reference, I wanted to keep it behind mutex guard so it is updated safely. Let me know if you think it should be implemented differently.

crates/client/data-availability/src/avail/mod.rs
vibhurajeev: I kept it this way, so the original error is propagated back to the worker.

crates/client/data-availability/src/avail/mod.rs
EvolveArt: Fair enough, thanks for the explanation

PR #1264, opened 2023-11-15
Polkadot-sdk v1.3.0 Dependency Update

notlesh:
This PR updates Madara's Substrate dependency tree to use the recent `polkadot-release-v1.3.0` branch. See: https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.3.0 for more details.

The previous version used was `v0.9.43`, so this effectively skips several releases. This is generally fine, but it would be good to ensure that this doesn't cause Madara to miss any pallet migrations. Since Madara is not in production, this isn't critical.

I don't believe this update should have any significant behavior changes, it mostly reflects upstream refactors.

Companion PR for `blockifier`: https://github.com/keep-starknet-strange/blockifier/pull/17

# Pull Request type

> Please add the labels corresponding to the type of changes your PR introduces:

I'm not able to edit labels, please assist :)

No (none identified so far, anyway). There may be new CLI options, RPC endpoints, or other features inherited from upstream; this will be reviewed as one of my TODO items below.

### TODO: 

 - [x] merge `master`
 - [x] CI green
 - [x] Review upstream pallet migrations
 - [x] Companion PR (I think just `blockifier` needs a small `Cargo.toml` change)


tdelabro (approved):
the crates client/block-proposer and client/transaction-pool are just copy paste of the original ones without any meaningful modification. Maybe we should remove those and go back to the canonical ones.

@apoorvsadana @EvolveArt wdyt?

tdelabro:
@notlesh the project doesn't build. Can you check the CI error and see if you can fix it?

apoorvsadana:
> the crates client/block-proposer and client/transaction-pool are just copy paste of the original ones without any meaningful modification. Maybe we should remove those and go back to the canonical ones.
> 
> @apoorvsadana @EvolveArt wdyt?

It does change the % of amount of time used during execution? One way would be to just increase the block time but this is another way. Although, I wonder why this would not be a part of the config directly. If it's not safe to change it like we've done right now, ya we might as well use the default on in that case

notlesh:
> the crates client/block-proposer and client/transaction-pool are just copy paste of the original ones without any meaningful modification. Maybe we should remove those and go back to the canonical ones.
> 
> @apoorvsadana @EvolveArt wdyt?

I wanted to follow up on this. What are the modifications that you made?

A large portion of the dependency upgrade effort went into maintaining this copied code, so there is a maintenance burden in keeping it around.

notlesh:
> @notlesh the project doesn't build. Can you check the CI error and see if you can fix it?

I believe the problem should be resolved by merging https://github.com/keep-starknet-strange/blockifier/pull/17 and then bumping `blockifier` in this PR once done.

apoorvsadana:
We changed the time for block production here - https://github.com/apoorvsadana/madara/blob/main/crates/client/block-proposer/src/lib.rs#L226-L226 @notlesh

notlesh:
> We changed the time for block production here - https://github.com/apoorvsadana/madara/blob/main/crates/client/block-proposer/src/lib.rs#L226-L226 @notlesh for the block proposer.

Ah, thanks for pointing that out. Quite a humble looking change!

There are probably other ways to accomplish this, I'd be happy to explore alternatives that don't require so much duplicated code.

For starters, there are some runtime constants that affect block time, such as https://github.com/paritytech/polkadot-sdk/blob/32a974088c43d9abf849b15349268e0a8426845a/substrate/bin/node-template/runtime/src/lib.rs#L121.

Also see the `fn slot_duration` in the Aura impl in the same file.

tdelabro:
@notlesh Just merged the one on blockifier. You can bump it. Do you also want to remove the two duplicated crates in this PR? I thinks it's would be for the best

notlesh:
> @notlesh Just merged the one on blockifier. You can bump it. Do you also want to remove the two duplicated crates in this PR? I thinks it's would be for the best

Thanks.

I think removing the duplicated crates is best left for another PR. For one thing, I'm not sure yet how to do that without knowing that anything broke. I could also use more context; for example, what is the need for more time when producing a block? Is it hardware-dependent, and would I perhaps not run into the problem on a fast Ryzen CPU?

notlesh:
Quick update on the compile errors seen in CI: This is caused by https://github.com/paritytech/polkadot-sdk/issues/2327. It seems Polkadot `v1.3.0` is just broken and I expect they'll release a `v1.3.1` soon to fix this.

I may look for a workaround (I was somehow able to circumvent this on my local machine earlier, although now I cannot). But otherwise it may be best to wait for an upstream fix.

notlesh:
> This is caused by [paritytech/polkadot-sdk#2327](https://github.com/paritytech/polkadot-sdk/issues/2327) 

I was able to workaround the problem in [55f0a68](https://github.com/keep-starknet-strange/madara/pull/1264/commits/55f0a6805d87907229fdf03820164ed78594bd23) with a crate override.

notlesh:
Looks like CI is green, PTAL @tdelabro :)

tdelabro (approved):

PR #1267, opened 2023-11-17
fix(starknet_rpc/get_transaction_receipt): async loop to get pending tx

lana-shanghai:

Dojo wasn't aware of pending transactions in the Madara tx pool. Pending tx was returing as not found. 

## What is the new behavior?

We loop for 1 minute to make sure the transaction is processed. 

No

### DIFF
--- a/crates/client/rpc/Cargo.toml
+++ b/crates/client/rpc/Cargo.toml
@@ -16,7 +16,7 @@ repository = "https://github.com/keep-starknet-strange/madara"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
-tokio = {workspace = true, default-features = true, features = ["time"]}
+tokio = { workspace = true, default-features = true, features = ["time"] }
 # Madara runtime
 pallet-starknet = { workspace = true, default-features = true }
 # Madara client
--- a/crates/client/rpc/Cargo.toml
+++ b/crates/client/rpc/Cargo.toml
@@ -16,7 +16,6 @@ repository = "https://github.com/keep-starknet-strange/madara"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
-tokio = { workspace = true, default-features = true, features = ["time"] }
 # Madara runtime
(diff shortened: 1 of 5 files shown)


crates/client/rpc/src/lib.rs
```diff
@@ -888,24 +889,53 @@ where
...
+            wait_for_tx_inclusion(self.backend.clone(), transaction_hash),
+        )
+        .await
+        {
+            Err(_) => {
+                error!("did not receive tx hash within 1 minute");
                 return Err(StarknetRpcApiError::TxnHashNotFound.into());
             }
-        };
+            Ok(tx_hash) => tx_hash,
```
tdelabro, line 937: this is `res` rather than than tx_hash

crates/client/rpc/Cargo.toml
```diff
@@ -16,6 +16,7 @@ repository = "https://github.com/keep-starknet-strange/madara"
 targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
+tokio = { workspace = true, default-features = true, features = ["time"] }
```
tdelabro, line 19: This can be set under section
`# Others`

crates/client/rpc/src/lib.rs
```diff
@@ -888,24 +889,53 @@ where
...
+        ) -> Result<<B as BlockT>::Hash, StarknetRpcApiError> {
+            let substrate_block_hash;
+
+            loop {
+                let block_hash_from_db = madara_backend
+                    .mapping()
+                    .block_hash_from_transaction_hash(H256::from(transaction_hash.to_bytes_be()))
+                    .map_err(|e| {
+                        error!("Failed to get transaction's substrate block hash from mapping_db: {e}");
+                        StarknetRpcApiError::TxnHashNotFound
```
tdelabro, line 908: This is more a `failed to interact with db backend error`
`StarknetRpcApiError::InternalError`

tdelabro (changes requested):

tdelabro (approved):

PR #1274, opened 2023-11-22
chore: split starknetRPC in two traits to follow openRPC spec definition

tdelabro:
https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json

StarknetRPC specs are split between multiple groups. This PR imitate this separation by creating two traits

### DIFF
--- a/crates/client/rpc-core/src/lib.rs
+++ b/crates/client/rpc-core/src/lib.rs
@@ -27,6 +27,31 @@ use starknet_core::types::{
 #[derive(Serialize, Deserialize)]
 pub struct Felt(#[serde_as(as = "UfeHex")] pub FieldElement);
 
+/// Starknet write rpc interface.
+#[rpc(server, namespace = "starknet")]
+pub trait StarknetWriteRpcApi {
+    /// Submit a new transaction to be added to the chain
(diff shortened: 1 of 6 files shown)


crates/client/rpc-core/src/lib.rs
```diff
@@ -27,6 +27,31 @@ use starknet_core::types::{
...
+    #[method(name = "addDeclareTransaction")]
+    async fn add_declare_transaction(
+        &self,
+        declare_transaction: BroadcastedDeclareTransaction,
+    ) -> RpcResult<DeclareTransactionResult>;
+}
+
 /// Starknet rpc interface.
 #[rpc(server, namespace = "starknet")]
 pub trait StarknetRpcApi {
```
apoorvsadana, line 57: I know the Starknet spec has it like this but will it make it more readable if we call this `StarknetReadRpcApi`?

apoorvsadana (approved):

crates/client/rpc-core/src/lib.rs
tdelabro: Good idea. I will do the change

EvolveArt (approved):

PR #1279, opened 2023-11-25
Client thread responsible for state updates and messaging on Ethereum

m-kus:
A rewrite of #1210


crates/pallets/starknet/src/lib.rs
```diff
@@ -167,6 +167,8 @@ pub mod pallet {
         type ChainId: Get<Felt252Wrapper>;
         #[pallet::constant]
         type MaxRecursionDepth: Get<u32>;
+        #[pallet::constant]
+        type ProgramHash: Get<Felt252Wrapper>;
```
tdelabro, line 171: I'm not sure about that. The StarknetOS is executed outside of the runtime. And I don't see reason of making the sequencer aware of it.

starknet-e2e-test/ethereum_core_contract.rs
```diff
@@ -0,0 +1,131 @@
...
+
+    let program_output = StarknetOsOutput {
+        new_state_root: 1u64.into(),
+        block_number: 1u64.into(),
+        config_hash: 1u64.into(),
+        messages_to_l1: vec![message.clone()],
+        ..Default::default()
+    };
+
+    // During the state update, the message will be consumed (removed from hash table)
```
tdelabro, line 125: I don't understand this

Issue #1298, opened 2023-12-06
dev: Implement tests for new rpc method starknet_getTransactionStatus


antiyro:
## Implement tests for new rpc method starknet_getTransactionStatus

The goal of this issue is to implement tests regarding the recently implemented rpc method starknet_getTransactionStatus under `starknet-rpc-tests/`

ybensacq:
@antiyro can you assign it to me
Plz ?

antiyro:
Sure if you need any help please let me know!

fishseabowl:
@tdelabro Could I take this? Thanks

tdelabro:
@fishseabowl sure

Issue #1291, opened 2023-12-04
dev: Upgrade Madara JSON-RPC specs to v0.6.0-rc5


antiyro:
# Upgrade Madara JSON-RPC specs to v0.6.0-rc5

## Overview
This issue concerns the implementation of the latest version of Starknet's JSON-RPC API: [v0.6.0-rc5](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.6.0-rc5) within Madara (currently on [v0.4.0](https://github.com/starkware-libs/starknet-specs/releases/tag/v0.4.0)) to ensure and maintain full compatibility with Starknet.

## Detailed Description
Since version 0.4.0 of Starknet RPC, here's what has changed:

### v0.5.1 New Features
- [x] **General Improvements**: Focused on covering information gaps between the feeder-gateway and the RPC.
- [x] **Extracted `broadcasted declare txn`**: Separated from `declare txn`.
- [x] **Schema Enhancements**: Fixed references and added "required" to write-API results.
- [x] **Pending Block Schema Fixes**: Adjustments for pending block & block with hash.
- [x] **Broadcasted Transaction**: Reintroduced with query version capability.
- [x] **Receipts & Block Header Fields**: 
  - [x] Introduced `starknet_specVersion` endpoint.
  - [x] Added type field to transaction trace objects.
  - [x] Added order field to events/messages in transaction trace.
  - [x] Added `l1_gas_price` and `starknet_version` to block header.
  - [x] Split transaction receipt objects by type with `execution_resources`.
- [x] **Trace Improvement**: Added state diffs and order to events/messages.
- [x] **`CONTRACT_ERROR` Enhancement**: Included string data.
- [x] **Transaction Status API**: Added `starknet_getTransactionStatus` and removed `starknet_pendingTransactions`.
- [x] **`PENDING_STATE_UPDATE` Schema Fix**: Addressed schema issues.
- [x] **`broadcasted_declare_txn_v1` Fields**: Added missing fields.
- [x] **Error Type Change**: Modified `traceBlockTransaction` error type.
- [x] **Pending Block Header**: Added missing fields.

### v0.6.0 New Features
- **v0.6.0-rc5**:
  - [ ] Imported `BROADCASTED_DECLARE_TXN` in write API.
  - [ ] Added string data to `VALIDATION_FAILURE`.
- **v0.6.0-rc4**:
  - [ ] Typo fix and property addition to schema object.
[...]

PR #1299, opened 2023-12-06
feat(rpc): Added starknet_getTransactionStatus and removed starknet_pendingTransactions

antiyro:
## feat(rpc): Added starknet_getTransactionStatus and removed starknet_pendingTransactions

This PR solves the implementation of a new call: starknet_getTransactionStatus and the removal of starknet_pendingTransactions according to the v0.5.1 RPC specs described in #1291 

`starknet_getTransactionStatus`

Gets the Transaction Status, Including Mempool Status and Execution Details
    
This method retrieves the status of a specified transaction. It provides information on
whether the transaction is still in the mempool, has been executed, or dropped from the
mempool. The status includes both finality status and execution status of the
transaction.

### Arguments

* `transaction_hash` - The hash of the transaction for which the status is requested.

### Returns

* `transaction_status` - An object containing the transaction status details:
 - `finality_status`: The finality status of the transaction, indicating whether it is
   confirmed, pending, or rejected.
 - `execution_status`: The execution status of the transaction, providing details on the
   execution outcome if the transaction has been processed.

### Expected behavior

image

## Notes

- TODO: https://github.com/antiyro/madara/blob/feat/starknet_getTransactionStatus/crates/primitives/transactions/src/lib.rs#L32
- Tests: The call has been tested manually in multiple cases but official tests needs to be implemented here: [#1298](https://github.com/keep-starknet-strange/madara/issues/1298)


crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+    ///
+    /// ### Returns
+    ///
+    /// * `transaction_status` - An object containing the transaction status details:
+    ///   - `finality_status`: The finality status of the transaction, indicating whether it is
+    ///     confirmed, pending, or rejected.
+    ///   - `execution_status`: The execution status of the transaction, providing details on the
+    ///     execution outcome if the transaction has been processed.
+    fn get_transaction_status(&self, transaction_hash: FieldElement) -> RpcResult<TransactionStatus> {
+        let block_hash_from_db = self
```
tdelabro, line 423: rename to `substrate_block_hash`

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+            .backend
+            .mapping()
+            .block_hash_from_transaction_hash(H256::from(transaction_hash.to_bytes_be()))
+            .map_err(|e| {
+                error!("Failed to get transaction's substrate block hash from mapping_db: {e}");
+                StarknetRpcApiError::TxnHashNotFound
+            })?;
+
+        let substrate_block_hash = block_hash_from_db.ok_or(StarknetRpcApiError::TxnHashNotFound)?;
```
tdelabro, line 432: chain those operations

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+
+        let block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash)
```
tdelabro, line 434: is it a substrate or a starknet block? Make it explicit in the variable name

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+            .ok_or(StarknetRpcApiError::BlockNotFound)?;
+
+        let chain_id = self.chain_id()?.0.into();
+
+        let find_tx = block
+            .transactions()
+            .iter()
+            .find(|tx| tx.compute_hash::<H>(chain_id, false).0 == transaction_hash)
```
tdelabro, line 442: don't we maintain a dbstorage mapping that spare us the need to recompute the tx hash.
I think we can retrieve it by blockhash and index.

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+            })?;
+
+        let substrate_block_hash = block_hash_from_db.ok_or(StarknetRpcApiError::TxnHashNotFound)?;
+
+        let block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash)
+            .ok_or(StarknetRpcApiError::BlockNotFound)?;
+
+        let chain_id = self.chain_id()?.0.into();
+
+        let find_tx = block
```
tdelabro, line 439: `found_tx` or just `tx`

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+            .map(|tx| to_starknet_core_tx(tx.clone(), transaction_hash))
+            .ok_or(StarknetRpcApiError::TxnHashNotFound)?;
+
+        let execution_result: TransactionExecutionStatus = self
+            .client
+            .runtime_api()
+            .get_tx_execution_outcome(substrate_block_hash, Felt252Wrapper(transaction_hash).into())
+            .map_err(|e| {
+                error!(
+                    "Failed to get transaction execution outcome. Substrate block hash: {substrate_block_hash}, \
```
tdelabro, line 452: `Failed to get transaction execution outcome through StarknetRuntimeApi. ...`

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+                     transaction hash: {transaction_hash}, error: {e}"
+                );
+                StarknetRpcApiError::InternalServerError
+            })?
+            .map_or(TransactionExecutionStatus::Succeeded, |message| TransactionExecutionStatus::Reverted);
```
tdelabro, line 457: `message` is not used? We discard the error message? 
Is it the expected behaviour of the RPC?

crates/client/rpc/src/lib.rs
```diff
@@ -400,6 +401,67 @@ where
...
+
+        Ok(TransactionStatus {
+            finality_status: TransactionFinalityStatus::AcceptedOnL2,
+            execution_status: execution_result,
+        })
+    }
```
tdelabro, line 463: What about the other finality status?
Recieved (in the memepool)
Rejected (could not be put in the mempool)

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
antiyro: My bad I forgot about it, will correct it inspired on `get_transaction_by_hash`

crates/client/rpc/src/lib.rs
antiyro: and `get_transaction_receipt`

crates/client/rpc/src/lib.rs
antiyro: Updated it in all other rpc calls for concensus

crates/client/rpc/src/lib.rs
antiyro: I used ExecutionResult back then but I confirm that message should not be used: `The execution status of the transaction, providing details on the execution outcome if the transaction has been processed`

crates/client/rpc/src/lib.rs
antiyro: I used `starknet_tx` for concensus wdyt

crates/client/rpc/src/lib.rs
tdelabro: Okay, but create the required issue with enough background for somebody to jump in

crates/client/rpc-core/Cargo.toml
```diff
@@ -32,6 +32,7 @@ jsonrpsee = { workspace = true, features = [
 ], default-features = true }
 mp-block = { workspace = true }
 mp-digest-log = { workspace = true }
+mp-transactions = { workspace = true, features = ["serde"] }
```
tdelabro, line 35: if you do that `serde = ["mp-transactions/serde"]`you don't need to do this `features = ["serde"]`

tdelabro (changes requested):

crates/client/rpc-core/Cargo.toml
tdelabro: But it look like serde is always imported. So its not an optional feature. So your most likely want `features = ["serde"]` and not `serde = ["mp-transactions/serde"]`

crates/client/rpc/src/lib.rs
tdelabro: you forgot this one

crates/client/rpc/src/lib.rs
tdelabro: And this one too

crates/client/rpc-core/Cargo.toml
antiyro: Done!

crates/client/rpc/src/lib.rs
antiyro: I did it in last commits

crates/client/rpc/src/lib.rs
antiyro: Same as above conversation

tdelabro (approved):

PR #1305, opened 2023-12-08
Use code from pathfinder to create byte array for path to avoid uninitialized memory

AurelienFT:
# Pull Request type

- Bugfix

In the merkle tree, in order to convert the path to a sequence of bytes to use it in the hash, the code used `into_vec()` method of bitvec. Following [their doc](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.into_vec) it's needed to call [.set_uninitialized()](https://docs.rs/bitvec/latest/bitvec/vec/struct.BitVec.html#method.set_uninitialized) before which was not the case in the previous code. This was causing making an hash on random uninitialized bytes and leads to possible wrong hash.

Resolves: #NA

## What is the new behavior?

In Pathfinder this logic as already been modified and so I change the code here using the one of Pathfinder that should produce a coherent output.

No

## Other information

/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- fix: Change seliazation of bitvec to &[u8] in merkle tree to avoid memory uninitialized
 - refacto: substrate/starknet names in rpc library
 - feat(rpc): Added starknet_getTransactionStatus and removed
   starknet_pendingTransactions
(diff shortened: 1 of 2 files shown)


tdelabro:
@AurelienFT can you add something in the CHANGELOG so that the CI can run?

tdelabro:
superseded by #1308

I couldn't push the CI fix directly to your repo brach. I don't have the permissions.

PR #1287, opened 2023-11-29
✨ v0.11 calldata parsing

EvolveArt:
Rebase and bug fixes of #1209

PR #1312, opened 2023-12-11
Merge l1 l2

tdelabro:


crates/node/src/commands/run.rs
```diff
@@ -86,13 +132,47 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
             None
         }
     };
+
+    let l1_messages_worker_config = extract_l1_messages_worker_config(&cli.run.l1_messages_worker)
+        .map_err(|e| sc_cli::Error::Input(e.to_string()))?;
+    let settlement_config: Option<(SettlementLayer, PathBuf)> = match cli.run.settlement {
+        Some(SettlementLayer::Ethereum) => {
+            let config_path = data_path.join("eth-config.json");
```
EvolveArt, line 140: shouldn't we have a cli param to specify the config path ? Like we have for DA layer in #1287

crates/node/src/service.rs
```diff
@@ -441,16 +449,35 @@ pub fn new_full(
...
+    if let Some((layer_kind, config_path)) = settlement_config {
+        let settlement_provider: Box<dyn SettlementProvider<_>> = match layer_kind {
+            SettlementLayer::Ethereum => {
+                let ethereum_conf = EthereumConfig::try_from(&config_path)?;
+                Box::new(
+                    StarknetContractClient::try_from(ethereum_conf).map_err(|e| ServiceError::Other(e.to_string()))?,
+                )
+            }
+        };
+        let retry_strategy = Box::new(RetryOnRecoverableErrors { delay: Duration::from_millis(100) });
```
EvolveArt, line 472:
```suggestion
        let retry_strategy = Box::new(RetryOnRecoverableErrors { delay: DEFAULT_SETTLEMENT_RETRY_MS });
```

crates/pallets/starknet/src/lib.rs
```diff
@@ -659,6 +650,17 @@ pub mod pallet {
...
+            let tx_hash = transaction.tx_hash;
+            let nonce: Nonce = transaction.tx.nonce;
+
+            // Ensure that L1 Message has not been executed
+            Self::ensure_l1_message_not_executed(&nonce).map_err(|_| Error::<T>::L1MessageAlreadyExecuted)?;
+
+            // Store infornamtion about message being processed
+            // The next instruction executes the message
+            // Either successfully  or not
+            L1Messages::<T>::mutate(|nonces| nonces.insert(nonce));
```
EvolveArt, line 662: Aren't there scenarios where the execution fails for different reasons ? Like the sequencer does not have enough ETH to pay the fees. In that case, we wouldn't want to mark it as executed but would be fine with it being retried next block.

crates/pallets/starknet/src/tests/l1_message.rs
```diff
@@ -1,61 +1,160 @@
...
+        basic_test_setup(2);
+        let contract_address =
+            Felt252Wrapper::from_hex_be("0x024d1e355f6b9d27a5a420c8f4b50cea9154a8e34ad30fc39d7c98d3c177d0d7").unwrap();
+        let from_address = Felt252Wrapper::from_hex_be("0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045").unwrap();
+
+        let transaction = HandleL1MessageTransaction {
+            nonce: 1,
+            contract_address,
+            entry_point_selector: Felt252Wrapper::from_hex_be(
+                "0x014093c40d95d0a3641c087f7d48d55160e1a58bc7c07b0d2323efeeb3087269",
```
EvolveArt, line 110: what does that correspond to ?

crates/primitives/program-hash/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use mp_felt::Felt252Wrapper;
+
+/// ProgramHash for Starknet OS Cairo program
+///
+/// How to calculate:
+///     1. Get Starknet OS program sources (e.g. check keep-starknet-strange/snos)
+///     2. Install Cairo-lang and run `cairo-hash-program --program <cairo-output>.json`
+///     3. Install Starqli and run `starqli mont <program hash>`
```
EvolveArt, line 11:
```suggestion
///     3. Install Starkli and run `starkli mont <program hash>`
```

crates/primitives/transactions/src/execution.rs
```diff
@@ -547,8 +551,7 @@ impl Execute for L1HandlerTransaction {
         block_context: &BlockContext,
         _account_tx_context: AccountTransactionContext,
     ) -> TransactionExecutionResult<(Fee, Option<CallInfo>, ResourcesMapping)> {
-        // The calldata includes the "from" field, which is not a part of the payload.
-        let l1_handler_payload_size = self.calldata().0.len() - 1;
+        let l1_handler_payload_size = self.calldata().0.len();
```
EvolveArt, line 554: is the `from` field part of the payload now ?

madara-test-runner/src/node.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+fn get_repository_root() -> PathBuf {
+    let manifest_path = Path::new(&env!("CARGO_MANIFEST_DIR"));
+    let repository_root = manifest_path.parent().expect("Failed to get parent directory of CARGO_MANIFEST_DIR");
+    repository_root.to_path_buf()
+}
+
+impl MadaraNode {
+    fn cargo_run(root_dir: &Path, options: Vec<Option<String>>) -> Child {
```
EvolveArt, line 51: rustdoc + should it always be release by default?

starknet-e2e-test/ethereum_e2e_settlement.rs
```diff
@@ -0,0 +1,176 @@
...
+#[test_context(Context)]
+#[rstest]
+#[tokio::test]
+async fn works_with_initialized_contract(ctx: &mut Context) -> Result<(), anyhow::Error> {
+    // Troubleshooting:
+    // RUST_LOG=mc_settlement=trace MADARA_LOG=1 cargo test --package starknet-e2e-test
+    // works_with_initialized_contract -- --nocapture
+
+    // At this point we have:
+    //   * spawnined Ethereum sandbox
```
EvolveArt, line 81:
```suggestion
    //   * spawned Ethereum sandbox
```

starknet-e2e-test/src/starknet_contract.rs
```diff
@@ -0,0 +1,187 @@
...
+        bytes[96..128].copy_from_slice(val.config_hash.bytes());
+        bytes[128..160].copy_from_slice(val.state_root.bytes());
+        bytes[160..192].copy_from_slice(val.block_number.bytes());
+        bytes[192..224].copy_from_slice(val.block_hash.bytes());
+        bytes.into()
+    }
+}
+
+impl InitData {
+    pub fn consistent() -> Self {
```
EvolveArt, line 80: add rustdoc, consistent here means that it corresponds to the current state of Starknet Goerli Testnet

EvolveArt (changes requested):
First round of review.

I tend to think that a lot of logic present in ethereum sandbox crate, responsible for deploying and managing settlement on starknet core contract or starknet validium contract should stay zaun's responsibility in the long run. Or at least shouldn't be internal to madara wdyt ?

crates/primitives/snos-output/src/codec.rs
```diff
@@ -0,0 +1,144 @@
...
+            from_address: StarkFelt::decode(input)?,
+            to_address: StarkFelt::decode(input)?,
+            payload: Vec::<StarkFelt>::decode(input)?,
+        })
+    }
+}
+
+impl SnosCodec for MessageL1ToL2 {
+    fn size_in_felts(&self) -> usize {
+        5 + self.payload.len()
```
apoorvsadana, line 91: same as above

crates/primitives/snos-output/src/codec.rs
```diff
@@ -0,0 +1,144 @@
...
+            nonce: StarkFelt::decode(input)?,
+            selector: StarkFelt::decode(input)?,
+            payload: Vec::<StarkFelt>::decode(input)?,
+        })
+    }
+}
+
+impl SnosCodec for StarknetOsOutput {
+    fn size_in_felts(&self) -> usize {
+        9 + self.messages_to_l1.iter().map(|msg| msg.size_in_felts()).sum::<usize>()
```
apoorvsadana, line 115: same as above

starknet-rpc-test/src/utils.rs
```diff
@@ -87,6 +87,14 @@ pub trait AccountActions {
         nonce: Option<u64>,
     ) -> TransactionExecution;
 
+    fn invoke_contract(
```
apoorvsadana, line 90: Let's call invoke_contract inside transfer_tokens. All the code in invoke_contract is duplicated in transfer_tokens.

crates/primitives/snos-output/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub state_updates: Vec<StarkFelt>,
+    /// List of the newly declared contract classes.
+    pub contract_class_diff: Vec<StarkFelt>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "scale-info", derive(scale_info::TypeInfo))]
+#[cfg_attr(feature = "parity-scale-codec", derive(parity_scale_codec::Encode, parity_scale_codec::Decode))]
+/// Message sent to L1 by invoking according Statknet syscall
+pub struct MessageL2ToL1 {
```
apoorvsadana, line 49: now that simulate txn is merged should we use this struct from there (or move the struct to the transactions primitive and import it at both places)

crates/primitives/snos-output/src/codec.rs
```diff
@@ -0,0 +1,144 @@
...
+        while segment_reader.remaining_len() > 0 {
+            elements.push(T::decode(&mut segment_reader)?);
+        }
+        Ok(elements)
+    }
+}
+
+impl SnosCodec for MessageL2ToL1 {
+    fn size_in_felts(&self) -> usize {
+        3 + self.payload.len()
```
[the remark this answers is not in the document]
tdelabro: But the size of a `StarkFelt` counted in felt, will always be 1.
It is its own unit of measure.
Am I missing something?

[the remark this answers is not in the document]
tdelabro: Okay. I get that. We can keep hardcoded `1` for the len of vecs, because those won't change. But you are right about the fields.
Good catch!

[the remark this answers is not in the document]
tdelabro: you are 100% right!

crates/client/settlement/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+async-trait = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
+futures = { workspace = true }
+futures-timer = { workspace = true }
+log = { workspace = true }
+rustc-hex = "2.0.0"
+serde = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
+url = "2.4.1"
```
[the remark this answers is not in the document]
tdelabro: done

crates/client/settlement/Cargo.toml
```diff
@@ -0,0 +1,52 @@
...
+license = "MIT"
+publish = false
+repository = "https://github.com/keep-starknet-strange/madara"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+async-trait = { workspace = true }
+clap = { workspace = true, features = ["derive"] }
```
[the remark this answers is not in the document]
tdelabro: done

crates/client/l1-messages/src/worker.rs
```diff
@@ -0,0 +1,140 @@
...
+    }?;
+
+    let extrinsic = client.runtime_api().convert_l1_transaction(best_block_hash, transaction, fee).map_err(|e| {
+        log::error!("⟠ Failed to convert L1 Transaction via Runtime Api: {:?}", e);
+        L1MessagesWorkerError::ConvertTransactionRuntimeApiError(e)
+    })?;
+
+    let tx_hash = pool.submit_one(best_block_hash, TX_SOURCE, extrinsic).await.map_err(|e| {
+        log::error!("⟠ Failed to submit transaction with L1 Message: {:?}", e);
+        L1MessagesWorkerError::SubmitTxError
```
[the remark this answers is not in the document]
tdelabro: done

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,62 @@
...
+        Self { http_provider, contract_address }
+    }
+
+    pub fn new_from_file(path: &PathBuf) -> Result<Self, L1MessagesWorkerConfigError> {
+        let file = File::open(path)?;
+        let config = serde_json::from_reader(file)?;
+        Ok(config)
+    }
+
+    pub fn new_from_params(provider_url: &str, contract_address: &str) -> Result<Self, L1MessagesWorkerConfigError> {
```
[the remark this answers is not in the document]
tdelabro: Builder patterns are for when each argument is an option. You can chose to set, or not, the different fields. 
Here there are only two, and both are required. So KISS

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,285 @@
...
+                log::info!("[settlement] Verifying state root for block {}", sync_to);
+                Self::verify_starknet_state(substrate_client, &last_settled_state)?;
+                continue;
+            }
+
+            log::info!("[settlement] Syncing state for blocks {} -> {}", sync_from, sync_to);
+            while sync_from < sync_to {
+                let (next_block, substrate_block_hash) = if sync_from + 1 == sync_to {
+                    // This is a typical scenario when we are up to speed with the chain
+                    (block.clone(), notification.hash)
```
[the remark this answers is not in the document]
tdelabro: This happen as soon as we are done syncing l2. Each new block recieve is the new head.
So yes it happen. And it has nothing to do with block time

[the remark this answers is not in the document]
tdelabro: This happens as soon as we are done syncing l2. Each new block received is the new head.
So yes it happens. And it has nothing to do with block time

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// sync), but eventually it will be ok.
+    ///
+    /// Secondly, we try to find a corresponding Substrate block (header) by its hash.
+    /// Lastly, we extract Starknet block from the Substrate block digest.
+    fn get_starknet_block(substrate_client: &SC, block_number: u64) -> Result<(StarknetBlock, B::Hash), B> {
+        let substrate_block_hash = substrate_client
+            .hash(UniqueSaturatedInto::unique_saturated_into(block_number))?
+            .ok_or_else(|| Error::UnknownStarknetBlock(block_number))?;
+
+        let substrate_block_header = substrate_client
```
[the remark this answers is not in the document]
tdelabro: Which "first query"?

crates/client/settlement/src/sync_state.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// If Madara chain haven't reached the given block level yet, it returns OK, assuming that as
+    /// soon as it catches up - this check will be done again.
+    fn verify_starknet_state(substrate_client: &SC, state: &StarknetState) -> Result<(), B> {
+        let height: u64 = state.block_number.try_into()?;
+
+        match Self::get_starknet_block(substrate_client, height) {
+            Ok((block, _)) => {
+                let state_root = block.header().global_state_root;
+                // Verify that current onchain state is consistent with corresponding Madara block
+                if state.state_root != state_root {
```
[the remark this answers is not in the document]
tdelabro: If it is not populated, it won't be set on l1 either. So this check should still pass.

crates/client/settlement/src/sync_state.rs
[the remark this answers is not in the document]
tdelabro: Which "first query"?

crates/client/settlement/src/sync_state.rs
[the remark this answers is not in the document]
tdelabro: This happens as soon as we are done syncing l2. Each new block received is the new head.
So yes it happens. And it has nothing to do with block time

crates/node/Cargo.toml
```diff
@@ -20,6 +20,9 @@ targets = ["x86_64-unknown-linux-gnu"]
 name = "madara"
 
 [dependencies]
+# Ethereum
+ethers = { version = "2.0.7", features = ["openssl"] }
```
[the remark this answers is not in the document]
tdelabro: done

crates/node/src/service.rs
tdelabro: done

crates/client/l1-messages/src/config.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    Url::parse(&endpoint).map_err(serde::de::Error::custom)
+}
+
+impl L1MessagesWorkerConfig {
+    pub fn new(http_provider: Url, contract_address: Address) -> Self {
+        Self { http_provider, contract_address }
+    }
+
+    pub fn new_from_file(path: &PathBuf) -> Result<Self, L1MessagesWorkerConfigError> {
```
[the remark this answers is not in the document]
tdelabro: I don't like using `From` for method that do io.
From is for simple type conversion. I would not use it if we are doing IO (disk or http) f any sort inside the function. Better make it explicit that yes, we are reading from a file.

`From` would be ok for conversion between `PathBuf` and `Path` or `String` or `Vev<u8>`

crates/pallets/starknet/src/lib.rs
tdelabro: To do this we need to spawn another service which monitor message from l1 extrinsic execution, and insert the transaction again if needed. It won't just "reexecute itself".

Fees for l1->l2 messages are paid on l1. No other fee is deducted during execution on l1. So it shouldn't fail this way.

crates/pallets/starknet/src/tests/l1_message.rs
tdelabro: added a comment

crates/primitives/program-hash/src/lib.rs
tdelabro: done

crates/primitives/transactions/src/execution.rs
tdelabro: mb

madara-test-runner/src/node.rs
tdelabro: let's keep it that way for now

starknet-e2e-test/ethereum_e2e_settlement.rs
tdelabro: done

starknet-e2e-test/src/starknet_contract.rs
tdelabro: done

crates/client/settlement/src/sync_state.rs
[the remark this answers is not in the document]
EvolveArt: the query for the block_hash, we try to get the block_hash and if it fails we try to get the block number

crates/pallets/starknet/src/lib.rs
EvolveArt: Well at least it should be noted that the sequencer always needs to have enough ETH on l2 to cover message fees otherwise they will *never* be executed

madara-test-runner/src/node.rs
EvolveArt: just precise it in the comment then cause it's not obvious

crates/client/settlement/src/sync_state.rs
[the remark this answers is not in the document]
EvolveArt: My bad thought this was syncing with L1

crates/pallets/starknet/src/lib.rs
tdelabro: What? No.
No fee are taken for l1-handler txs.
```rust
        let paid_fee = self.paid_fee_on_l1;
        // For now, assert only that any amount of fee was paid.
        // The error message still indicates the required fee.
        if paid_fee == Fee(0) {
            return Err(TransactionFeeError::InsufficientL1Fee { paid_fee, actual_fee })?;
        }
```

That's the only fee-related thing is the code.
Nobody needs any ETH to pay for anything on l2.

crates/client/settlement/src/sync_state.rs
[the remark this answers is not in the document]
tdelabro:
```rust
.hash(number: UniqueSaturatedInto::unique_saturated_into(self: block_number))?
```
This one?
This is not an `UnknownStarknetBlock` error. This is a `CouldNotCommunicateProperlyWithSubstrateDatabase`.

crates/primitives/snos-output/src/codec.rs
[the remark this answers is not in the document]
tdelabro: done

crates/primitives/snos-output/src/codec.rs
tdelabro: done

crates/primitives/snos-output/src/codec.rs
tdelabro: done

starknet-rpc-test/src/utils.rs
tdelabro: done

madara-test-runner/src/node.rs
tdelabro: done

crates/primitives/snos-output/src/codec.rs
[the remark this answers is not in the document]
tdelabro: I kept hardcoded `3` or `5` for fields that are felts. Same for a vec of felt: i use it's length rather than iterating over and calling size_in_felt

crates/primitives/snos-output/src/lib.rs
tdelabro: done

crates/pallets/starknet/src/tests/transaction_validation.rs
apoorvsadana, line 1: lets rename file to `l1_transaction_validation` or lets add test cases for non L1 txns as well

starknet-e2e-test/ethereum_e2e_settlement.rs
```diff
@@ -0,0 +1,176 @@
...
+    let mut madara = ctx.launch_madara().await;
+
+    madara.create_empty_block().await?;
+    // Give the client thread some time to handle the finalized block
+    sleep(Duration::from_millis(100)).await;
+
+    // Expected connection refused because Madara is shut down at this point
+    let is_connection_refused =
+        assert_matches!(madara.create_empty_block().await, Err(err) => err.downcast_ref::<reqwest::Error>().is_some());
+    assert!(is_connection_refused);
```
apoorvsadana, line 153: why are we asserting the result of a result?

starknet-e2e-test/ethereum_e2e_settlement.rs
```diff
@@ -0,0 +1,176 @@
...
```
apoorvsadana, line 173: same as above

starknet-rpc-test/Cargo.toml
```diff
@@ -26,9 +26,9 @@ thiserror = { workspace = true }
 tokio = { version = "1.34.0", features = ["rt", "macros", "parking_lot"] }
 url = "2.4.1"
 
-[[test]]
-name = "starknet_spec_version"
-path = "spec_version.rs"
+# [[test]]
```
apoorvsadana, line 29: ?

crates/primitives/snos-output/src/codec.rs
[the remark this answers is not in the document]
apoorvsadana: But doesn't that beat the purpose again? The idea is to use size_in_felts() wherever possible. You don't need to iterate, just call size_in_felts() (which would internally iterate but thats abstracted from you)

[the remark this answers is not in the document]
tdelabro: It should not iterate. This method is to be used for optimization purposes. To avoid relocation as your buffer grows. Best, case scenario you get the exact predicted size, you allocate with this capacity, and no reallocation happens.

The worst, case scenario is that you create the default vec with no capacity, and relocate a few times as you write potentially thousands and thousands of bytes.

Here we will end up in the middle.

But iterating over a vec is potentially wayyyy more expensive than doing a reallocation. So we don't want to iterate under any condition, it would beat the whole purpose of doing this estimation.
We prefer a not perfect but okayish that will help us in most cases, than the worst-case scenario. Not perfect but still useful.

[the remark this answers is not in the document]
tdelabro: The way your are supposed to use it is the following:
```rust
let data_to_encode = SomeType::new();
let buffer: Vec<StarkFelt> = Vec::new(data_to_encode::size_in_felt());
data_to_encode.encode(buffer);
```

If you replace the second line by `Vec::new()`, without having a hint of the encoded size, the os will reallocate multiple times.
For 100 elements, there will be 5 relocations. Relocation is one of the most expensive operations there is for a program.

Here is a snippet to help you visualize: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c973a9fcc1354167a95f315a5b063bae

Now if you allocate use `with_capcity`. You will only allocate once if you have the perfect estimation.
If you don't you will probably have to relocate a few times, but you are already sparing some useless computation..

But all this optimization is wasted if you have to iterate over your elements. It is overall still cheaper to reallocate, than to iterate over an array (for struct that are not too small).

crates/client/settlement/src/ethereum/client.rs
```diff
@@ -0,0 +1,105 @@
...
+// Read this great overview to learn more about SNOS:
+// https://hackmd.io/@pragma/ByP-iux1T
+abigen!(
+    StarknetContract,
+    r#"[
+        function programHash() public view returns (uint256)
+        function stateBlockNumber() external view returns (int256)
+        function stateRoot() external view returns (uint256)
+        function configHash() public view returns (uint256)
+        function updateState(uint256[] calldata programOutput) external
```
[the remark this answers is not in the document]
apoorvsadana: nevermind, just saw this is for our custom contract

crates/node/src/commands/run.rs
apoorvsadana: also, there seems to be no way to have this config created with the setup command?

tdelabro: Fixed @EvolveArt comment.
@apoorvsadana I don't think the setup should create a default one, at least for now.

I also created https://github.com/keep-starknet-strange/madara/issues/1334, for someone to pick later

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 ## Next release
 
+- fix(node/commands): md5 are aslo checked when running setup --from-local
```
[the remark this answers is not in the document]
tdelabro: done

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
...
+- feat(data-availability): extend eth config with poll interval
+- fix(snos-output): expose snos codec, remove unused `get_starknet_messages`
+  runtime method, and unnecessary mp-snos-output dependencies
+- feat(program-hash): add new pallete constant for Starknet OS progam hash;
```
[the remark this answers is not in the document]
tdelabro: has been fixed

crates/client/l1-messages/src/worker.rs
```diff
@@ -0,0 +1,141 @@
...
+        log::error!("⟠ Failed to convert L1 Transaction via Runtime Api: {:?}", e);
+        L1MessagesWorkerError::ConvertTransactionRuntimeApiError(e)
+    })?;
+
+    let tx_hash = pool.submit_one(best_block_hash, TX_SOURCE, extrinsic).await.map_err(|e| {
+        log::error!("⟠ Failed to submit transaction with L1 Message: {:?}", e);
+        L1MessagesWorkerError::SubmitTxError(e)
+    })?;
+
+    backend.messaging().update_last_synced_l1_block(l1_block_number).map_err(|e| {
```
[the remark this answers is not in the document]
tdelabro: this is very correct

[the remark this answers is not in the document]
tdelabro: If there is no message for a block, there will be nothing pushed to `event_stream`. So there will be no triggering of our code.

crates/node/src/commands/setup.rs
```diff
@@ -70,60 +149,49 @@ impl SetupCmd {
...
+            ConfigSource::Remote(url)
         } else {
             unreachable!(
                 "clap::Args is derived upon `SetupSource` in a way that guarantee that either `from_remote` or \
                  `from_local` is present"
             );
+        };
+
+        let configs_file_content = config_source.load_config()?;
+        // Make sure it is valid data before writing it to disck
```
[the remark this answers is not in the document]
tdelabro: has been fixed

crates/pallets/starknet/src/tests/transaction_validation.rs
tdelabro: renamed to `l1_handler_validation`

starknet-e2e-test/ethereum_e2e_settlement.rs
tdelabro: replaced by `    assert_matches!(madara.create_empty_block().await, Err(err) => assert!(err.downcast_ref::<reqwest::Error>().is_some()));`
we want to be sure that it fails because of a request error, but we have an `anyhow` error, so we need to downcast it

starknet-e2e-test/ethereum_e2e_settlement.rs
tdelabro: same as above

starknet-rpc-test/Cargo.toml
tdelabro: This test is not ready and triggers clippy unused variables and so on. Better ignore it for noew

crates/client/l1-messages/src/worker.rs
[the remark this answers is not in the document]
apoorvsadana: Shouldn't our code ideally listen to new Eth blocks and update the last synced block every time? I mean, right now, let's say I start an app chain and I have been running it for like 1 month but noone ever sent a message from the L1. As a result, no event was ever fired and so the last sync block never got updated. Later, if I shut down and start my node again it will listen from the first block.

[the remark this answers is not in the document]
tdelabro: The lib is just a wrapper around etherum rpc: https://www.quicknode.com/docs/ethereum/eth_getFilterChanges
So that's not really how it works.

PR #1318, opened 2023-12-15
first tx after account deploy is valid

apoorvsadana:
Currently, this is what happens when you deploy an Argent Account on Starknet vs Madara

In Starknet
1. Send deploy_account txn => accepted
3. Immediately send invoke txn => accepted

In Madara
1. Send deploy_account txn => accepted
2. Immediately send invoke txn => rejected because deploy account is still in mempool and so we can't validate this txn

The fix implemented in this PR is to assume tx with nonce 1 to be valid if the account isn't deployed yet. In this case, tx with nonce 1 enters the mempool without being validated and is executed after the account is deployed. In case it fails, it won't be included in the Starknet block but will be in the Substrate block.

This PR also adds the `MADARA_CHAIN_ID` as there is an issue in Argent where simulation fails with SN_GOERLI chain_id because Argent uses the incorrect RPC.

**Why not assume all future txs to be valid and validate them during execution like we did for nonce 1?**
This opens the attack vector that I can send a large number of invalid txs which will enter the mempool. Then once I send the tx with the current nocne, all these future txs will be executed one by one and will fill the Substrate block + won't be chargeable because validation fails.

**Why not validate everything again in pre dispatch**
This affects the TPS by a lot. When I tried it, the TPS went from 39.8 -> 29.3.

Some points
1. We could try in the future to to store txs which skiped validation in an off chain storage and revalidate them during pre_dispatch
2. Add batched txs so you can send deploy and invoke together?


crates/pallets/starknet/src/lib.rs
```diff
@@ -775,25 +777,36 @@ pub mod pallet {
...
+                            contract_address,
+                        )),
+                    )) => {
+                        // If the the txn is failing because the account isn't deployed yet
+                        // then don't throw an error and add a `require` tag for transaction
+                        // of nonce 0 to be executed first.
+                        if contract_address.0.0 != sender_address.into()
+                            || transaction_nonce != Some(Felt252Wrapper::ONE)
+                        {
+                            Err(InvalidTransaction::BadProof)?
```
lana-shanghai, line 805: Is there any point to differentiate between the error when the nonce isn't 0 or 1, and other cases? Like below the error says the same, `BadProof`. Could be more verbose, wdyt?

lana-shanghai (approved):
Foundry CI failing again. Otherwise LGTM

crates/pallets/starknet/src/lib.rs
apoorvsadana: There is no error thrown when the nonce is 1 and the failure reason is `UninitializedStorageAddress`. All other cases its `BadProof`

PR #1341, opened 2023-12-30
fix: call proving only after update state

apoorvsadana:
There is no ordering between `prove_current_block` and `update_state`, any function can be called first. For context, `prove_current_block` updates the state diff in the db and `update_state` fetches those diffs. 

Case 1: `prove_current_block` is called first, DA works fine `store_state_diff` is called before `state_diff` which gets the state diff
Case 2: `update_state` is called first, DA throws an error as it first calls `state_diff` and then `store_state_diff`.

Since in the validity case, the proof be dependent on the DA, we should be calling `update_state` first. So we need to move the `store_state_diff` to update_state, call it first and then call `prove_current_block` at the end of update state.

### DIFF
--- a/crates/client/data-availability/src/lib.rs
+++ b/crates/client/data-availability/src/lib.rs
@@ -14,15 +14,12 @@ use futures::channel::mpsc;
 use futures::StreamExt;
 use mc_commitment_state_diff::BlockDAData;
 use mp_hashers::HasherT;
-use pallet_starknet_runtime_api::StarknetRuntimeApi;
-use sc_client_api::client::BlockchainEvents;
 use serde::{Deserialize, Serialize};
-use sp_api::ProvideRuntimeApi;
-use sp_blockchain::HeaderBackend;
-use sp_runtime::traits::{Block as BlockT, Header};
+use sp_runtime::traits::Block as BlockT;
+use starknet_api::block::BlockHash;
 use utils::state_diff_to_calldata;
 
-pub struct DataAvailabilityWorker<B, C, H>(PhantomData<(B, C, H)>);
+pub struct DataAvailabilityWorker<B, H>(PhantomData<(B, H)>);
 pub struct DataAvailabilityWorkerProving<B>(PhantomData<B>);
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


crates/client/data-availability/src/lib.rs
```diff
@@ -77,61 +74,53 @@ impl<B> DataAvailabilityWorkerProving<B>
...
+        while let Some(BlockDAData(starknet_block_hash, csd, num_addr_accessed)) = state_diffs_rx.next().await {
+            log::info!(
+                "received state diff for block {starknet_block_hash}: {csd:?}. {num_addr_accessed} addresses accessed."
+            );
 
-        while let Some(notification) = notification_st.next().await {
+            // store the state diff
+            if let Err(db_err) = madara_backend
+                .da()
+                .store_state_diff(&starknet_block_hash, state_diff_to_calldata(csd, num_addr_accessed))
```
tdelabro, line 120: The value returned by `state_diff_to_calldata(csd, num_addr_accessed)` is read again from db on line 150: 
`DaMode::Sovereign => match madara_backend.da().state_diff(&starknet_block_hash)`

It feels like a waste. It can be kept in memory, in a variable an used directly from there.

crates/client/data-availability/src/lib.rs
```diff
@@ -77,61 +74,53 @@ impl<B> DataAvailabilityWorkerProving<B>
...
+            {
+                log::error!("db err: {db_err}");
+            };
             // Query last written state
             // TODO: this value will be used to ensure the correct state diff is being written in Validity mode
             let _last_published_state = match da_client.last_published_state().await {
```
tdelabro, line 126: If it's only used in validity mode, it can probably be moved inside the corresponding match arm

crates/client/data-availability/src/lib.rs
tdelabro: Does it even need to be stored?
It is the only place this is used. Nobody else is reading it

tdelabro (changes requested):

crates/client/data-availability/src/lib.rs
apoorvsadana: It will be read again during proving. We can however pass it directly as well as now proving is called from `update_state`. However, I am not sure how safe just in memory would be during node restarts or crashes.

tdelabro: Will the `prove` task need the encoded stated diff (the payload expected by l1 contract) or the `ThinkStateDiff`?
We are currently storing the former, but I think the latter is what we need.

apoorvsadana: I am not sure, maybe @drspacemn can comment on this. Either ways, I think it will be easy to switch to the latter if needed later on if we are sure we need to store it right now.

tdelabro (approved):

PR #1343, opened 2024-01-03
fix: reorder DA tasks

tdelabro:

Publish state then prove

Resolves: #NA

prove then publish state, as in validity mode the eth core contracts update_state check for the proof


apoorvsadana (approved):

apoorvsadana:
lgtm

Issue #1334, opened 2023-12-28
feat: fallback default file for DA and Settlement configuration files 


tdelabro:
# Feature Request

**Describe the Feature Request**

In `crates/node/src/commands/run.rs`:
```rust
let da_conf: PathBuf = cli.run.da_conf.expect(msg: "clap requires da_conf when da_layer is present");
```
If `cli.run.da_conf` is `None` we panic. We could instead, go look into the data directory, where the setup command writes stuff, for any file named `da_conf.<json/toml>`. If we find one, we deserialize it and use it as conf.
Same for `cli.settlement_conf`

Tidus91:
Hey, i could do this one

PR #1344, opened 2024-01-03
feat: fallback default file for DA and Settlement configuration files #1334

Tidus91:
# Pull Request type

- Feature

No


crates/node/src/commands/run.rs
```diff
@@ -127,13 +129,43 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
...
+                            (Some(res),None)
+                        },
+                        (false,true) => {
+                            let file = File::open(path_da_conf_toml).unwrap();
+                            let reader = BufReader::new(file);
+                            let res : PathBuf = serde_json::from_reader(reader).unwrap();
+                            (None,Some(res))
+                        },
+                        _ => {
+                            log::info!("does not contain DA config");
```
tdelabro, line 156: No need to log if you return an error

crates/node/src/commands/run.rs
```diff
@@ -127,13 +129,43 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
...
+                        (true,false) => {
+                            let file = File::open(path_da_conf_json).unwrap();
+                            let reader = BufReader::new(file);
+                            let res : PathBuf = serde_json::from_reader(reader).unwrap();
+                            (Some(res),None)
+                        },
+                        (false,true) => {
+                            let file = File::open(path_da_conf_toml).unwrap();
+                            let reader = BufReader::new(file);
+                            let res : PathBuf = serde_json::from_reader(reader).unwrap();
```
tdelabro, line 152: Don't use serde_json for toml, use https://docs.rs/toml/latest/toml/

crates/node/src/commands/run.rs
```diff
@@ -146,14 +178,45 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
...
+                                "Ethereum config not found in fn run_node"
+                            )));
+                        }
+                    };
+                    match (test_path_sett_conf_json,test_path_sett_conf_toml) {
+                        (Some(x),None) => Some(x),
+                        (None,Some(x)) => Some(x),
+                        _ => None
+                    }
+                }
```
tdelabro, line 216: Refactor by creating a generic function that will work for both cases rather than copy pasting

tdelabro (changes requested):
Edit CHANGELOG

crates/node/src/commands/run.rs
```diff
@@ -2,7 +2,13 @@ use std::path::PathBuf;
...
 use clap::ValueHint::FilePath;
 use madara_runtime::SealingMode;
-use mc_data_availability::DaLayer;
+use mc_data_availability::avail::config::AvailConfig;
+use mc_data_availability::avail::AvailClient;
+use mc_data_availability::celestia::config::CelestiaConfig;
+use mc_data_availability::celestia::CelestiaClient;
+use mc_data_availability::ethereum::config::EthereumConfig;
+use mc_data_availability::ethereum::EthereumClient;
+use mc_data_availability::{DaClient, DaLayer};
```
tdelabro, line 11: When you rebase you will have to add `features = ["avai", "celestia"]` to your import of `mc_data_availability`

crates/node/src/commands/run.rs
```diff
@@ -119,25 +125,47 @@ impl ExtendedRunCmd {
...
-            let da_conf = cli.run.da_conf.expect("clap requires da_conf when da_layer is present");
-            if !da_conf.exists() {
-                log::info!("{} does not contain DA config", da_conf.display());
-                return Err("DA config not available".into());
-            }
-
-            Some((da_layer, da_conf))
+            let da_conf = cli.run.clone().da_conf.unwrap_or({
+                let path_base_path = cli.run.base_path()?;
+                let path_da_conf_json = path_base_path.path().join("chains/dev/ethereum.json");
```
tdelabro, line 157: `let path_da_conf_json = path_base_path.path().join("chains/dev/ethereum.json");`

This will only work if `da_layer ==  DaLayer::Ethereum`, but you also have to handle `avail` and `celestia`

crates/node/src/commands/run.rs
```diff
@@ -119,25 +125,47 @@ impl ExtendedRunCmd {
...
+                if !path_da_conf_json.exists() {
+                    return Err(sc_cli::Error::Input("no file ethereum.json in base_path".to_string()));
```
tdelabro, line 159: change error message not to be `etherem`

crates/node/src/commands/run.rs
tdelabro: I suggest you `impl Display for DaLayer` and then do:
 ```rust
 let path_da_conf_json = path_base_path.path().join("chains/dev").join(format!("{da_layer}.json");
 ```

tdelabro (changes requested):

crates/node/src/commands/run.rs
```diff
@@ -119,25 +125,47 @@ impl ExtendedRunCmd {
...
+}
+
 pub fn run_node(mut cli: Cli) -> Result<()> {
     if cli.run.base.shared_params.dev {
         override_dev_environment(&mut cli.run);
     }
     let runner = cli.create_runner(&cli.run.base)?;
 
-    let da_config: Option<(DaLayer, PathBuf)> = match cli.run.da_layer {
+    let (da_config, _da_client) = match cli.run.da_layer {
```
tdelabro, line 153: `_da_client` should be used and passed as an argument to `service::new_full`.
The content of `init_da_client` is duplicated in `service.rs` where it should be removed.

tdelabro (changes requested):

crates/node/src/commands/run.rs
```diff
@@ -119,25 +125,49 @@ impl ExtendedRunCmd {
...
-            let da_conf = cli.run.da_conf.expect("clap requires da_conf when da_layer is present");
-            if !da_conf.exists() {
-                log::info!("{} does not contain DA config", da_conf.display());
-                return Err("DA config not available".into());
-            }
-
-            Some((da_layer, da_conf))
+            let da_conf = cli.run.clone().da_conf.unwrap_or({
+                let path_base_path = cli.run.base_path()?;
+                let path_da_conf_json = path_base_path.path().join("chains/dev").join(format!("{da_layer}.json"));
```
tdelabro, line 159: `.join("chains/dev")`

it's `chains/dev` only if you pass `--chain=dev` or `--dev` but you can give any name. So you have to make it dynamic here

crates/node/src/commands/run.rs
```diff
@@ -146,15 +176,18 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
...
-            let settlement_conf = cli.run.settlement_conf.expect("clap requires da_conf when settlement is present");
-            if !settlement_conf.exists() {
-                return Err(sc_cli::Error::Input(format!(
-                    "Ethereum config not found at {}",
-                    settlement_conf.display()
-                )));
-            }
+            let settlement_conf = cli.run.clone().settlement_conf.unwrap_or({
+                let path_base_path = cli.run.base_path()?;
+                let path_sett_conf_json = path_base_path.path().join("settlement_conf.json");
```
tdelabro, line 181: missing the `chain/<chain_name>` part of the path

crates/node/src/service.rs
```diff
@@ -423,41 +423,26 @@ pub fn new_full(
     let (commitment_state_diff_tx, commitment_state_diff_rx) = mpsc::channel(5);
 
     // initialize data availability worker
-    if let Some((da_layer, da_path)) = da_layer {
+    if let Some((_da_layer, _da_path)) = da_layer {
```
tdelabro, line 426: The fact they are not used it the signal that they should not be passed as arguments to the function.

I think you can remove this arg: `da_layer: Option<(DaLayer, PathBuf)>`
And just check `if let Some(da_client) = da_client` in order to know if we should run a da client

tdelabro (changes requested):

crates/node/src/service.rs
Tidus91: if i remove da_layer from the function i guess i'll have to do in run.rs this part :
```
task_manager.spawn_essential_handle().spawn(
            "commitment-state-diff",
            Some("madara"),
            CommitmentStateDiffWorker::<_, _, StarknetHasher>::new(client.clone(), commitment_state_diff_tx)
                .for_each(|()| future::ready(())),
        );
```
        
but i won't have access to all the context like the taskManager, commitment_state_diff etc..., what do you advice ?

Tidus91: Or maybe do you want me to run `service::new_full` only if there is a da_layer ?

tdelabro (approved):

PR #1346, opened 2024-01-04
feat: add sierra to casm class hash mapping in genesis assets

tdelabro:


LucasLvy (approved):

crates/client/data-availability/src/lib.rs
```diff
@@ -63,6 +64,16 @@ pub enum DaMode {
     Sovereign,
 }
 
+impl Display for DaMode {
```
EvolveArt, line 67: we can probably just use StringEnum from strum/serde ?

EvolveArt (approved):

crates/client/data-availability/src/lib.rs
tdelabro: What is StringEnum?

PR #1356, opened 2024-01-08
Braavos fix

apoorvsadana:
Right now `is_query` is overloaded where if `is_query` is true it means

=> the transaction version is SIMULATION_OFFSET + VERSION
=> the transaction is an estimate_fee transaction

However, Starknet supports estimate fee without simulation offset which is what braavos has been doing. So this PR decouples the too and creates 

1. offset_version => use simulation offset
2. is_query => is estimate_fee


lana-shanghai (approved):
lgtm

PR #1359, opened 2024-01-08
support native execution

apoorvsadana:


tdelabro:
This reverts to a previous version of substrate. We don't want to have it merged into main.

But we can keep it here in a branch and rebase from time to time.

The solution we want is to have the current substrate run in Native, probably through some changes in the substrate framework

tdelabro:
replaced by #1407

tdelabro:
replaced by https://github.com/keep-starknet-strange/madara/pull/1407

PR #1360, opened 2024-01-09
add feature flags for DA

apoorvsadana:


EvolveArt:
@apoorvsadana Should we add steps in the CI to build/release these new images ?

apoorvsadana:
You mean just having multiple images during the release or do we also need to build with these features for some CI? Right now the tests are passing so I guess there is nothing in the CI which needs avail/celestia as of now.

EvolveArt:
> You mean just having multiple images during the release or do we also need to build with these features for some CI? Right now the tests are passing so I guess there is nothing in the CI which needs avail/celestia as of now.

Yes I meant having multiple released images, we will have integration tests for DA layer which will use these feature flags too but they are not fully ready yet.

EvolveArt (approved):

apoorvsadana:
I will have to check the release workflow but do we need that right now? Maybe we can create an issue for that for someone to pick it up if that works?

tdelabro (approved):

Issue #1335, opened 2023-12-28
dev: refactor `LAST_SYNCED_L1_BLOCK` to be more clear


apoorvsadana:
Right now, Madara fetches the event `LogMessageToL2` from Ethereum and accordingly creates a tx on the chain. Every time we get an event, we process it and update `LAST_SYNCED_L1_BLOCK` to the block number of the latest event. The name `LAST_SYNCED_L1_BLOCK` however is not accurate. The issue with this is that if there are no events from block n to block n+t, then `LAST_SYNCED_L1_BLOCK` won't be updated. So we want to refactor this variable to actually be a pair of

1. Block number
2. Event index

Where block_number represents the last_synced_l1_block_with_event and the event index represents the index of the event in the block.

okhaimie-dev:
I'd love to try this out.

dangush:
Hi, I'll pick this up!

MSghais:
Hi, can I pick this one please?

I'm new to this repo, I would love to have some information on what is the best way to modify it, please.
I think we have few solutions: 

1.  Add others variable on top of it: 

- Add others columns in the DB columns, 
- add static_keys to get the LAST_L1_EVENT_INDEX
- Save also this variable on the DB 
- Do check in the L1_MESSAGES worker.rs to also update if we have others event index or not  in this block.

3. Modify struct of backend.messaging(): 

    a. Modify the MESSAGING columns to be an Hash in :
    https://github.com/keep-starknet-strange/madara/blob/bd1898aec7004ea08b5fedaf515012d44a29a14f/crates/client/db/src/lib.rs#L64
    
    b. Modify the messaging for GET and UPDATE functions of LAST_SYNCED_L1_BLOCK here:
    
    https://github.com/keep-starknet-strange/madara/blob/bd1898aec7004ea08b5fedaf515012d44a29a14f/crates/client/db/src/messaging_db.rs#L17

  c. Update the Client for L1_MESSAGE worker accordingly: 
https://github.com/keep-starknet-strange/madara/blob/bd1898aec7004ea08b5fedaf515012d44a29a14f/crates/client/l1-messages/src/worker.rs#L34

dangush:
Hi, reposting some questions sent to TG (my bad, should have posted here)

I've opened a PR with what seem as the relevant changes, but I would like to develop some tests for this part of the code to make sure. I'm trying to orient myself around the testing setup, and would love some pointers on this. 

[...]

PR #1361, opened 2024-01-09
Refactoring: Rename LAST_SYNCED_L1_BLOCK to be more clear

dangush:
# Refactor

Closes Issue #1335 by refactoring LAST_SYNCED_L1_BLOCK to a struct named LastSyncedEventBlock which contains a block_number field and event_index field. 

When `process_l1_message` is called, the L1 block number is written to the MessagingDb. 

Resolves: #1335 

## What is the new behavior?

When `process_l1_message` is called, a struct containing both the L1 block number and event_index is written to the MessagingDb. 

I'm not yet sure if/how the MessagingDb schema needs to be changed to accept the encoded Struct instead of previously-encoded u64 value. Accorrding to the [SCALE encoding docs](https://docs.substrate.io/reference/scale-codec/) I believe structs get encoded as a vector of bytes, so it seems the difference is that a value twice the current size will need to be written to the DB.

## Todo

I would like to write tests for this part of the code, and have asked a few questions in the issue thread asking how to approach this.


apoorvsadana (approved):
lgtm

apoorvsadana:
@tdelabro can you review this once?

crates/client/db/src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub use mapping_db::MappingCommitment;
 use starknet_api::hash::StarkHash;
 mod da_db;
 mod db_opening_utils;
-mod messaging_db;
+pub mod messaging_db;
```
tdelabro, line 22: We don't want any of the db `mod` to be public. They should only be accessible through our `Backend` struct

tdelabro: I would reexport `LastSyncedEventBlock` struct in `lib.rs`

tdelabro (changes requested):

crates/client/db/src/messaging_db.rs
```diff
@@ -14,18 +14,37 @@ pub struct MessagingDb<B: BlockT> {
...
+        LastSyncedEventBlock { block_number, event_index }
+    }
+}
+
 impl<B: BlockT> MessagingDb<B> {
-    pub fn last_synced_l1_block(&self) -> Result<u64, DbError> {
-        match self.db.get(crate::columns::MESSAGING, crate::static_keys::LAST_SYNCED_L1_BLOCK) {
-            Some(raw) => Ok(u64::decode(&mut &raw[..])?),
-            None => Ok(0),
+    pub fn last_synced_l1_block_with_event(&self) -> Result<LastSyncedEventBlock, DbError> {
```
tdelabro, line 30: `last_synced_l1_block_and_event`

tdelabro (approved):

PR #1363, opened 2024-01-10
fix eth da for sovereign mode, add header to calldata

apoorvsadana:
Fixes

1. When converting state_diff to calldata, we don't add the headers to the calldata because of which update_state fails with `STARKNET_OUTPUT_TOO_SHORT`. This PR adds the required headers but we still need to implement proving for update_state to work.
2. Eth Da should work in sovereign mode when we are using the zaun contracts. However, current implementation assumes the core contracts are the ones used for Validity mode (which has two extra parameters in update_state). So now, the code checks DaMode and accordingly calls the update_state method.


EvolveArt (approved):

crates/client/data-availability/src/lib.rs
```diff
@@ -200,14 +203,12 @@ pub async fn prove<B: BlockT>(
...
-    starknet_block_hash: BlockHash,
-    csd: ThinStateDiff,
-    num_addr_accessed: usize,
+    block_da_data: BlockDAData,
 ) -> Result<(), anyhow::Error> {
     // store the state diff
     madara_backend
         .da()
-        .store_state_diff(&starknet_block_hash, state_diff_to_calldata(csd, num_addr_accessed))
+        .store_state_diff(&block_da_data.block_hash, state_diff_to_calldata(block_da_data.clone()))
```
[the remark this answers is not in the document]
apoorvsadana: fixed

crates/pallets/starknet/src/lib.rs
```diff
@@ -1229,6 +1231,14 @@ impl<T: Config> Pallet<T> {
         T::ProgramHash::get()
     }
 
+    pub fn config_hash() -> StarkHash {
+        pedersen_hash_array(&[
```
EvolveArt, line 1235: shouldn't we use the pallet's hash ?

EvolveArt (approved):

crates/pallets/starknet/src/lib.rs
apoorvsadana: fixed

PR #1365, opened 2024-01-10
chore: remove primitives/commitments

tdelabro:
# Pull Request type

- `Default` value for globabl_state_root in StarknetBlockHeader
- event and tx commitments computed in the runtime

## What is the new behavior?

- remove those 3 fields
- add a way to get a `Default` value for global_state_root 
- event and tx commitments are not created anymore (they weren't used apparently)

Yes. The wrapped block header has changed.

## Other information

The reason of those changes is to prepare our codebase to the computation of all commitments client-side using the starklnet-bonzai-trie


crates/primitives/genesis-config/Cargo.toml
```diff
@@ -17,7 +17,7 @@ blockifier = { workspace = true, features = [
 ] }
 #
 # third party
-derive_more = { workspace = true }
+derive_more = { workspace = true, features = ["constructor"] }
```
EvolveArt, line 20: what's that used for ?

EvolveArt (approved):

crates/primitives/genesis-config/Cargo.toml
tdelabro: I got an error when compiling the crate. 

It was used here:
```
#[derive(Constructor)]
pub struct GenesisLoader {
    base_path: PathBuf,
    data: GenesisData,
}
```

But the feature was not present

PR #1371, opened 2024-01-12
refacto(simulate): move type conversion logic into client

tdelabro:

We duplicate all starknet-rs types involved in the simulate rpc logic.
We do all the type conversion in the runtime.

## What is the new behavior?

We use starknet-rs types in the RPC interface
We do the tx execution in the runtime and the types conversion in the client


antiyro:
This refacto makes sense as it is needed for further trace api implementations. Seems mostly good. Thanks.

crates/client/rpc/src/lib.rs
```diff
@@ -1501,7 +1499,17 @@ where
                 messages_sent: messages.into_iter().map(message_conversion).collect(),
                 events: events_converted,
                 execution_result,
-                execution_resources: execution_resources.into(),
+                execution_resources: ExecutionResources {
```
apoorvsadana, line 1502: what was the issue with using the default value? also, why not have one variable and reuse that for easier readability.

apoorvsadana (approved):

crates/client/rpc/src/lib.rs
tdelabro: Default is not implemented on `starknet_core::types::ExecutionResources`

Issue #1373, opened 2024-01-16
bug(rpc): getNonce unexisting contract not detected


antiyro:
Currently if I get a nonce on a non existing contract it returns `0x0`

to reproduce:

```
curl --request POST \
     --url http://localhost:9944 \
     --header 'accept: application/json' \
     --header 'content-type: application/json' \
     --data '{
       "id": 1,
       "jsonrpc": "2.0",
       "method": "starknet_getNonce",
       "params": [
         {"block_number": 2},
         "0x0"
       ]
     }'
```

result

```
{"jsonrpc":"2.0","result":"0x0","id":1}
```

Tbelleng:
Can i take this one sir ?

Tbelleng:
the problem seems to come from this part : https://github.com/keep-starknet-strange/madara/blob/8c285b45101dd20736233408f37d91ab87a995cd/crates/client/storage/src/overrides/schema_v1_override.rs#L126C1-L131C2

In this part, the Nonce is set-up to 0 (so a valid Nonce) even if its do not exist.

Tbelleng:
Fixed by #1374 ✅

PR #1374, opened 2024-01-16
feat: :white_check_mark: Adding test for getNonce RPC Call if Contrac…

Tbelleng:
# Description

Solve issue : #1373 

Adding a test for getNonce RPC call in case the contract doesnt exist/ is invalid. 
I Also changed the "nonce" function in order to return None instead of a Nonce::Default().
The precedent version of Madara, when a invalid contract was tested, didnt produce any error and was
just return a default Nonce, equal to 0, which is a valid nonce on Starknet. 
By this return, the ContractError was never triggered, resulting with a valid nonce return.

# Pull Request type

- Bugfix
- Testing

## What is the new behavior?

By changing the nonce return in case of error and adding a test in case of an invalid contract,
its return the right error.

No

### DIFF
--- a/starknet-rpc-test/get_block_with_txs.rs
+++ b/starknet-rpc-test/get_block_with_txs.rs
@@ -101,15 +101,13 @@ async fn works_with_deploy_account_txn(madara: &ThreadSafeMadaraClient) -> Resul
     let contract_address_salt = FieldElement::ONE;
     let max_fee = FieldElement::from_hex_be(MAX_FEE_OVERRIDE).unwrap();
 
-    let (block) = {
+    let block = {
         let mut madara_write_lock = madara.write().await;
         let oz_factory = build_oz_account_factory(&rpc, "0x123", class_hash).await;
         let account_deploy_txn = build_deploy_account_tx(&oz_factory, FieldElement::ONE);
 
         let funding_account = build_single_owner_account(&rpc, SIGNER_PRIVATE, ARGENT_CONTRACT_ADDRESS, true);
         let account_address = account_deploy_txn.address();
-        // let deploy_nonce = rpc.get_nonce(BlockId::Tag(BlockTag::Latest),
-        // account_deploy_txn.address()).await?;
 
         // We execute the funding in a different block, because we have no way to guarantee the tx execution
         // order once in the mempool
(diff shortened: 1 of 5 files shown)


Tbelleng:
Hello @tdelabro 
Its seems that turning nonce to None instead of default (which turn nonce into a 0 which is a valid nonce) invalid 4 tests into madara : 
- fail_execution_step_with_no_storage_change
- fail_validation_step
- fails_already_declared
- works_with_storage_change

So 3 of these test fail earlier than expected by madara because of the nonce = None.
Is it a normal behavior or those tests need to be fixed ?

tdelabro:
> Hello @tdelabro Its seems that turning nonce to None instead of default (which turn nonce into a 0 which is a valid nonce) invalid 4 tests into madara :
> 
>     * fail_execution_step_with_no_storage_change
> 
>     * fail_validation_step
> 
>     * fails_already_declared
> 
>     * works_with_storage_change
> 
> 
> So 3 of these test fail earlier than expected by madara because of the nonce = None. Is it a normal behavior or those tests need to be fixed ?

I'm we will need you to fix those tests. And I'm unsure why they fail. But I will be happy to help you if you need

Tbelleng:
thanks @tdelabro !
By doing differently i fixed those tests but i got a failure at this one : works_with_deploy_account_txn
My implmentation trigger an error when the "get_nonce" function is used.
I guess its might come from the fact that within this test, madara use to set-up a valid nonce at 0, because the contract doesnt have one.
But with the new implementation, if a contract doesnt have any nonce (not even 0), its trigger the ContractNotFound error (because a contract with no nonce cant exist), so the test that was working before, fail with the new implementation.

Tbelleng:
Tests passed ✅ The one that fail now is : work_ok_non_used_contract_address because this test must fail and its fail with the new implementation, but for an unknow reason the test, in the current madara version, is considered as valid.
My suggestion is to remove this test, or to modify it to expect a fail instead of an Ok().

starknet-rpc-test/get_block_with_txs.rs
```diff
@@ -101,14 +101,15 @@ async fn works_with_deploy_account_txn(madara: &ThreadSafeMadaraClient) -> Resul
...
+    let (block) = {
         let mut madara_write_lock = madara.write().await;
         let oz_factory = build_oz_account_factory(&rpc, "0x123", class_hash).await;
         let account_deploy_txn = build_deploy_account_tx(&oz_factory, FieldElement::ONE);
 
         let funding_account = build_single_owner_account(&rpc, SIGNER_PRIVATE, ARGENT_CONTRACT_ADDRESS, true);
         let account_address = account_deploy_txn.address();
-        let deploy_nonce = rpc.get_nonce(BlockId::Tag(BlockTag::Latest), account_deploy_txn.address()).await?;
+        // let deploy_nonce = rpc.get_nonce(BlockId::Tag(BlockTag::Latest),
+        // account_deploy_txn.address()).await?;
```
tdelabro, line 112: Remove rather than comment

tdelabro (changes requested):

tdelabro (approved):

PR #1377, opened 2024-01-17
add MVP EigenDA support 

mbtariq82:
## Pull Request type
Feature

## What is the new behavior?
Publishes the state diff to EigenDA through the disperser on testnet and then queries the disperser for the blob status every 5 seconds until confirmed.

No


crates/client/data-availability/src/eigenda/blob.rs
```diff
@@ -0,0 +1,182 @@
...
+impl Default for DisperseBlobResponse {
+    fn default() -> Self {
+        DisperseBlobResponse { status: Default::default(), request_id: Default::default() }
+    }
+}
+impl From<String> for DisperseBlobResponse {
+    fn from(value: String) -> Self {
+        if let Some(start_index) = value.find('{') {
+            let json_str = &value[start_index..];
+            println!("{}", &json_str);
```
apoorvsadana, line 28: we can remove the println!

crates/client/data-availability/src/eigenda/blob.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+}
+impl Default for DisperseBlobResponse {
+    fn default() -> Self {
+        DisperseBlobResponse { status: Default::default(), request_id: Default::default() }
+    }
+}
+impl From<String> for DisperseBlobResponse {
+    fn from(value: String) -> Self {
+        if let Some(start_index) = value.find('{') {
```
apoorvsadana, line 26: why do we need the find? does it not start with "{" all the time?

crates/client/data-availability/src/eigenda/blob.rs
```diff
@@ -0,0 +1,182 @@
...
+                Ok(response) => {
+                    return response
+                }
+                Err(err) => {
+                    println!("{}", &err);
+                    return DisperseBlobResponse::default() 
+                }
+            }
+        } else {
+            return DisperseBlobResponse::default()
```
apoorvsadana, line 42: instead of returning default should we do `TryFrom` and throw an error?

crates/client/data-availability/src/eigenda/blob.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "UPPERCASE")]
+pub enum BlobStatus {
+    Processing,
+    Confirmed,
+    Failed,
+    Other(String)
+}
+impl Default for BlobStatus {
```
apoorvsadana, line 55: what's this needed for?

crates/client/data-availability/src/eigenda/blob.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+impl BlobStatusResponse {
+    pub fn status(&self) -> &BlobStatus {
+        &self.status
+    }
+    pub fn info(&self) -> &BlobInfo {
+        &self.info
+    }
+}
+impl Default for BlobStatusResponse {
```
apoorvsadana, line 74: what's the Default needed for?

crates/client/data-availability/src/eigenda/blob.rs
```diff
@@ -0,0 +1,182 @@
...
+            match blob_status {
+                Ok(status) => {
+                    return status
+                }
+                Err(_) => {
+                    return BlobStatusResponse::default()
+                }
+            }
+        } else {
+            return BlobStatusResponse::default()
```
apoorvsadana, line 96: let's use `TryFrom`

crates/client/data-availability/src/eigenda/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+            return Ok(response);
+        } else {
+            let error_message = String::from_utf8(output.stderr).map_err(|err| {
+                std::io::Error::new(std::io::ErrorKind::Other, err.to_string())
+            })?;
+            return Err(std::io::Error::new(std::io::ErrorKind::Other, error_message))
+        }
+    }
+
+    fn create_disperse_blob_payload(&self, state_diff: Vec<U256>) -> String {
```
apoorvsadana, line 103: Let's create a rust Struct and serialize that using serde, this format would be difficult to edit later on

crates/client/data-availability/src/eigenda/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        let payload = self.create_disperse_blob_payload(state_diff);
+        let output = grpcurl_command!(
+            "-proto", &self.config.proto_path,
+            "-d", &payload,
+            &self.config.grpc_provider,
+            "disperser.Disperser/DisperseBlob"
+        )?;
+        if output.status.success() {
+            let disperse_blob_response: DisperseBlobResponse = String::from_utf8(output.stdout).map_err(|err| {
+                std::io::Error::new(std::io::ErrorKind::Other, err.to_string())
```
apoorvsadana, line 71: we can use anyhow::Result and throw the actual error instead of throwing an io error here

crates/client/data-availability/src/eigenda/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        payload.push_str(&self.config.quorum_id.to_string());
+        payload.push_str(r#","adversary_threshold":"#);
+        payload.push_str(&self.config.adversary_threshold.to_string());
+        payload.push_str(r#","quorum_threshold":"#);
+        payload.push_str(&self.config.quorum_threshold.to_string());
+        payload.push_str(r#"}]}"#);
+        payload
+    }
+
+    fn create_get_blob_status_payload(request_id: String) -> String {
```
apoorvsadana, line 117: same as above

crates/client/data-availability/src/eigenda/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        let payload = Self::create_get_blob_status_payload(request_id);
+        let output = grpcurl_command!(
+            "-proto", &self.config.proto_path,
+            "-d", &payload,
+            &self.config.grpc_provider,
+            "disperser.Disperser/GetBlobStatus"
+        )?;
+        if output.status.success() {
+            let response: BlobStatusResponse = String::from_utf8(output.stdout).map_err(|err| {
+                std::io::Error::new(std::io::ErrorKind::Other, err.to_string())
```
apoorvsadana, line 92: same as above

Issue #1331, opened 2023-12-25
feat(rpc): Compute state diff fields in get_state_update/simulate_transaction


antiyro:
### feat(rpc): Compute state diff fields in get_state_update/simulate_transaction

Currently they're 2 implementations of `StateDiff` within madara:

1. One used from [starknet.rs](https://github.com/xJonathanLEI/starknet-rs/blob/6cadb198644cf5e942187896c2e6bb3ecc9edc6e/starknet-core/src/types/codegen.rs#L1684) defaulted in [get_state_update](https://github.com/keep-starknet-strange/madara/blob/2079af1275f09e16d6fa5f441309e1fedfb59d3c/crates/client/rpc/src/lib.rs#L1164)
2. A copy of the Starknet rs implementation supporting encode/decode parity scale codec trait supported by madara

The goal of this issue would be to assign everywhere the seconde implementation of state diff field (located under mp-state still to be merged in #1313) and compute it correctly in `get_state_update`/`simulate_transaction` rpc based on v0.5.1 rpc specs as defined [here](https://github.com/starkware-libs/starknet-specs/blob/v0.5.1/api/starknet_api_openrpc.json)

antiyro:
@EvolveArt are you taking this one?

PR #1378, opened 2024-01-17
✨ get_state_update

EvolveArt:

Resolves: #1331 

## What is the new behavior?

Reads the state diff from the DA db. Decodes it into a `starknet::core::types::StateDiff` struct following 0.11 format and returns it within the `get_state_update` RPC call.

No.


crates/client/rpc/src/lib.rs
```diff
@@ -1149,18 +1172,18 @@ where
...
             old_root,
-            state_diff: StateDiff {
-                storage_diffs: Vec::new(),
-                deprecated_declared_classes: Vec::new(),
-                declared_classes: Vec::new(),
-                deployed_contracts: Vec::new(),
-                replaced_classes: Vec::new(),
-                nonces: Vec::new(),
-            },
+            state_diff,
```
antiyro, line 1186: could you add this field to simulate tx where it is defaulted to none https://github.com/keep-starknet-strange/madara/blob/main/crates/client/rpc/src/trace_api.rs#L257

EvolveArt: This is a different type of state diff, just the one induced bt the given transaction. Should be dealt with in a different way. 

Best approach is to have a different type of `BlockifierStateAdapter` that stores intermediary changes imo.

crates/client/rpc/src/lib.rs
```diff
@@ -195,6 +199,25 @@ where
             None
         })
     }
+
+    /// Returns the state diff for the given block.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_hash` - The hash of the block containing the state diff (starknet block).
+    fn get_state_diff(&self, block_hash: B::Hash, starknet_block_hash: &BlockHash) -> Result<StateDiff, String> {
```
apoorvsadana, line 208: There are only 2 places we read `state_diff` right now. One is here and the other is when doing the state update. The conversion from `ThinStateDiff` to `Vec<U256>` also happens during state update so we already have the vector form available in memory over there. So why don't we just store actual state diff which we need to read here, that will save us the decoding part?

Moreover, if the update_state part would anyways move to a new service, then this is the only place where we read it right?

EvolveArt: Unfortunately, we can't truly rely on this. I think ideally, we'd want to have a fallback mechanism that fetches the state diff from L1 for instance. We might also have concurrency issues with this approach. 

Your idea is to directly store the `starknet_rs::StateDiff` in the DA db ? And completely get rid of `ThinStateDiff` ?

apoorvsadana: You mean we can't rely on an external service doing DA or we can't rely on storing state diffs directly?

I mean we can get rid of the decode function if we just store `StateDiff` directly. There's no need to encode it, store and then decode it. The only place I see the encoded value being used is during update state and that's also where we do the encoding, so its already in memory.

apoorvsadana: We could store `ThinStateDiff` or `StateDiff` whatever is convenient.

EvolveArt: Unfortunately, neither of them is SCALE encodable, what's the best approach there ? I think it ends up the same overhead as encoding/decoding to Vec<U256> no?

tdelabro: And here we decode the whole thing when we only need a subpart of it

tdelabro: SCALE could easily be impl on starknet_api ThinStateDiff and StateDiff, they both use IndexMap now

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
tdelabro: What about the part where we store the whole `BlockDAData` instead of just the state diff. It does not sound like a huge issue but wdyt?

crates/client/data-availability/Cargo.toml
```diff
@@ -43,13 +43,15 @@ blockifier = { workspace = true, default-features = true }
 mc-commitment-state-diff = { workspace = true, default-features = true }
 mc-db = { workspace = true, default-features = true }
 pallet-starknet-runtime-api = { workspace = true, features = ["std"] }
+starknet-core = { workspace = true }
```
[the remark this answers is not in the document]
EvolveArt: not needed afaik

crates/client/data-availability/src/utils.rs
```diff
@@ -1,11 +1,23 @@
...
 use starknet_api::hash::StarkFelt;
+use starknet_core::types::{
+    ContractStorageDiffItem, DeclaredClassItem, DeployedContractItem, FieldElement, NonceUpdate, ReplacedClassItem,
+    StateDiff, StorageEntry,
+};
 use url::{ParseError, Url};
 
 const CLASS_FLAG_TRUE: &str = "0x100000000000000000000000000000001"; // 2 ^ 128 + 1
-const NONCE_BASE: &str = "0x10000000000000000"; // 2 ^ 64
+const NONCE_BASE: &str = "0xFFFFFFFFFFFFFFFF"; // 2 ^ 64 - 1
```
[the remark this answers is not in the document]
EvolveArt: yup hence why I added some unit tests for `da_word`. Have been looking for a python implementation to add more tests cases but couldnt find yet..

crates/client/rpc/src/lib.rs
EvolveArt: So what is happening right now is:
ThinStateDiff -> Vec<U256> -> SCALE encode -> SCALE decode -> StateDiff

What you think would be better is:
ThinStateDiff -> SCALE encode -> SCALE decode -> Convert to StateDiff

What I'm saying is that if we can avoid this conversion function between `ThinStateDIff` and `StateDiff` that would be better and this `Vec<U256>` intermediate representation is perfect for this.

apoorvsadana: But why is a converstion b/w `ThinStateDIff` and `StateDiff` more difficult, I think it will be more readable.

apoorvsadana: also, ya we can rename the function `state_diff_to_calldata` to `block_data_to_call_data`?

EvolveArt: > But why is a converstion b/w `ThinStateDIff` and `StateDiff` more difficult, I think it will be more readable.

Maybe more readable and even that I'm not sure but high probably less efficient if we have to loop over all the mappings and convert the types.

apoorvsadana: But we are anyways looping over all the mappings and converting, we just have an extra intermediate state.

crates/client/data-availability/Cargo.toml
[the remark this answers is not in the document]
tdelabro: It can contain some optimization. Or the dependency it has can contain some optimization with "std".
There is no reason not to have at least `features = ["std"]`

crates/client/data-availability/src/utils.rs
```diff
@@ -146,3 +159,193 @@ pub fn bytes_to_felt(raw: &[u8]) -> StarkFelt {
...
+///
+/// A `Result` indicating whether the contract is deployed or not.
+///
+/// # Example
+///
+/// ```rust
+/// // Example usage
+/// let result = contract_deployed(address, block_hash, client);
+/// ```
+#[allow(unused)]
```
tdelabro, line 213: This one is actually used.

crates/client/data-availability/src/utils.rs
```diff
@@ -146,3 +159,193 @@ pub fn bytes_to_felt(raw: &[u8]) -> StarkFelt {
...
+///
+/// A `Result` containing the decoded `StateDiff` or an `Error` if decoding fails.
+///
+/// # Example
+///
+/// ```rust
+/// // Example usage
+/// let result = decode_011_diff(&encoded_diff, block_hash, client);
+/// ```
+pub fn decode_011_diff<B, C>(encoded_diff: &[U256], block_hash: B::Hash, client: Arc<C>) -> Result<StateDiff, Error>
```
tdelabro, line 249: This whole method code is really hard to read.
Can you add a link to what is the `starknet v0.11.0 logic`?
And some comments in the impl itself.

I have no idea what is going on rn

crates/client/data-availability/src/lib.rs
```diff
@@ -206,11 +206,9 @@ pub async fn update_state<B: BlockT, H: HasherT>(
     block_da_data: BlockDAData,
 ) -> Result<(), anyhow::Error> {
     let block_hash = block_da_data.block_hash;
+    let calldata = block_data_to_calldata(block_da_data);
     // store the state diff
-    madara_backend
-        .da()
-        .store_state_diff(&block_hash, state_diff_to_calldata(block_da_data))
-        .map_err(|e| anyhow!("{e}"))?;
+    madara_backend.da().store_state_diff(&block_hash, calldata.clone()).map_err(|e| anyhow!("{e}"))?;
```
tdelabro, line 211: So we are still storing the calldata. With a method named `store_state_diff`

tdelabro: I guess we didn't decide anything clearly. But what I had in mind was.

Store stuffs in the db scale-encoded, read them, convert them to calldata only when you actually need calldata.

@apoorvsadana

tdelabro (changes requested):

crates/client/data-availability/Cargo.toml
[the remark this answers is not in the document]
EvolveArt: We should probably change that in all client crates then ?

crates/client/data-availability/src/lib.rs
EvolveArt: I personnaly don't see why it's not fine to store the state diff encoded as calldata, it's the best intermediate representation we have and consumers can just decode it into whatever struct they need.

crates/client/data-availability/src/utils.rs
```diff
@@ -146,3 +159,193 @@ pub fn bytes_to_felt(raw: &[u8]) -> StarkFelt {
...
+#[allow(unused)]
+fn contract_deployed<B, C>(address: ContractAddress, block_hash: B::Hash, client: Arc<C>) -> Result<bool, Error>
+where
+    B: BlockT,
+    C: ProvideRuntimeApi<B> + HeaderBackend<B>,
+    C::Api: StarknetRuntimeApi<B>,
+{
+    #[cfg(test)]
+    // When testing, we have only an empty client.
+    return Ok(false);
```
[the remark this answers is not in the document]
EvolveArt: Fixed

crates/client/data-availability/src/utils.rs
EvolveArt: This method basically does the opposite than the `block_data_to_calldata` method.

EvolveArt: Also added more comments

crates/client/data-availability/src/lib.rs
apoorvsadana: But the SCALE encoding is already an intermediate representation. Why do we need to create another one i.e. calldata, if whenever it is read, we have to decode it anyways? Like, if there was a use case where we were reading the stored data and encoding it into call data, I think it would make more sense. However, right now, we encode it, store it and then later fetch it and decode it. My point is, the intermediate representation isn't solving much and it makes it difficult to edit later on.

tdelabro: the calldata repr is used to push on ethereum. 
A similar logic would mean store everything else as JSON because we are going to convert it to JSON in the RPC return type. 
Doesn't make sense. We should store it as the most polyvalent repr as much as possible, which given our diversity of usage, is scale encode. As it is easily decoded to the rust types

.github/workflows/starknet-rpc-tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           ./target/release/madara setup --chain=dev --from-local=configs
       - name: Run rpc test
         run: |-
-          ./target/release/madara --dev --sealing=manual &
+          ./target/release/madara --dev --sealing=manual --da-layer=ethereum --da-conf=examples/da-confs/ethereum.json &
```
tdelabro, line 36: So we need to run the node with `--da-layer=ethereum --da-conf=examples/da-confs/ethereum.json` in order to be able to answer `getStateUpdate`. What if we don't?

tdelabro (changes requested):

crates/client/data-availability/src/lib.rs
apoorvsadana: > the calldata repr is used to push on ethereum.
Yes, but the function where we do the encoding is also the function where we push it. So, its in memory already at that point when we are pushing the encoded state to Ethereum. So the what we store is being used for other places like for the RPC. Over there, it makes more sense to read directly a Rust type as its easily readable and changeable later on. In the current form, if I want to change this I need to change the decode and encode functions as well.

apoorvsadana: > the calldata repr is used to push on ethereum.

Yes, but the function where we do the encoding is also the function where we push it. So, its in memory already at that point when we are pushing the encoded state to Ethereum. So the what we store is being used for other places like for the RPC. Over there, it makes more sense to read directly a Rust type as its easily readable and changeable later on. In the current form, if I want to change this I need to change the decode and encode functions as well.

apoorvsadana: So I agree that we should store in a polyvalent form and that SCALE already solves that for us. I am saying that if it was the case that whenever we were reading state diffs we were encoding them to make them usable, then it would make sense to just store the encoded form. For example, if there was one worker storing the diff and then there were multiple workers reading it and encoding it, then I would vote for encoding it and storing it. In the current case, I feel storing the rust type directly solves the problem better.

.github/workflows/starknet-rpc-tests.yml
EvolveArt: the `commitment-state-diff` worker is only initialized if we specify a da layer yes

crates/client/data-availability/src/lib.rs
EvolveArt: Ended up storing `ThinStateDiff` in the DA db, waiting for https://github.com/keep-starknet-strange/starknet-api/pull/15 to be merged

This way we can get rid of the `decode_v011` function but we now have a `to_rpc_state_diff` conversion util

.github/workflows/starknet-rpc-tests.yml
tdelabro: So we are not able to correctly answer the rpc.
Two solutions:
- always spawn `commitment-state-diff` worker and store them in db
- if state-diff-not in db, simulate the block in the runtime and get the state diff out of this 

@apoorvsadana @antiyro @EvolveArt wdyt?

EvolveArt: see https://github.com/keep-starknet-strange/madara/issues/1428

crates/client/rpc/src/lib.rs
```diff
@@ -194,6 +199,29 @@ where
...
+    fn get_state_diff(&self, starknet_block_hash: &BlockHash) -> Result<StateDiff, String> {
+        let state_diff = self.backend.da().state_diff(starknet_block_hash).unwrap_or_else(|err| {
+            error!("{err}");
+            ThinStateDiff {
+                deployed_contracts: IndexMap::with_capacity_and_hasher(0, Default::default()),
+                storage_diffs: IndexMap::with_capacity_and_hasher(0, Default::default()),
+                declared_classes: IndexMap::with_capacity_and_hasher(0, Default::default()),
+                deprecated_declared_classes: Vec::with_capacity(0),
+                nonces: IndexMap::with_capacity_and_hasher(0, Default::default()),
+                replaced_classes: IndexMap::with_capacity_and_hasher(0, Default::default()),
```
tdelabro, line 217: Why can't you just use with_capacity here?
And if it is for a 0 capacity, why not just use `new`?

crates/client/rpc/src/lib.rs
```diff
@@ -194,6 +199,29 @@ where
...
     }
+
+    /// Returns the state diff for the given block.
+    ///
+    /// # Arguments
+    ///
+    /// * `block_hash` - The hash of the block containing the state diff (starknet block).
+    fn get_state_diff(&self, starknet_block_hash: &BlockHash) -> Result<StateDiff, String> {
+        let state_diff = self.backend.da().state_diff(starknet_block_hash).unwrap_or_else(|err| {
+            error!("{err}");
```
tdelabro, line 210: Why are you silencing the error? You can bubble it up. `MissingStateDiff` or something along those lines

tdelabro: Let the caller decide if he want to dismiss it or not

crates/client/rpc/src/lib.rs
```diff
@@ -1226,18 +1254,18 @@ where
             FieldElement::default()
         };
 
+        let starknet_block_hash = BlockHash(starknet_block.header().hash::<H>().into());
+
+        let state_diff = self.get_state_diff(&starknet_block_hash).map_err(|e| {
+            error!("Failed to get state diff. Starknet block hash: {starknet_block_hash}, error: {e}");
```
tdelabro, line 1260: Coz here you will get a state diff, just a false one

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
EvolveArt: For alignment with what we do in `BlockifierStateAdapter` but agreed it's weird

crates/client/rpc/src/lib.rs
EvolveArt: Fixed

crates/client/rpc/src/lib.rs
EvolveArt: Fixed

starknet-rpc-test/get_state_update.rs
```diff
@@ -63,9 +63,9 @@ async fn returns_correct_state_diff_transfer(madara: &ThreadSafeMadaraClient) ->
     assert_eq!(state_update.old_root, FieldElement::ZERO);
     assert_eq!(state_update.new_root, FieldElement::ZERO);
 
-    assert_eq!(state_update.state_diff.nonces.len(), 2);
+    assert_eq!(state_update.state_diff.nonces.len(), 1);
     assert_eq!(
-        state_update.state_diff.nonces[1],
+        state_update.state_diff.nonces[0],
```
tdelabro: What changed?

tdelabro (approved):

starknet-rpc-test/get_state_update.rs
EvolveArt: Nothing just my mistake before

PR #1382, opened 2024-01-19
refacto: early exit estimate fee in runtime

### LINKED ISSUE #1315: dev: refactor estimate_fee in pallets/starknet/src/lib.rs
## Description 

Currently in `estimate_fee` we [iterate](https://github.com/keep-starknet-strange/madara/blob/main/crates/pallets/starknet/src/lib.rs#L1102) with a `for` loop over transactions, but it is better to use `.into_iter()` and `map_err` and then `collect` the results into a vector, in order to optimize the function and stop the computation if one the results errors. Similar refactoring already [takes place](https://github.com/keep-starknet-strange/madara/blob/main/crates/client/rpc/src/lib.rs#L900) in the RPC: 

```
let transactions =
    request.into_iter().map(|tx| tx.try_into()).collect::<Result<Vec<UserTransaction>, _>>().map_err(|e| {
        error!("Failed to convert BroadcastedTransaction to UserTransaction: {e}");
    })?;
```

tdelabro:
close #1315


crates/pallets/starknet/src/utils.rs
```diff
@@ -48,3 +48,50 @@ pub fn execute_txs_and_rollback<T: pallet::Config>(
...
+                    executable.execute(&mut BlockifierStateAdapter::<T>::default(), block_context, execution_config)
+                }
+            };
+
+            // In case of revert (cairo1) or failed assertion (cairo0) we stop iterating
+            // For now we don't know how to move the revert_reason out of the runtime, so we can keep things
+            // simple with the error handling
+            match result {
+                Ok(tx_info) if !tx_info.is_reverted() => execution_infos.push(tx_info),
+                _ => break,
```
EvolveArt, line 84: I don't think we should simply break but rather return early with an error in the rollback no?

EvolveArt: Also we might want to simply use the transactional [macro](https://paritytech.github.io/polkadot-sdk/master/frame_support/attr.transactional.html) wdyt ?

crates/pallets/starknet/src/utils.rs
```diff
@@ -48,3 +48,50 @@ pub fn execute_txs_and_rollback<T: pallet::Config>(
...
+            };
+        }
+        storage::TransactionOutcome::Rollback(Result::<_, DispatchError>::Ok(()))
+    })
+    .map_err(|_| Error::<T>::FailedToCreateATransactionalStorageExecution)?;
+
+    // All transation were successfully executed
```
EvolveArt, line 91:
```suggestion
    // All transactions were successfully executed
```

crates/pallets/starknet/src/utils.rs
tdelabro: > I don't think we should simply break but rather return early with an error in the rollback no?

We need to execute this line of code `storage::TransactionOutcome::Rollback(Result::<_, DispatchError>::Ok(()))`. Which says "I successfully executed what I needed to execut

tdelabro: > Also we might want to simply use the transactional [macro](https://paritytech.github.io/polkadot-sdk/master/frame_support/attr.transactional.html) wdyt ?

For transactional function we need to return an error in order to revert storage changes. But we also need to get a vec execution infos out of this. Both can't be done

crates/pallets/starknet/src/lib.rs
```diff
@@ -1079,31 +1081,27 @@ impl<T: Config> Pallet<T> {
...
-                }
-                Err(e) => {
-                    log!(info, "Failed to estimate fee: {:?}", e);
-                    return Err(Error::<T>::TransactionExecutionFailed.into());
-                }
-            }
-        }
-        Ok(results)
+        )?
+        .map_err(|_| DispatchError::from(Error::<T>::TransactionExecutionFailed))?;
```
apoorvsadana, line 1090: can we log the error

crates/pallets/starknet/src/utils.rs
apoorvsadana, line 84: can we log the error here?

crates/pallets/starknet/src/lib.rs
tdelabro: we would be logging `PlaceHolderErrorTypeForFailedStarknetExecution`. Nothing really intresting

tdelabro:
@apoorvsadana I made some changes. More PR around here is coming due to deoxys implementing the rpc v0.6.0 that have more trace_api method.
I removed the db transactional thing. We now have to be extra careful but it simplifies the code for everybody.
And I moved all this logic to its own module, with clear warnings.

Could you review it again?

crates/pallets/starknet/src/simulations.rs
```diff
@@ -0,0 +1,134 @@
...
+                    }
+                    UserTransaction::Invoke(tx) => {
+                        let executable = tx.into_executable::<T::SystemHash>(chain_id, tx.offset_version());
+                        executable.execute(
+                            &mut BlockifierStateAdapter::<T>::default(),
+                            &block_context,
+                            &execution_config,
+                        )
+                    }
+                }
```
apoorvsadana, line 124: let's move the common code to get exeuction_info to a function? also, i didn't understand, in this case how are the changes reverted?

tdelabro: They are not. Except if you call from the RuntimeAPI, then everything you do is reverted anyway.
That's why I added those big warning comments. And we have to be extra careful not to let anyone write code that would use those from inside an extrinsic

tdelabro: > let's move the common code to get exeuction_info to a function?

Done

crates/pallets/starknet/src/utils.rs
tdelabro: > For transactional function we need to return an error in order to revert storage changes. But we also need to get a vec execution infos out of this. Both can't be done


I was wrong! It is doable. And it's in the new commits

PR #1386, opened 2024-01-21
Move Starknet core contract bindings and Anvil sandbox to Zaun

### LINKED ISSUE #1339: dev: move core contract api and client to Zaun
In the e2e test, we use `ethers::abi_gen` to create rust methods allowing us to send calls to the l1 core smart contract.
Instead of defining them here, we should use [zaun](https://github.com/keep-starknet-strange/zaun) as a lib to interact with l1.

So the mission is to removed any use of `abi_gen` from madara and use Zaun instead. `sandbox` should also go into Zaun repo (already partially achieved here: https://github.com/keep-starknet-strange/zaun/pull/7).

Last thing, in their current design the e2e test requires 1 node by test because some fail. 
We don't want this. We want the e2e test, like the rpc ones, to run a single madara node, and a single anvil instance, and have a successful back-and-forth between them.
e2e test is overkill to test some failures. We can probably do that with some integration/component tests and mocking.

m-kus:
Resolves #1339 
Blocked by https://github.com/keep-starknet-strange/zaun/pull/8

Zaun dependency needs to be fixed once the Zaun PR is merged.
I kept abigen occurrences in the DA crate because it has to be revised anyways (and it will not use core contract).


starknet-e2e-test/ethereum_settlement.rs
```diff
@@ -1,98 +1,54 @@
...
-        let config_file = File::create(config_dir.join("eth-config.json")).expect("Failed to open file for writing");
-        serde_json::to_writer(config_file, &config).expect("Failed to write eth config");
-
-        Self { madara_path, sandbox, starknet_contract }
+        let starknet_sovereign = StarknetSovereign::deploy().await;
+        let madara_temp_dir = MadaraTempDir::default();
+        Self { madara_temp_dir, starknet_sovereign }
     }
 
     async fn teardown(self) {
```
tdelabro, line 47: I don't see this method being called anywhere in the tests.
You could maybe implement `Drop` on `Context` and call `teardown` there

tdelabro: Oh mb. This is `AsyncTestContext` trait. They are automatically called by the test runner

starknet-e2e-test/src/lib.rs
```diff
@@ -1,2 +1 @@
-pub mod ethereum_sandbox;
-pub mod starknet_contract;
+pub mod starknet_sovereign;
```
tdelabro, line 1: Any reason not to move the whole `starknet_sovereign.rs` content into `lib.rs`?
We do this in order to support `starknet_validity.rs` in the future? (I'm fine with that)

tdelabro (changes requested):
Looking good!!!

starknet-e2e-test/ethereum_e2e.rs
```diff
@@ -0,0 +1,46 @@
...
+        settlement_conf: Some(settlement_conf),
+        base_path: Some(madara_temp_dir.base_path()),
+    })
+    .await;
+
+    madara.create_n_blocks(3).await?;
+    sleep(Duration::from_millis(300)).await;
+
+    let client = StarknetContractClient::new(starknet_sovereign.address(), starknet_sovereign.client());
+    let state = SettlementProvider::<Block>::get_state(&client).await.expect("Failed to get state");
```
tdelabro, line 41: you test signature is `-> Result<(), anyhow::Error>` so you can `?` rather than expect

starknet-e2e-test/ethereum_e2e.rs
```diff
@@ -0,0 +1,46 @@
...
+use madara_test_runner::{MadaraArgs, MadaraRunner, Settlement};
+use mc_settlement::ethereum::StarknetContractClient;
+use mc_settlement::{SettlementProvider, StarknetState};
+use rstest::rstest;
+use starknet_e2e_test::starknet_sovereign::StarknetSovereign;
+use tokio::time::sleep;
+
+#[rstest]
+#[tokio::test]
+async fn madara_with_sovereign_core_contract_on_ethereum() -> Result<(), anyhow::Error> {
```
tdelabro, line 16: It tests that it connect correctly?  And shares data?
Maybe a name a lil more descriptive could help

starknet-e2e-test/src/starknet_sovereign.rs
```diff
@@ -0,0 +1,147 @@
...
+        };
+
+        let client = deploy_starknet_sovereign_behind_unsafe_proxy(sandbox.client())
+            .await
+            .expect("Failed to deploy starknet contract");
+
+        Self { _sandbox: sandbox, client }
+    }
+
+    pub async fn create_settlement_conf(&self, data_path: PathBuf) -> PathBuf {
```
tdelabro, line 55: A lil bit of docstring here.
It write a settlement_conf to the disk? Based on some other file it read?

starknet-e2e-test/src/starknet_sovereign.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub async fn initialize(&self, program_hash: StarkFelt, config_hash: StarkFelt) {
+        self.initialize_with(CoreContractInitData {
+            program_hash: convert_felt_to_u256(program_hash),
+            config_hash: convert_felt_to_u256(config_hash),
+            ..Default::default()
+        })
+        .await;
+    }
+
+    pub async fn initialize_for_goerli(&self, block_number: StarkFelt, state_root: StarkFelt) {
```
tdelabro, line 88: Yeah a bit of docstring could help understand what "initialize" implies for this specific struct. Coz what is happening is not that simple

tdelabro (changes requested):

tdelabro (approved):

tdelabro:
@m-kus https://github.com/keep-starknet-strange/zaun/pull/8 is merged. Change the branch referenced and I will merge this

Issue #1345, opened 2024-01-04
dev: check that class exist before using it in `BuildGenesisConfig`


tdelabro:
### Problem

```rust
    #[pallet::genesis_build]
    impl<T: Config> BuildGenesisConfig for GenesisConfig<T> {
        fn build(&self) {
            <Pallet<T>>::store_block(0);
            frame_support::storage::unhashed::put::<StarknetStorageSchemaVersion>(
                PALLET_STARKNET_SCHEMA,
                &StarknetStorageSchemaVersion::V1,
            );

            for (class_hash, contract_class) in self.contract_classes.iter() {
                ContractClasses::<T>::insert(class_hash, contract_class);
            }

            for (sierra_class_hash, casm_class_hash) in self.sierra_to_casm_class_hash.iter() {
                CompiledClassHashes::<T>::insert(sierra_class_hash, CompiledClassHash(casm_class_hash.0));
            }

            for (address, class_hash) in self.contracts.iter() {
                ContractClassHashes::<T>::insert(address, class_hash);
            }

            for (key, value) in self.storage.iter() {
                StorageView::<T>::insert(key, value);
            }

            LastKnownEthBlock::<T>::set(None);
            // Set the fee token address from the genesis config.
            FeeTokenAddress::<T>::set(self.fee_token_address);
            SeqAddrUpdate::<T>::put(true);
        }
    }
```

Here we "declare" classes:
```
 for (class_hash, contract_class) in self.contract_classes.iter() {
        ContractClasses::<T>::insert(class_hash, contract_class);
}
```

Here we map them to their sierra
```
for (sierra_class_hash, casm_class_hash) in self.sierra_to_casm_class_hash.iter() {
       CompiledClassHashes::<T>::insert(sierra_class_hash, CompiledClassHash(casm_class_hash.0));
}
```

And here we "deploy" new contracts:
```
            for (address, class_hash) in self.contracts.iter() {
                ContractClassHashes::<T>::insert(address, class_hash);
            }
```

It would be nice for "declare", "map ot sierra" and "deploy" to make sure we are refering to `ContractClass` that actually exist. Meaning that have been stored in the first storage `ContractClasses`.

### What to do?
Add those checks, panic if we are using a `ClassHash` which don't have been declared before in the first `for loop`

kfastov:
I would like to take this

tdelabro:
Rather than checking it is in the storage. You can check it is in the in-memory array. I will probably be a bit more efficient

[...]

PR #1389, opened 2024-01-22
dev: check that class exist before using it in BuildGenesisConfig

kfastov:
# Pull Request type

- Feature

Genesis config builder does not check the existence of contract classes

Resolves: #1345

## What is the new behavior?

- The executable panics if a class hash mentioned in `contracts` or `sierra_class_hash_to_casm_class_hash` is not defined in `contract_classes`.

No


tdelabro (approved):
Great job!

tdelabro:
@kfastov can you add a line in the CHANGELOG.md file please? So the CI can run ;)

kfastov:
@tdelabro Yes sure

tdelabro:
@kfastov just fix the linters and we will merge that asap

kfastov:
@tdelabro Prettier reports errors in README.md even though I haven't changed it. I tried updating my branch but it didn't solve the problem. Can you help please?
When I run prettier locally, it reports no errors

crates/pallets/starknet/src/lib.rs
```diff
@@ -378,10 +378,21 @@ pub mod pallet {
             }
 
             for (sierra_class_hash, casm_class_hash) in self.sierra_to_casm_class_hash.iter() {
+                assert!(
+                    ContractClasses::<T>::contains_key(casm_class_hash),
```
apoorvsadana, line 382: isn't this suppoed to be Sierra class hash?

kfastov: You're right, 'class hashes' typically refer to Sierra hashes as outlined [here](https://docs.starknet.io/documentation/architecture_and_concepts/Smart_Contracts/class-hash/). However, in the genesis.json file within the test folder, the only contract class hash listed under `sierra_class_hash_to_casm_class_hash` (0x02f99bf9799ada84cd5ac0d0fe36b9d8f65efcb377cd2e8cf8309ad2daf15e4b) is actually a CASM hash, according to that mapping. This led to my confusion.
Is it a mistake so I can change that file or does it have some special meaning there?

kfastov: changed to sierra hash

kfastov: They are now, I am not sure if it's correct though. I can change the behavior of this check to succeed either the sierra or casm hash from the mapping is in the `ContractClasses`, but not sure if it should be like this

apoorvsadana: > the only contract class hash listed under sierra_class_hash_to_casm_class_hash

I didn't get this @kfastov, based on the naming it appears arr[0] is sierra class hash and arr[1] is casm class hash, are you implying that's not the case? If yes, do you know what those hashes are?

apoorvsadana: > So the keys in ContractClasses can be Casm or Sierra ContractClassHash?

No @tdelabro, they should be sierra class hashes only. The only place we use Casm class hash should be `CompiledClassHashes` afaik,

tdelabro: @apoorvsadana this won't work for cairo0 contracts. They don't have sierra, they can't have a sierra class hash

apoorvsadana: Oh ok, ya in that context it would casm, understood.

tdelabro: The design decision we are facing now is:

1)
ContractClasses: k:CasmContractClassHash v: CasmContractClass
To execute cairo0, directly index ContractClasses
To execute cairo-1 index  `CompiledClassHashes` to get the casm hash then index the `ContractClasses` with it

2)
ContractClasses: k: CasmContractClassHash |  SierraContractClassHash v: CasmContractClass
To execute bot cairo0 and cairo1, directly index `ContractClasses`

tdelabro: 1 is cleaner but 2 is probably more performant coz it spares one read.
But it's kinda negligible imo

tdelabro: @apoorvsadana @antiyro @EvolveArt wdyt?

EvolveArt: 2 looks way better to me. Why negligible we have a linear improvement no?

tdelabro: Works for me.

@kfastov can you go solution 2?
Maybe with some type aliasing so the name of the types used as key is more explicit. `SierraOrCasmClassHash`

kfastov: Do I have this right:

1) The keys in `ContractClasses` include both Sierra and Casm hashes.
2) This is also the case for `contract_classes` in the genesis.json file.
3) Therefore, when validating the Sierra to Casm mapping in `genesis.json`, we should ensure that only one of the hashes (either Sierra or Casm) is present in `ContractClasses`.
4) Once this mapping is validated, the corresponding hash (whichever is not already present) should also be added. This allows for indexing the same class using both types of hashes.

tdelabro (approved):

tdelabro:
@kfastov Good job! I wait for the CI to pass and I will merge. Thanks a lot

tdelabro:
@kfastov tests are failing

kfastov:
@tdelabro I will check tests tomorrow and finish this PR finally. Also, I have done what you said (solution 2), but not yet finished the tests (finding a way to instantiate contract classes in test).

kfastov:
There is probably an issue with `taplo`, which had an update recently. I am checking out what the reason may be. We can pin it to v0.5.2 in CI script for now

tdelabro:
@kfastov you also have to fix a lil discrepancy between the error message expect and the actual one in rpc-tests

crates/pallets/starknet/src/lib.rs
kfastov: @apoorvsadana Can I mark it as resolved?

crates/pallets/starknet/src/lib.rs
```diff
@@ -376,10 +377,26 @@ pub mod pallet {
             }
 
             for (sierra_class_hash, casm_class_hash) in self.sierra_to_casm_class_hash.iter() {
+                // ContractClasses can be indexed both by Sierra and Casm hashes, so both should be checked
+                let is_sierra_hash_in_contracts = ContractClasses::<T>::contains_key(sierra_class_hash);
+                let is_casm_hash_in_contracts = ContractClasses::<T>::contains_key(casm_class_hash);
```
tdelabro, line 382: Only `is_sierra_hash_in_contracts` should be true. 
`is_casm_hash_in_contracts` is true we have a problem: we are storing a cairo1 contract in `ContractClasses` by its casm class instead of its sierra one.
This should not happen.

crates/pallets/starknet/src/tests/build_genesis_config.rs
```diff
@@ -0,0 +1,48 @@
...
+    let genesis: GenesisConfig<default_mock::MockRuntime> = GenesisConfig {
+        sierra_to_casm_class_hash: vec![(ClassHash(1u8.into()), ClassHash(42u8.into()))],
+        contract_classes: vec![(ClassHash(1u8.into()), get_contract_class("ERC20.json", 0))],
+        ..Default::default()
+    };
+    genesis.assimilate_storage(&mut t).unwrap();
+}
+
+#[test]
+fn works_when_casm_clash_hash_in_mapping_is_known() {
```
tdelabro, line 20: This test should not fail imo

tdelabro (changes requested):

crates/pallets/starknet/src/lib.rs
kfastov: Okay I understand it now. Fixed it.

crates/pallets/starknet/src/tests/build_genesis_config.rs
kfastov: Changed this one accordingly

crates/pallets/starknet/src/lib.rs
kfastov: @tdelabro Can you check it now, please?

tdelabro:
@kfastov thanks a lot for staying involved in this one although all the rebase conflicts

PR #1390, opened 2024-01-22
✨ add state_diff to trace API

EvolveArt:

Resolves: #1331 

## What is the new behavior?

- Creates a new `CachedBlockifierStateAdapter` struct that's only being used in the tracing runtime API, it uses the blockifier caching mechanism to output a state diff after each transaction.
- This state diff is then propagated to the `simulateTransactions` endpoint.

No


crates/client/rpc-core/src/utils.rs
```diff
@@ -44,6 +46,54 @@ pub fn to_rpc_contract_class(contract_class: BlockifierContractClass) -> Result<
     }
 }
 
+/// Returns a [`StateDiff`] from a [`CommitmentStateDiff`]
+pub fn to_rpc_state_diff(commitment_state_diff: CommitmentStateDiff) -> Result<StateDiff> {
```
tdelabro, line 50: more descriptive name: `blockifier_to_rpc_state_diff_types`

crates/pallets/starknet/src/lib.rs
```diff
@@ -1087,7 +1087,7 @@ impl<T: Config> Pallet<T> {
         )?;
 
         let mut results = vec![];
-        for res in execution_results {
+        for (res, _) in execution_results {
```
tdelabro, line 1090: The fact it is not used here mean we are doing extra work for nothing.
Create another method: `execute_txs_and_rollback_with_state_diff`. Or make it return `Option<CommitmentStateDiff>` and dynamically decide which one to use

tdelabro (changes requested):
`execute_txs_and_rollback` should not waste computation

crates/client/rpc-core/src/utils.rs
EvolveArt: It was written like so to comply with the above `to_rpc_contract_class`, should we change both maybe ?

EvolveArt: Done.

crates/pallets/starknet/src/lib.rs
EvolveArt: Done, went for the 2nd approach to avoid duplication

crates/pallets/starknet/src/utils.rs
```diff
@@ -57,7 +70,8 @@ pub fn execute_txs_and_rollback<T: pallet::Config>(
                 PlaceHolderErrorTypeForFailedStarknetExecution
             });
 
-            execution_results.push((result, cached_state.to_state_diff()));
+            let state_diff = if with_state_diff { Some(cached_state.to_state_diff()) } else { None };
+            execution_results.push((result, state_diff));
```
tdelabro, line 74: The problem is that even with `with_state_diff == false` it still uses `CachedBlockifierStateAdapter` which means extra work (maintaining the cache) that won't be used.

You have to either have two methods. Or internally branch to using `CachedBlockifierStateAdapter`/`BlockifierStateAdapter` depending on the bool

tdelabro (changes requested):

crates/pallets/starknet/src/utils.rs
EvolveArt: Fixed, ended up going with another function as we want to keep static sized traits for execution-related functions imo

crates/client/rpc/src/trace_api.rs
```diff
@@ -82,6 +84,8 @@ where
                 StarknetRpcApiError::ContractError
             })?;
 
+        log::info!("is_ok: {:#?}", res[1].0.is_ok());
```
tdelabro, line 87: Debug log?

crates/client/rpc/src/trace_api.rs
```diff
@@ -240,6 +246,8 @@ fn tx_execution_infos_to_simulated_transactions<B: BlockT>(
                     })
                     .transpose()?;
 
+                let state_diff = blockifier_to_rpc_state_diff_types(state_diff.clone()).ok();
```
tdelabro, line 249: You both clone and shadow the var `state_diff`. Is there a way you could just take ownership of it?

tdelabro: Maybe using `into_iter` there `tx_types.iter().zip(transaction_execution_results.iter())`

crates/client/rpc/src/trace_api.rs
```diff
@@ -213,14 +217,16 @@ fn tx_execution_infos_to_simulated_transactions<B: BlockT>(
...
-    >,
+        CommitmentStateDiff,
+    )>,
 ) -> Result<Vec<SimulatedTransaction>, ConvertCallInfoToExecuteInvocationError> {
     let mut results = vec![];
-    for (tx_type, res) in tx_types.iter().zip(transaction_execution_results.iter()) {
+    for (tx_type, (res, state_diff)) in tx_types.iter().zip(transaction_execution_results.iter()) {
         match res {
             Ok(tx_exec_info) => {
+                // println!("tx_exec_info: {:#?}", tx_exec_info);
```
tdelabro, line 229: debug log to cleanup

crates/node/src/commands/run.rs
```diff
@@ -186,13 +186,21 @@ pub fn run_node(mut cli: Cli) -> Result<()> {
 
     let da_client = match cli.run.da_layer {
         Some(da_layer) => {
-            let da_conf = cli.run.clone().da_conf.unwrap_or({
```
tdelabro, line 189: The issue was using unwrap_or rather than unwrap_or_else?

crates/pallets/starknet/src/lib.rs
```diff
@@ -1134,11 +1134,10 @@ impl<T: Config> Pallet<T> {
     pub fn simulate_transactions(
         transactions: Vec<UserTransaction>,
         simulation_flags: SimulationFlags,
-    ) -> Result<Vec<Result<TransactionExecutionInfo, PlaceHolderErrorTypeForFailedStarknetExecution>>, DispatchError>
-    {
+    ) -> Result<Vec<(TransactionSimulationResult, CommitmentStateDiff)>, DispatchError> {
         let chain_id = Self::chain_id();
 
-        let tx_execution_results = execute_txs_and_rollback::<T>(
+        let tx_execution_results = execute_txs_and_rollback_with_state_diff::<T>(
```
tdelabro, line 1140: When you will rebase the method won't be there. it's moved to pallet/starnet/src/simulations.rs

crates/runtime/Cargo.toml
```diff
@@ -67,6 +67,8 @@ starknet-core = { workspace = true }
 starknet-ff = { workspace = true }
 starknet_api = { workspace = true }
 
+log = { workspace = true }
```
tdelabro, line 70: cleanup

crates/runtime/src/lib.rs
```diff
@@ -284,8 +285,10 @@ impl_runtime_apis! {
             Starknet::estimate_message_fee(message)
         }
 
-        fn simulate_transactions(transactions: Vec<UserTransaction>, simulation_flags: SimulationFlags) -> Result<Vec<Result<TransactionExecutionInfo, PlaceHolderErrorTypeForFailedStarknetExecution>>, DispatchError> {
-            Starknet::simulate_transactions(transactions, simulation_flags)
+        fn simulate_transactions(transactions: Vec<UserTransaction>, simulation_flags: SimulationFlags) -> Result<Vec<(TransactionSimulationResult, CommitmentStateDiff)>, DispatchError> {
+            let txs = Starknet::simulate_transactions(transactions, simulation_flags);
+            log::info!("is_ok: {:#?}", txs.as_ref().unwrap()[1].0.is_ok());
```
tdelabro, line 290: cleanup

tdelabro (changes requested):

crates/node/src/commands/run.rs
EvolveArt: yes

EvolveArt: but we can't use unwrap or else in this case so I figured a match would work just as well

crates/pallets/starknet/src/types.rs
```diff
@@ -22,6 +24,8 @@ pub type StorageSlot = (StorageKey, Felt252Wrapper);
 pub type CasmClassHash = ClassHash;
 pub type SierraClassHash = ClassHash;
 
+pub type TransactionSimulationResult = Result<TransactionExecutionInfo, PlaceHolderErrorTypeForFailedStarknetExecution>;
```
tdelabro, line 27: This is the sencond time you define this type.
Define it in primitive/simulations instead, so you can access it from both crates

tdelabro (changes requested):

tdelabro (approved):

tdelabro (approved):

PR #1392, opened 2024-01-22
feat: L1 gas price is configurable in the RuntimeConfig

tdelabro:
Lucas needs it for https://github.com/keep-starknet-strange/snos/issues/68

It make sure that the value used in execution block context and the one in the block header are the same

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next release
 
 - feat: types in `mp-transactions` impl a method to get their version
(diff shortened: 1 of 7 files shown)


apoorvsadana (approved):
Is this a temporary solution for now because ideally we would need to fetch this data?

tdelabro:
> Is this a temporary solution for now because ideally, we would need to fetch this data?

Yes. The next step is to have it be an "inherent", set by the block producer based on fetching data and computing a moving average.

PR #1397, opened 2024-01-23
feat: ✅ estimateMessageFee RPC call implementation

Tbelleng:
# Pull Request type

Implementation of the following RPC call : 

```- estimateMessageFee```

- Feature

Solve issue #1301 

No

### CODE REVIEW

File: crates/client/rpc/Cargo.toml [hunk 1]
```diff
@@ -42,6 +42,7 @@ sc-network-sync = { workspace = true }
 blockifier = { workspace = true, default-features = true }
 starknet-core = { workspace = true }
 starknet-ff = { workspace = true }
+starknet-providers = { workspace = true }
```
tdelabro, line 45: It looks like it is not used

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
...
+    /// Estimate the L2 fee of a message sent on L1
+    ///
+    /// # Arguments
+    ///
+    /// * `message` - starknet transaction request
+    /// * `block_id` - hash of the requested block, number (height), or tag
+    ///
+    /// # Returns
+    ///
+    /// * `fee_estimation` - the fee estimation (gas consumed, gas price, overall fee, unit)
```
tdelabro, line 989: `fee_estimation` is not a variable that exists anywhere. Return objects are not named in rust.
I think you can just use the type itself here

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
         Ok(estimates)
     }
 
+    /// Estimate the L2 fee of a message sent on L1
+    ///
+    /// # Arguments
+    ///
+    /// * `message` - starknet transaction request
+    /// * `block_id` - hash of the requested block, number (height), or tag
```
tdelabro, line 985: "hash, number (height), or tag of the requested block"

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
         Ok(estimates)
     }
 
+    /// Estimate the L2 fee of a message sent on L1
+    ///
+    /// # Arguments
+    ///
+    /// * `message` - starknet transaction request
```
tdelabro, line 984: "starknet transaction request" that doesn't make a lot of sense to me.
Wouldn't we better with something along the lines of "the message to estimate"

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
...
+    ///
+    /// BlockNotFound : If the specified block does not exist.
+    /// ContractNotFound : If the specified contract address does not exist.
+    /// ContractError : If there is an error with the contract.
+    async fn estimate_message_fee(&self, message: MsgFromL1, block_id: BlockId) -> RpcResult<FeeEstimate> {
+        let substrate_block_hash = self.substrate_block_hash_from_starknet_block(block_id).map_err(|e| {
+            error!("'{e}'");
+            StarknetRpcApiError::BlockNotFound
+        })?;
+        let best_block_hash = self.client.info().best_hash;
```
tdelabro, line 1001: This variable is not used

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
...
+            error!("Failed to convert MsgFromL1 to UserTransaction: {e}");
+            StarknetRpcApiError::InternalServerError
+        })?;
+
+        let fee_estimate = self
+            .client
+            .runtime_api()
+            .estimate_message_fee(substrate_block_hash, message)
+            .map_err(|e| {
+                error!("Request parameters error: {e}");
```
tdelabro, line 1014: "Request parameters error" are you sure?
I think it's just a runtimeAPI error. So it means we failed to create the condition to run your code. I don't think it has anything to do with the parameters

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
...
+                StarknetRpcApiError::InternalServerError
+            })?
+            .map_err(|e| {
+                error!("Failed to call function: {:#?}", e);
+                StarknetRpcApiError::ContractError
```
tdelabro, line 1019: Here it's not that you failed to call the function. It could also be that that the function execution failed

crates/client/rpc/src/lib.rs
```diff
@@ -977,6 +977,53 @@ where
...
+            })?;
+
+        let estimate = FeeEstimate { gas_price: 10, gas_consumed: fee_estimate.1, overall_fee: fee_estimate.0 };
```
tdelabro, line 1022: you should rebase maybe. I think you should return the gas_price from the runtime at the same time. YOu can get it by running ` T::L1GasPrice::get()` in the pallet

crates/pallets/starknet/src/lib.rs
```diff
@@ -1106,6 +1106,28 @@ impl<T: Config> Pallet<T> {
         Ok(results)
     }
 
+    pub fn estimate_message_fee(message: HandleL1MessageTransaction) -> Result<(u64, u64), DispatchError> {
```
tdelabro, line 1109: Add some warning, saying that this should only be called from the runtimeAPI, never from inside an extrinsic execution flow coz i WILL execute the message

crates/pallets/starknet/src/lib.rs
```diff
@@ -1106,6 +1106,28 @@ impl<T: Config> Pallet<T> {
...
+            )
+            .map_err(|e| {
+                log::error!("failed to execute invoke tx: {:?}", e);
+                Error::<T>::TransactionExecutionFailed
+            })?;
+
+        if let Some(l1_gas_usage) = tx_execution_infos.actual_resources.0.get("l1_gas_usage") {
+            Ok((tx_execution_infos.actual_fee.0 as u64, *l1_gas_usage))
+        } else {
+            Err(Error::<T>::TransactionExecutionFailed.into())
```
tdelabro, line 1127: If you rebase you will find a dedicated error message for l1_gas_usage field missing

crates/primitives/transactions/src/lib.rs
```diff
@@ -190,3 +191,17 @@ pub struct HandleL1MessageTransaction {
     pub entry_point_selector: Felt252Wrapper,
     pub calldata: Vec<Felt252Wrapper>,
 }
+
+impl TryFrom<MsgFromL1> for HandleL1MessageTransaction {
```
tdelabro, line 195: Can't you use `From` ?
I don't see any error being returned

crates/primitives/transactions/src/lib.rs
```diff
@@ -190,3 +191,17 @@ pub struct HandleL1MessageTransaction {
...
+    type Error = Error;
+
+    fn try_from(msg: MsgFromL1) -> Result<Self, Self::Error> {
+        let calldata = msg.payload.into_iter().map(|felt| felt.into()).collect();
+        Ok(Self {
+            contract_address: msg.to_address.into(),
+            nonce: 0u32.into(),
```
tdelabro, line 202: I don't think you should do that. They need a nonce.
Try `impl From<(Nonce, MsgFromL1)>` instead.

Or don't impl from. If you are just using it once in you very special case where you want nonce to be 0.

tdelabro: If you send a real message from l1 with nonce 0 and then try to execute that, it will probably fail.

I think you have to disable nonce validation here: `&RuntimeExecutionConfigBuilder::new::<T>().with_query_mode().build(),`

crates/pallets/starknet/src/lib.rs
tdelabro: Oh no sorry. It's a PR of mine which is not merged yet.

tdelabro (approved):

Issue #1369, opened 2024-01-12
dev: Get the block status from the actual block in `get_block_with_tx_hashes`


LucasLvy:
Currently we hardcode `AcceptedOnL2`, get the block status from the block itself.

azurwastaken:
would like to take a look at this one if possible

tdelabro:
@azurwastaken sure!

I think you can use the `MessagingDb::last_synced_l1_block_with_event` method to get the last block number we know for sure has been seen on L1.
If the block requested has a block number lower or equal to that one, it is `AcceptedOnL1`. 

If not, but we are aware of this block, it's `AcceptedOnL2`.

PR #1402, opened 2024-01-24
dev: Get the block status from the actual block in get_block_with_tx_hashes

azurwastaken:
# Pull Request type

- Feature

Currently the block status is hardcoded with Accepted_on_L2

Resolves: #1369

## What is the new behavior?

- If the last block number is superior or equal to retrieved block number, we set the status to Accepted_on_L1 
- Otherwise, we know the block so its Accepted on L2

No


crates/client/rpc/src/lib.rs
```diff
@@ -847,12 +848,21 @@ where
...
         };
+        let block_status = match self.backend.messaging().last_synced_l1_block_with_event() {
+            Ok(l1_block) => {
+                if l1_block.block_number >= starknet_block.header().block_number {
+                    BlockStatus::AcceptedOnL1
+                } else {
+                    BlockStatus::AcceptedOnL2
+                }
+            },
+            Err(e) => Err(anyhow!("couldn't retrieve block from l1 : {}", e)) ?,
```
EvolveArt, line 859: we should probably do the same as above here i.e return a CallError::Failed
```
CallError::Failed(anyhow::anyhow!(
                        "The hash cached for block with hash {block_hash:?} is an invalid felt: '{tx_hash}'. The \
                         caching db has probably been tempered"
                    ))
                
```

EvolveArt (changes requested):

crates/client/rpc/src/lib.rs
azurwastaken: done

EvolveArt (approved):

crates/client/rpc/src/lib.rs
```diff
@@ -847,12 +847,21 @@ where
...
         };
+        let block_status = match self.backend.messaging().last_synced_l1_block_with_event() {
+            Ok(l1_block) => {
+                if l1_block.block_number >= starknet_block.header().block_number {
+                    BlockStatus::AcceptedOnL1
+                } else {
+                    BlockStatus::AcceptedOnL2
+                }
+            }
+            Err(e) => Err(CallError::Failed(anyhow::anyhow!("couldn't retrieve block from l1 : {}", e)))?,
```
tdelabro, line 858: return `StarknetRpcApiError::InternalServerError` and log the DbError you got from `last_synced_l1_block_with_event`

tdelabro (changes requested):
good job

crates/client/rpc/src/lib.rs
azurwastaken: modified :)

tdelabro (approved):

Issue #775, opened 2023-06-28
feat: store class hash and sierra program mapping


apoorvsadana:
# Feature Request

**Describe the Feature Request**

Once we have Declare V2 implemented from this #744, we will need a way to store the Sierra program which is passed from the RPC call. Right now, the Sierra program isn't stored anywhere and so we aren't able to return it when the user calls `get_class_at`. When storing this mapping, we must do it in a way that it is communicated to all the other nodes.

**Describe Preferred Solution**

We can implement this in the run time but that should be our last option as the runtime should only have the information needed for the chain state. The Sierra program is not essential for the chain state and so we must store them elsewhere. The db in the client isn't gossiped between the nodes and is a local copy. So we need to explore how we can get this done.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes

apoorvsadana:
the implementation here would mostly use the same approach used here #723

apoorvsadana:
@tdelabro can we open this issue as well

tdelabro:
I think I'm tackling this in my refactoring PR

apoorvsadana:
Oh ok, awesome!

tdelabro:
Actually I think this is still required

tdelabro:
closed by #1409

Issue #1381, opened 2024-01-18
feat: Save sierra program in the runtime


LucasLvy:
To run the os we need the sierra contract class, it's currently not saved anywhere, save it somewhere that is accessible for the commitment state diff service.

tdelabro:
I think it should be in the client, not the runtime.
The runtime is supposed to receive the sierra contract and the sierra hash. Check the hash matches the sierra contract, compile the sierra to casm, check the casm hash. But it does not need to be stored for anything else in the runtime.
So, the client can store the contract. It should probably hash it in order to make sure the sierra_hash contained in the tx is the good one.

tdelabro:
In client, `add_declare_transaction`, for ` BroadcastedDeclareTransactionV2` only
- [ ] hash the contract_class, make sure it it's hash matches the content of the `class_hash` field in the tx
- [ ] store a mapping from `class_hash` to `contract_class` into the madara backend db, in a new db created in `crates/client/db`
- [ ] proceed to publish the tx as usual


Problem with this approach:
- if the tx then fail in the runtime (no accepted in the mempool, execution fails coz no funds), the node would have stored an unusable class. Someone could spam this in order to fill the node disk with garbage class
- it only works for a single node, the others ones won't receive the RPC so they won't be able to replicate this

So I think we should go for something else

tdelabro:
Another approach:

- Don't change the RPC at all
- when a node receives a block that went through consensus, parse the list of txs, and find the ones that are `DeclareTransactionV2` and upper. Do the storing here. No need to check hashes, it should be correct if part of a block. Same DB mapping in the client.

tdelabro:
closed by #1409

PR #1409, opened 2024-01-26
feat: on  store sierra contract class in madara backend db

tdelabro:

Data about the sierra class is lost after it is used

## What is the new behavior?

- the sierra contract class is now stored in the Madara backend database

## Other information

In the future we will make node ask each other for those sierra contract classes


tdelabro:
closes #775 #1381

LucasLvy (approved):

Issue #1415, opened 2024-01-29
bug: 


tdelabro:
## Problem
There is a lot of places in the rpc implementation where this line has been copy-pasted:
```rust
        let starknet_block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).unwrap_or_default();
```

What it does is that it returns a block from storage or an empty block with all values at default if it does not find it.
There may be exceptions but we would rather be returning errors in the RPC if the block doesn't exist.

## What to do?
Go check all those occurrences and make sure to replace them by proper error handling when required

fishseabowl:
@tdelabro could I take this? Thanks

PR #1420, opened 2024-01-30
Return an error to replace the default value if the block does not exist

fishseabowl:
# Pull Request type

- Bugfix

If the block does not exist, an empty block is returned with all values as defaults
Resolves: #1415 

## What is the new behavior?

If the block does not exist, an error will be reported.


crates/client/rpc/src/lib.rs
```diff
@@ -137,7 +137,10 @@ where
     pub fn current_block_hash(&self) -> Result<H256, ApiError> {
         let substrate_block_hash = self.client.info().best_hash;
 
-        let starknet_block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).unwrap_or_default();
+        let starknet_block =
+            get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).ok_or(ApiError::UnknownBlock(
```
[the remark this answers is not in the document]
tdelabro: And rather use StarknetAPiError

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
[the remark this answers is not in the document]
fishseabowl: @tdelabro Thank you for your comment. Do you mean if get_block_by_block_hash returns Err here, just ignore this error and don't propagate this error?

> ```
> pub fn get_block_by_block_hash<B, C>(client: &C, block_hash: <B as BlockT>::Hash) -> Option<StarknetBlock>
> where
>     B: BlockT,
>     C: HeaderBackend<B>,
> {
>     let header = client.header(block_hash).ok().flatten()?;
>     let digest = header.digest();
>     let block = find_starknet_block(digest).ok()?;
>     Some(block)
> }
> ```
> 
> Change the signature of `get_block_by_block_hash` so that it returns a Result. We are ignoring the errors here.

@tdelabro Thank you for your comment. Do you mean if `get_block_by_block_hash` returns `Err` here, just ignore this error and don't propagate this error? Thanks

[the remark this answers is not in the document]
tdelabro: No I meant that the current behaviour is that we are ignoring the error. Changing them to `None` with `ok()`. We should bubble those errors up instead

crates/client/rpc/src/lib.rs
```diff
@@ -843,7 +849,11 @@ where
             StarknetRpcApiError::BlockNotFound
         })?;
 
-        let starknet_block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).unwrap_or_default();
+        let starknet_block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).ok_or_else(|| {
+            error!("Block not found");
```
apoorvsadana, line 853: we can move the "Block not found" log inside `get_block_by_block_hash` and throw the error directly here after implementing what @tdelabro said.

crates/client/rpc-core/src/utils.rs
```diff
@@ -89,15 +89,16 @@ fn to_legacy_entry_point(entry_point: EntryPoint) -> Result<LegacyContractEntryP
 }
 
 /// Returns the current Starknet block from the block header's digest
-pub fn get_block_by_block_hash<B, C>(client: &C, block_hash: <B as BlockT>::Hash) -> Option<StarknetBlock>
+pub fn get_block_by_block_hash<B, C>(client: &C, block_hash: <B as BlockT>::Hash) -> Result<StarknetBlock>
```
apoorvsadana, line 92: let's throw a `StarknetRpcApiError`?

crates/client/rpc/src/lib.rs
```diff
@@ -137,11 +137,10 @@ where
...
         let substrate_block_hash = self.client.info().best_hash;
 
-        let starknet_block =
-            get_block_by_block_hash(self.client.as_ref(), substrate_block_hash).ok_or(ApiError::UnknownBlock(
-                format!("Failed to get Starknet block for Substrate block hash '{substrate_block_hash}' "),
-            ))?;
-
+        let starknet_block = match get_block_by_block_hash(self.client.as_ref(), substrate_block_hash) {
+            Ok(block) => block,
+            Err(err) => return Err(ApiError::UnknownBlock(err.to_string())),
```
tdelabro, line 142: StarknetRpcApiError::BlockNotFound

crates/client/rpc/src/lib.rs
```diff
@@ -180,8 +179,10 @@ where
 
         let substrate_block_hash = self.substrate_block_hash_from_starknet_block(block_id)?;
 
-        let starknet_block = get_block_by_block_hash(self.client.as_ref(), substrate_block_hash)
-            .ok_or("Failed to retrieve the substrate block number".to_string())?;
+        let starknet_block = match get_block_by_block_hash(self.client.as_ref(), substrate_block_hash) {
+            Ok(block) => block,
+            Err(err) => return Err(format!("Error getting Starknet block by block hash: {}", err)),
```
tdelabro, line 184: StarknetRpcApiError::BlockNotFound

tdelabro: or just use `?`

crates/client/rpc/src/lib.rs
tdelabro: same maybe just use `?`

tdelabro (changes requested):

crates/client/rpc-core/src/utils.rs
fishseabowl: @apoorvsadana `StarknetRpcApiError` is defined in the `mc-rpc` crate, `get_block_by_block_hash` is defined in `mc-rpc-core`, and currently `mc-rpc-core` is a dependency of `mc-rpc`. It will cause a circular dependency between two crates if  throwing a `StarknetRpcApiError`. Do you have any suggestions? Thanks

crates/client/rpc/src/lib.rs
fishseabowl: @tdelabro 
```rust
137 pub fn current_block_hash(&self) -> Result<H256, ApiError> 
```
Calling function `current_block_hash` signature requires ApiError, do you mean I change `current_block_hash` signature to return StarknetRpcApiError? Thanks

crates/client/rpc/src/lib.rs
fishseabowl: @tdelabro 
```rust
fn substrate_block_number_from_starknet_block(&self, block_id: BlockId) -> Result<u64, String> {
```
Calling function `substrate_block_number_from_starknet_block` signature requires `String`, Should I return `err.to_string()`?  Thanks

crates/client/rpc/src/lib.rs
tdelabro: yes

crates/client/rpc/src/lib.rs
tdelabro: Change the funtion signature so that the error type is not `String` anymore but `StarknetRpcApiError`

tdelabro (changes requested):

crates/client/rpc-core/src/utils.rs
apoorvsadana: Understood, we can keep it as it is. Can we log the error in this function before returning it?

PR #1426, opened 2024-01-31
fix: id cairo1 contract by their sierra class hash

tdelabro:

In genesis.json and in tests, cairo1 contract were identified by their casm class hash

## What is the new behavior?

In genesis.json and in tests, cairo1 contract are identified by their sierra class hash


apoorvsadana (approved):
lgtm

Issue #1414, opened 2024-01-29
bug: parallel read requests are very slow


apoorvsadana:
Currently the RPC layer of Madara can't handle read requests in parallel. As a result, sending multiple read calls on a Madara node takes a lot of time. The issue hasn't been diagnosed yet, possible reasons include a lock which is being taken by read calls or something mentioned [here](https://github.com/paritytech/substrate/issues/8846).

### Expected behaviour
Multiple read calls can be processed in parallel.

apoorvsadana:
It is because 

1. Even though users are reading via full nodes, there will be multiple full nodes syncing on Madara. At the current speed, this will make syncing slow for the full nodes and will be a bottleneck when autoscaling full nodes.
2. For now, till we don't have the p2p spec, the full nodes being used will be Madara nodes. That's what we are planning for the Sharingan testnet.

tdelabro:
>  there will be multiple full nodes syncing on Madara

They are not syncing through RPC calls, are they?

> the full nodes being used will be Madara nodes

More likely they will be deoxys nodes? With some optimization for reading and not for writing

apoorvsadana:
> They are not syncing through RPC calls, are they?
I mean until we have p2p it has to be via the RPC right?

> More likely they will be deoxys nodes? With some optimization for reading and not for writing
Ideally it should be deoxys but that would also need p2p first right?

apoorvsadana:
Also, since deoxys is also a Madara fork I guess it would fall into the same issue? Do you know how this works on current Substrate chains? Because over there it's the same node running as a full node instead of a validator. So we should be missing something otherwise parallel read calls should work by default.

tdelabro:
> I mean until we have p2p it has to be via the RPC right?

We already have p2p. Substrate p2p to share block. I think we will only be missing the sharing of sierra classes, which I will impl soon

tdelabro:
I think deoxys should try to minimize runtime call to only the simulate/estimate/trace ones. And have a big disk storage + cache to serve requests as fast as possible.
In madara we purposefully refuse to store some stuffs and rely to runtime call to do some read. Maybe that can be changed for us too, as long as it doesn't slow down block production.
Overall I agree, the less runtime call the better

apoorvsadana:
[...]

PR #1427, opened 2024-01-31
use cache for receipts and event txs

apoorvsadana:
Modifies code to use the cache for `getEvents` and `getTransactionReceipt` for reading txn_hashes if `cache` flag is enabled.

### DIFF
--- a/.github/workflows/starknet-rpc-tests.yml
+++ b/.github/workflows/starknet-rpc-tests.yml
@@ -31,9 +31,17 @@ jobs:
       - name: Setup dev chain
         run: |
           ./target/release/madara setup --chain=dev --from-local=configs
-      - name: Run rpc test
+      - name: Run rpc test without cache
         run: |-
           ./target/release/madara --dev --sealing=manual &
+          MADARA_RUN_PID=$!
+          while ! echo exit | nc localhost 9944; do sleep 1; done
+          cd starknet-rpc-test
+          cargo test
+          kill $MADARA_RUN_PID
+      - name: Run rpc test with cache
+        run: |-
+          ./target/release/madara --dev --sealing=manual --cache &
           MADARA_RUN_PID=$!
           while ! echo exit | nc localhost 9944; do sleep 1; done
           cd starknet-rpc-test
--- a/crates/client/rpc/src/events/mod.rs
+++ b/crates/client/rpc/src/events/mod.rs
@@ -90,28 +90,8 @@ where
         let mut tx_hash_and_events: Vec<(Felt252Wrapper, starknet_api::transaction::Event)> = vec![];
         for (index, event) in index_and_events {
             let tx_index = index as usize - inherent_count;
-            if let Some(txn_hashes) = &txn_hashes {
-                let tx_hash = (&txn_hashes
-                    .get(tx_index)
-                    .ok_or_else(|| {
-                        error!("Failed to retrieve transaction hash from cache, invalid index {}", tx_index);
-                        StarknetRpcApiError::InternalServerError
-                    })?
-                    .0)
-                    .try_into()
-                    .map_err(|_| {
-                        error!("Failed to convert transaction hash");
-                        StarknetRpcApiError::InternalServerError
-                    })?;
-                tx_hash_and_events.push((tx_hash, event));
-            } else {
-                let transaction = &starknet_txs.get(tx_index).ok_or_else(|| {
-                    error!("Failed to retrieve transaction hash from starknet txs, invalid index {}", tx_index);
-                    StarknetRpcApiError::InternalServerError
-                })?;
-                let tx_hash = transaction.compute_hash::<H>(chain_id, false);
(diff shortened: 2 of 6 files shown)


EvolveArt (changes requested):
I see that we have this
```
fn get_cached_transaction_hashes(&self, block_hash: H256) -> Option<Vec<H256>> {
        self.backend.mapping().cached_transaction_hashes_from_block_hash(block_hash).unwrap_or_else(|err| {
            error!("{err}");
            None
        })
    }
    ```
    
    Not really clear to me it should live inside the mapping DB, we should probably create a separated cache DB.
On another note the error handling is not really satisfying here, wdyt ?

**Comment by apoorvsadana** [comment]
> Not really clear to me it should live inside the mapping DB, we should probably create a separated cache DB.
On another note the error handling is not really satisfying here, wdyt ?

Ya we need to implement the cache properly as we were discussing here #1414. For now, this is a fix to allow us to integrate gomu gomu and till we have the full nodes ready. For the error message, I have added a better error message, I am not sure how else we should handle it.

File: crates/client/rpc/src/events/mod.rs [hunk 1]
```diff
@@ -83,8 +109,8 @@ where
             .into_iter()
             .map(|(tx_hash, event)| EmittedEvent {
                 from_address: Felt252Wrapper::from(event.from_address).0,
-                keys: event.content.keys.into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
-                data: event.content.data.0.into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
+                keys: event.content.keys.clone().into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
+                data: event.content.data.0.clone().into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
```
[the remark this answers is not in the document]
apoorvsadana: fixed

EvolveArt:
> > Not really clear to me it should live inside the mapping DB, we should probably create a separated cache DB.
> > On another note the error handling is not really satisfying here, wdyt ?
> 
> Ya we need to implement the cache properly as we were discussing here #1414. For now, this is a fix to allow us to integrate gomu gomu and till we have the full nodes ready. For the error message, I have added a better error message, I am not sure how else we should handle it.

Alright let's fix the CI and create a new issue then

crates/client/rpc/src/lib.rs
```diff
@@ -1436,23 +1436,71 @@ where
...
+            self.client.runtime_api().extrinsic_filter(substrate_block_hash, block_extrinsics).map_err(|e| {
+                error!("Failed to filter extrinsics. Substrate block hash: {substrate_block_hash}, error: {e}");
                 StarknetRpcApiError::InternalServerError
-            })?
-            .expect("the transaction should be present in the substrate extrinsics"); // not reachable
+            })?;
+        let txn_hashes = self.get_cached_transaction_hashes(starknet_block.header().hash::<H>().into());
+        let mut tx_index = None;
+        let mut transaction = None;
+        for (index, tx) in transactions.iter().enumerate() {
```
EvolveArt, line 1448: let's move this logic to a function as it's present twice

EvolveArt (changes requested):
Just to confirm these RPC calls are already covered by tests right ?

apoorvsadana:
> Alright let's fix the CI and create a new issue then

We have this one already, or do you want something else? https://github.com/keep-starknet-strange/madara/issues/947

crates/client/rpc/src/lib.rs
apoorvsadana: fixed

apoorvsadana:
> Just to confirm these RPC calls are already covered by tests right ?

The RPC calls are but they weren't running in cache mode. So I added rpc tests with cache which runs all the tests again but in cache mode.

.github/workflows/starknet-rpc-tests.yml
```diff
@@ -31,11 +31,19 @@ jobs:
...
-      - name: Run rpc test
+      - name: Run rpc test without cache
         run: |-
           ./target/release/madara --dev --sealing=manual &
           MADARA_RUN_PID=$!
           while ! echo exit | nc localhost 9944; do sleep 1; done
           cd starknet-rpc-test
           cargo test
           kill $MADARA_RUN_PID
+      - name: Run rpc test with cache
```
tdelabro, line 42: Is it really necessary?

tdelabro: Our CI is already quite long

apoorvsadana: Ya I agree its long as well. @EvolveArt do you have any views?

EvolveArt: well it's definitely possible that the cache flag breaks stuff. If we have to choose one path I'd rather go for with cache only then.

EvolveArt (approved):

Issue #1412, opened 2024-01-29
feat(rpc): Implement testing suit for estimate message fee method


antiyro:
## feat(rpc): Implement testing suit for estimate message fee method

Please implement tests for the `starknet_estimateMessageFee` method as implemented [here](https://github.com/keep-starknet-strange/madara/blob/main/crates/client/rpc/src/lib.rs#L1007) and referenced in the official [JSON-RPC specs](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L643)

Tbelleng:
I can take it sir !

PR #1431, opened 2024-02-02
Add tests for estimate_message_fee RPC Call

Tbelleng:
Add some tests for estimate_message_fee RPC call 

# Pull Request type

- Testing

Resolves: #1412

No


Tbelleng:
Do you guys know can i create a message L1->L2 in order to estimate it thanks to the rpc Call ? 
Im trying to do that in my last test and @tdelabro told me that i have to use the ```"test_l1_handler_store_under_caller_address"``` cairo function in order to do so. But as far as i know, all cairo function are executed by using an account (because to simulate tx you just have to do L2->L2). But concerning estimateMessageFee, the transaction that we have to estimate is a L1->L2 message (where L1 account is Starknet contract on ETH and L2 account is ETH contract on Starknet), so i guess i dont have to initialize any account right ?

starknet-rpc-test/estimate_message_fee.rs
```diff
@@ -0,0 +1,70 @@
...
+
+use assert_matches::assert_matches;
+use rstest::rstest;
+use starknet_core::types::{BlockId, BlockTag, EthAddress, MsgFromL1, StarknetError};
+use starknet_core::utils::get_selector_from_name;
+use starknet_ff::FieldElement;
+use starknet_providers::ProviderError::StarknetError as StarknetProviderError;
+use starknet_providers::{MaybeUnknownErrorCode, Provider, StarknetErrorWithMessage};
+use starknet_rpc_test::constants::{L1_CONTRACT_ADDRESS, TEST_CONTRACT_ADDRESS};
+use starknet_rpc_test::fixtures::{madara, ThreadSafeMadaraClient};
```
tdelabro, line 11: fmt, add a space there

starknet-rpc-test/estimate_message_fee.rs
```diff
@@ -0,0 +1,70 @@
...
+        rpc.estimate_message_fee(message, BlockId::Tag(BlockTag::Latest)).await,
+        Err(StarknetProviderError(StarknetErrorWithMessage { code: MaybeUnknownErrorCode::Known(code), .. })) if code == StarknetError::ContractError
+    );
+
+    Ok(())
+}
+
+// #[rstest]
+// #[tokio::test]
+// async fn works_ok(madara: &ThreadSafeMadaraClient) -> Result<(), anyhow::Error> {
```
tdelabro, line 54: You need to uncomment this and do an estimate_message that works

tdelabro: The todo later is only about the fact of actually executing the message afterward and compare it's cost with the estimation

tdelabro (changes requested):

starknet-rpc-test/estimate_message_fee.rs
```diff
@@ -0,0 +1,72 @@
...
+#[tokio::test]
+async fn works_ok(madara: &ThreadSafeMadaraClient) -> Result<(), anyhow::Error> {
+    let rpc = madara.get_starknet_client().await;
+
+    let message: MsgFromL1 = MsgFromL1 {
+        from_address: EthAddress::from_hex(L1_CONTRACT_ADDRESS).unwrap(),
+        to_address: 3u64.into(),
+        entry_point_selector: 2u64.into(),
+        payload: vec![1u64.into()],
+    };
```
tdelabro, line 63: Bha la par contre il va failloir utiliser des valeurs qui fonctionnent.

La fonction que tu veux call c'est celle là:
```rust
@l1_handler
func test_l1_handler_store_under_caller_address{syscall_ptr: felt*}(from_address: felt, value: felt) {
    storage_write(address=from_address, value=value);
    return ();
}
```
dans  cairo-contracts/src/test_contracts/test.cairo

Donc là il faut que tu changes to entry_point_selector:
```shell
$ starkli selector test_l1_handler_store_under_caller_address
0x014093c40d95d0a3641c087f7d48d55160e1a58bc7c07b0d2323efeeb3087269
```

Et ton addresse:
```rust
pub const TEST_CONTRACT_ADDRESS: &str = "0x0000000000000000000000000000000000000000000000000000000000001111";
```
dans `starknet-rpc-test/src/constants.rs`

Sinon il appelle un entrypoint qui existe pas sur un contract qui existe pas.

tdelabro (changes requested):

tdelabro (approved):

PR #1434, opened 2024-02-03
RPC - Support for pending state

### LINKED ISSUE #1405: feat(rpc): Support for pending state
### Current Behavior

Block times on `mainnet` can be prohibitively long for certain applications. To address this, Starknet introduced the concept of a `pending` block, which is the block currently under construction.

Currently, Madara does not support pending state and only serves finalized state. However, as defined in the [JSON-RPC specs](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_api_openrpc.json), supporting pending blocks is necessary for full Starknet compatibility.

### Solution

1. **Identify RPC calls affected by this issue**:
   - `starknet_getBlockWithTxs`: Pending blocks can be called [here](https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L81).
   - `starknet_getBlockWithTxHashes`: Similar to the above, pending blocks [here](https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L50).
   - `starknet_getStateUpdate`: Pending state updates can be called [here](https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L124).
   - `starknet_getTransactionReceipt`: Pending transactions can be called [here](https://github.com/starkware-libs/starknet-specs/blob/49665932a97f8fdef7ac5869755d2858c5e3a687/api/starknet_api_openrpc.json#L353).

2. **Implementing the Solution**:
   - Determine if the `block_id` parameters correspond to `pending`.
   - Call the substrate `author_pendingExtrinsics` method to access the pending extrinsic data where blocks and transactions are added and encoded.
   - Identify how to decode them on the client side.
   - Format and return them as they are.

Please discuss under this issue the potential implications of such an implementation. We'll proceed with a simple implementation within Madara for `starknet_getBlockWithTxHashes` to gain a clearer understanding of how to proceed.

bidzyyys:
Closes #1405

### DIFF
--- a/crates/client/rpc/src/lib.rs
+++ b/crates/client/rpc/src/lib.rs
@@ -22,7 +22,7 @@ pub use mc_rpc_core::{
     StarknetWriteRpcApiServer,
 };
 use mc_storage::OverrideHandle;
-use mp_felt::Felt252Wrapper;
+use mp_felt::{Felt252Wrapper, Felt252WrapperError};
 use mp_hashers::HasherT;
 use mp_transactions::compute_hash::ComputeTransactionHash;
 use mp_transactions::to_starknet_core_transaction::to_starknet_core_tx;
@@ -33,7 +33,7 @@ use sc_client_api::BlockBackend;
 use sc_network_sync::SyncingService;
 use sc_transaction_pool::{ChainApi, Pool};
 use sc_transaction_pool_api::error::{Error as PoolError, IntoPoolError};
-use sc_transaction_pool_api::{TransactionPool, TransactionSource};
+use sc_transaction_pool_api::{InPoolTransaction, TransactionPool, TransactionSource};
 use sp_api::ProvideRuntimeApi;
 use sp_arithmetic::traits::UniqueSaturatedInto;
 use sp_blockchain::HeaderBackend;
@@ -51,8 +51,8 @@ use starknet_core::types::{
     DeployAccountTransactionResult, EventFilterWithPage, EventsPage, ExecutionResources, ExecutionResult, FeeEstimate,
     FieldElement, FunctionCall, Hash256, InvokeTransactionReceipt, InvokeTransactionResult,
     L1HandlerTransactionReceipt, MaybePendingBlockWithTxHashes, MaybePendingBlockWithTxs,
-    MaybePendingTransactionReceipt, MsgFromL1, StateDiff, StateUpdate, SyncStatus, SyncStatusType, Transaction,
-    TransactionExecutionStatus, TransactionFinalityStatus, TransactionReceipt,
+    MaybePendingTransactionReceipt, MsgFromL1, PendingBlockWithTxHashes, StateDiff, StateUpdate, SyncStatus,
+    SyncStatusType, Transaction, TransactionExecutionStatus, TransactionFinalityStatus, TransactionReceipt,
 };
 use starknet_core::utils::get_selector_from_name;
 
@@ -890,6 +890,12 @@ where
     /// a pending block with transaction hashes, depending on the state of the requested block.
     /// In case the block is not found, returns a `StarknetRpcApiError` with `BlockNotFound`.
     fn get_block_with_tx_hashes(&self, block_id: BlockId) -> RpcResult<MaybePendingBlockWithTxHashes> {
+        if is_pending_block(block_id) {
+            let pending_block =
+                prepare_pending_block_with_tx_hashes::<B, C, P, H>(self.client.clone(), self.pool.clone())?;
+            return Ok(MaybePendingBlockWithTxHashes::PendingBlock(pending_block));
+        }
+
         let substrate_block_hash = self.substrate_block_hash_from_starknet_block(block_id).map_err(|e| {
             error!("'{e}'");
             StarknetRpcApiError::BlockNotFound
@@ -1579,7 +1585,7 @@ where
             }
         }
 
-        // TODO: use actual execution ressources
+        // TODO: use actual execution resources
         let receipt = match transaction {
             mp_transactions::Transaction::Declare(_) => TransactionReceipt::Declare(DeclareTransactionReceipt {
                 transaction_hash,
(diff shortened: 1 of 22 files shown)


crates/client/rpc/src/lib.rs
```diff
@@ -1580,47 +1645,24 @@ where
...
-            starknet_core::types::MsgToL1 {
-                from_address: Felt252Wrapper::from(message.from_address).0,
-                to_address: FieldElement::from_bytes_be(&to_address).unwrap(),
-                payload: message.payload.0.into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
-            }
-        }
+        let messages_sent = messages.into_iter().map(message_conversion).collect();
+
+        // TODO: use actual execution resources
+        let execution_resources = get_execution_resources();
```
tdelabro, line 1653: You don't want to do in this PR?

crates/client/rpc/src/lib.rs
```diff
@@ -1714,43 +1872,60 @@ where
...
+        pedersen_builtin_applications: 0,
+        poseidon_builtin_applications: 0,
+        ec_op_builtin_applications: 0,
+        ecdsa_builtin_applications: 0,
+        bitwise_builtin_applications: 0,
+        keccak_builtin_applications: 0,
+    }
+}
+
+fn translate_revert_error(revert_error: Option<String>) -> ExecutionResult {
```
tdelabro, line 1926: `revert_error_to_execution_result`

crates/client/rpc/src/lib.rs
```diff
@@ -1714,43 +1872,60 @@ where
...
+    starknet_core::types::Event {
+        from_address: Felt252Wrapper::from(event.from_address).0,
+        keys: event.content.keys.into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
+        data: event.content.data.0.into_iter().map(|felt| Felt252Wrapper::from(felt).0).collect(),
     }
 }
 
-fn h256_to_felt(h256: H256) -> Result<FieldElement, Felt252WrapperError> {
-    Felt252Wrapper::try_from(h256).map(|f| f.0)
+fn message_conversion(message: starknet_api::transaction::MessageToL1) -> starknet_core::types::MsgToL1 {
```
tdelabro, line 1902: `starknet_api_to_starknet_core_message_to_l1`

crates/client/rpc/src/lib.rs
```diff
@@ -1714,43 +1872,60 @@ where
...
+fn is_pending_block(block_id: BlockId) -> bool {
+    block_id == BlockId::Tag(BlockTag::Pending)
+}
+
+fn calculate_l1_message_hash(_tx: &HandleL1MessageTransaction) -> Hash256 {
+    // TODO(#1291): compute message hash correctly to L1HandlerTransactionReceipt
+    Hash256::from_felt(&FieldElement::default())
+}
+
+fn event_conversion(event: starknet_api::transaction::Event) -> starknet_core::types::Event {
```
tdelabro, line 1894: `starknet_api_to_starknet_core_event`

tdelabro: we are dealing with too many of those types, prefer to be as explicit as possible

crates/client/rpc/src/lib.rs
```diff
@@ -1714,43 +1872,60 @@ where
...
-            Ok(starknet_error) => Err(starknet_error.into()),
-            Err(_) => Err(StarknetRpcApiError::InternalServerError),
-        },
+fn is_pending_block(block_id: BlockId) -> bool {
+    block_id == BlockId::Tag(BlockTag::Pending)
+}
+
+fn calculate_l1_message_hash(_tx: &HandleL1MessageTransaction) -> Hash256 {
+    // TODO(#1291): compute message hash correctly to L1HandlerTransactionReceipt
+    Hash256::from_felt(&FieldElement::default())
```
tdelabro, line 1891: we have `impl ComputeTransactionHash for HandleL1MessageTransaction` in   `crates/primitives/transactions/src/compute_hash.rs`. Why not use it?

tdelabro (changes requested):
We need rpc tests.
Where you submit tx but not seal the block, call this RPC, check that returns yours tx.

crates/client/rpc/src/lib.rs
bidzyyys: Sure!

antiyro (approved):

crates/client/rpc/src/lib.rs
bidzyyys: Unfortunately, I have no idea how to get it... I can do that if you @tdelabro have any thoughts.

crates/client/rpc/src/lib.rs
```diff
@@ -1433,18 +1425,107 @@ where
...
+        &self,
+        chain_id: Felt252Wrapper,
+    ) -> Result<PendingBlockWithTxHashes, StarknetRpcApiError> {
+        let parent_hash = self.client.info().best_hash;
+        let latest_block = get_block_by_block_hash(self.client.as_ref(), parent_hash).unwrap_or_default();
+        let latest_block_header = latest_block.header();
+        let transactions = self
+            .get_pending_txs(parent_hash)?
+            .iter()
+            .map(|tx| tx.compute_hash::<H>(chain_id.0.into(), false).0)
```
apoorvsadana, line 1470: let's use the cache here if possible

crates/client/rpc/src/lib.rs
```diff
@@ -1433,18 +1425,107 @@ where
...
+        chain_id: Felt252Wrapper,
+    ) -> Result<PendingBlockWithTxs, StarknetRpcApiError> {
+        let parent_hash = self.client.info().best_hash;
+        let latest_block = get_block_by_block_hash(self.client.as_ref(), parent_hash).unwrap_or_default();
+        let latest_block_header = latest_block.header();
+
+        let transactions = self
+            .get_pending_txs(parent_hash)?
+            .iter()
+            .map(|tx| to_starknet_core_tx(tx.clone(), tx.compute_hash::<H>(chain_id.0.into(), false).0))
```
apoorvsadana, line 1495: same as above

apoorvsadana (commented):
Amazing job on the refactor! Just added a small comment.

crates/client/rpc/src/lib.rs
```diff
@@ -1237,7 +1211,25 @@ where
...
+    fn get_state_update(&self, block_id: BlockId) -> RpcResult<MaybePendingStateUpdate> {
+        if is_pending_block(block_id) {
+            let state_diff = StateDiff {
+                storage_diffs: Vec::new(),
+                deprecated_declared_classes: Vec::new(),
+                declared_classes: Vec::new(),
+                deployed_contracts: Vec::new(),
+                replaced_classes: Vec::new(),
+                nonces: Vec::new(),
+            };
```
tdelabro, line 1223: So we return empty state diff in case of pending state?
I don't think this is the expected behavior. We should return the state changes generated by this set of pending txs

crates/client/rpc/src/lib.rs
```diff
@@ -1694,39 +1815,58 @@ where
...
+        // Missing param
+        memory_holes: None,
+        range_check_builtin_applications: *resources.get("range_check_builtin").unwrap_or(&0),
+        pedersen_builtin_applications: *resources.get("pedersen_builtin").unwrap_or(&0),
+        poseidon_builtin_applications: *resources.get("poseidon_builtin").unwrap_or(&0),
+        ec_op_builtin_applications: *resources.get("ec_op_builtin").unwrap_or(&0),
+        ecdsa_builtin_applications: *resources.get("ecdsa_builtin").unwrap_or(&0),
+        bitwise_builtin_applications: *resources.get("bitwise_builtin").unwrap_or(&0),
+        // Missing param
+        keccak_builtin_applications: 0,
```
tdelabro, line 1870: How is it missing? Can't you do:
```rust
bitwise_builtin_applications: *resources.get("keccak_builtin").unwrap_or(&0),
```
?

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
tdelabro: @apoorvsadana @antiyro to be honest I don't even think this one is doable this way.
We could pre-execute the ready tx and gather the storage changes, but we can't guarantee they will be executed in the actual order the sequencer will chose later when producing the block

crates/client/rpc/src/lib.rs
bidzyyys: I can but I see no point in doing this in that way because this param is missing here:
```rust
pub static VM_RESOURCE_FEE_COSTS: [(&str, FixedU128); 7] = [
    ("n_steps", FixedU128::from_inner(10_000_000_000_000_000)),
    ("pedersen_builtin", FixedU128::from_inner(320_000_000_000_000_000)),
    ("range_check_builtin", FixedU128::from_inner(160_000_000_000_000_000)),
    ("ecdsa_builtin", FixedU128::from_inner(20_480_000_000_000_000_000)),
    ("bitwise_builtin", FixedU128::from_inner(640_000_000_000_000_000)),
    ("poseidon_builtin", FixedU128::from_inner(320_000_000_000_000_000)),
    ("ec_op_builtin", FixedU128::from_inner(10_240_000_000_000_000_000)),
];

```

crates/client/rpc/src/lib.rs
bidzyyys: So do we have any consensus how to get it properly?

Regarding `pre-execution` is there any way that I can get state diff from all the pending txs?

crates/client/rpc/src/lib.rs
tdelabro: In the future we will use the blockifier impl itself. And if they decide to add it later on and it already contains the keccak_builtin.
Right now the RPC lib we use is in the future in comparison to the blockifier one. But it will be fixed later and adding it will make the update frictionless

crates/client/rpc/src/lib.rs
tdelabro: > Regarding pre-execution is there any way that I can get state diff from all the pending txs?

Yep, by executing them with the runtime API, we will be able to reuse this logic: https://github.com/keep-starknet-strange/madara/pull/1411

crates/client/rpc/src/lib.rs
bidzyyys: To be decided above.

crates/client/rpc/src/lib.rs
tdelabro: Yeah, maybe it's prematurate to store the hash of tx that may never be actually executed.
Even if they are ready, they may end up being invalid and not making it into the block (for old tx version, the new one will just be reverted but there)

tdelabro (approved):

PR #1440, opened 2024-02-07
Feat/felt252 wrapper transparent

edisontim:
We cannot make any assumption regarding the layout of a wrapper type, hence we should add the `#[repr(transparent)]` attribute to the `Felt252Wrapper` type and consistently check that the invariants are respected.

`Felt252Wrapper` could possibly be unaligned with `FieldElement`

## What is the new behavior?

`Felt252Wrapper` is aligned with `FieldElement`

No


.github/workflows/starknet-rpc-tests.yml
```diff
@@ -37,8 +37,10 @@ jobs:
           MADARA_RUN_PID=$!
           while ! echo exit | nc localhost 9944; do sleep 1; done
           cd starknet-rpc-test
-          cargo test
-          kill $MADARA_RUN_PID
+          ls ./contracts/
+          ls ./contracts/counter5
+          # cargo test
+          # kill $MADARA_RUN_PID
```
tdelabro, line 43: What is going on here?

tdelabro (changes requested):

.github/workflows/starknet-rpc-tests.yml
edisontim: Was looking at what was happening with the CI

tdelabro (approved):

PR #1446, opened 2024-02-12
feat: move StarkEvents to storage 

tonypony220:
# Pull Request type

- Feature

We store Starknet events in Substrate events which is not pretty convenient and infate [block size](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/system/src/lib.rs#L627)

Resolves: #NA

## What is the new behavior?

- `TxEvents` are not deleted after storing block, so that it's main storage for starknet events
- `StarkEvent`s does not stored in Substrate events

No

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -986,7 +986,7 @@ impl<T: Config> Pallet<T> {
         BlockHash::<T>::insert(block_number, blockhash);
 
         // Kill pending storage.
-        // There is no need to kill `TxEvents` as we use them.
+        // There is no need to kill `TxEvents` as we store them.
         Pending::<T>::kill();
         PendingHashes::<T>::kill();
 
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -4,7 +4,7 @@
 //! The code consists of the following sections:
 //! 1. Config: The trait Config is defined, which is used to configure the pallet by specifying the
 //! parameters and types on which it depends. The trait also includes associated types for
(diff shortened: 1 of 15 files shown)


tonypony220:
Should we store in [Box](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/system/src/lib.rs#L633) as well?

tdelabro:
> Should we store in [Box](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/frame/system/src/lib.rs#L633) as well?

I don't think so

crates/pallets/starknet/src/lib.rs
```diff
@@ -409,11 +409,9 @@ pub mod pallet {
     /// EVENTS
     /// See: `<https://docs.substrate.io/main-docs/build/events-errors/>`
     #[pallet::event]
```
tdelabro, line 411: Can you remove the whole Event section? If it is not used at all

crates/pallets/starknet/src/lib.rs
```diff
@@ -988,7 +986,7 @@ impl<T: Config> Pallet<T> {
         BlockHash::<T>::insert(block_number, blockhash);
 
         // Kill pending storage.
-        // There is no need to kill `TxEvents` as we used `take` while iterating over it.
+        // There is no need to kill `TxEvents` as we store them.
```
tdelabro, line 989: Remove the whole comment :)

crates/pallets/starknet/src/lib.rs
tonypony220: done

tdelabro:
It's weird. The CI pass when I run it in local!

PR #1449, opened 2024-02-14
ci: add gomu gomu no gatling perfomrance test

isavov:
# Pull Request type
- Build-related changes

Currently there are no performance tests run on PR checks

Resolves: #NA

## What is the new behavior?

This PR adds a `Gomu Gomu No Gatling` benchmark on PRs

- a new task is created that runs an instance of the current `madara` code and tests it with `gomu gomu no gatling`
- this task is run as part of the PR checks

No

### DIFF
--- a/.github/workflows/gomu-gomu-tests.yml
+++ b/.github/workflows/gomu-gomu-tests.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Run gomu gomu test
         run: |-
           set -v -x +e
-          ./target/release/madara --dev --pool-limit=100000 --pool-kbytes=500000 &
+          ./target/release/madara --dev --pool-limit=100000 --pool-kbytes=500000  > madara.log 2>&1 &
           MADARA_RUN_PID=$!
           while ! echo exit | nc localhost 9944; do sleep 1; done
           pwd
--- a/.github/workflows/gomu-gomu-tests.yml
+++ b/.github/workflows/gomu-gomu-tests.yml
@@ -1,0 +1,54 @@
+name: Gomu Gomu no Gatling Performance Tests
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      BINARY_PATH: ../target/release/madara
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: |
+            target/release/madara
+          key:
+            ${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}-${{
+            github.run_id }}
+          fail-on-cache-miss: true
+
+      - name: Setup rust toolchain
+        run: |
+          rustup target add x86_64-unknown-linux-gnu
+
+      - name: Setup build deps (linux)
+        run: |
+          sudo apt update
+          sudo apt install -y clang llvm libudev-dev protobuf-compiler libssl-dev pkg-config
+      - name: Setup gomu gomu
+        run: |
+          set -v -x +e
+          cd ..
+          git clone https://github.com/keep-starknet-strange/gomu-gomu-no-gatling.git
+          cd gomu-gomu-no-gatling
+          pwd
+          cargo install --path .
+
+      - name: Setup dev chain
+        run: |
+          cd ..
+          cd madara
+          ./target/release/madara setup --chain=dev --from-local=configs
+      - name: Run gomu gomu test
+        run: |-
+          set -v -x +e
+          ./target/release/madara --dev --pool-limit=100000 --pool-kbytes=500000 &
+          MADARA_RUN_PID=$!
+          while ! echo exit | nc localhost 9944; do sleep 1; done
+          pwd
+          cd ../gomu-gomu-no-gatling
+          RUST_LOG=info cargo run -- shoot -c config/default.yaml
+          kill $MADARA_RUN_PID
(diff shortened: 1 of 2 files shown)


tdelabro (approved):

anshalshukla (approved):

.github/workflows/gomu-gomu-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          sudo apt update
+          sudo apt install -y clang llvm libudev-dev protobuf-compiler libssl-dev pkg-config
+      - name: Setup gomu gomu
+        run: |
+          set -v -x +e
+          cd ..
+          git clone https://github.com/keep-starknet-strange/gomu-gomu-no-gatling.git
+          cd gomu-gomu-no-gatling
+          cargo install --path .
```
d-roak, line 37: Shouldn't it be a separate action for gomu gomu?

.github/workflows/pull-request.yml
```diff
@@ -54,3 +59,8 @@ jobs:
       SCARB_VERSION: "2.3.1"
     steps:
       - uses: keep-starknet-strange/starknet-foundry-compatibility-tests@main
+
+  gomu-gomu-tests:
```
d-roak, line 63: should also be added to push on main? @tdelabro

.github/workflows/gomu-gomu-tests.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Setup dev chain
+        run: |
+          cd ../madara
+          ./target/release/madara setup --chain=dev --from-local=configs
+      - name: Run gomu gomu test
+        run: |-
+          set -v -x +e
+          ./target/release/madara --dev --cache --pool-limit=100000 --pool-kbytes=500000  > madara.log 2>&1 &
```
d-roak, line 46: why are we limiting the pool?

d-roak:
The results are not easily readable. Can we add them as a step summary?
e.g. https://github.com/kkrt-labs/kakarot/actions/runs/7905359595

.github/workflows/gomu-gomu-tests.yml
isavov: I thought those were standard settings, I got them from here
https://github.com/keep-starknet-strange/madara/blob/main/scripts/run_node.sh#L6
What should the correct flags be?

isavov:
> The results are not easily readable. Can we add them as a step summary? e.g. https://github.com/kkrt-labs/kakarot/actions/runs/7905359595

Yes I agree with you. Step summary will be integrated in a separate PR once gomu-gomu implements better reports output as the current output from it is not finished and getting it displayed in the step summary would require a lot of processing/cleanup that will be obsolete when the updates to reporting get done.

.github/workflows/gomu-gomu-tests.yml
apoorvsadana: we are actually increasing the limit @d-roak, the default one is 8192 iirc

apoorvsadana: i don't know if there's an option to not have a limit if that's what you meant

d-roak: You can run madara without using those flags AFAIK (unless something changed past months). Just asking what's the motivation behind using them instead of just the `--dev` and keep the substrate default. Why do we need more than 8192 to run gomu gomu?
If it's a requirement for madara itself, should be changed on our cli instead of changing it with flags

.github/workflows/pull-request.yml
d-roak: @isavov can you add it in the push workflow also, please?

d-roak: ohh it's already done

.github/workflows/gomu-gomu-tests.yml
apoorvsadana: There was a case during benchmarking where we were firing a lot of txns and the mempool was getting filled. In that case, the node rejects transactions immediately which is isn't ideal for the gomu gomu. It might not be needed for the default case of gomu gomu. However, I don't see the issue in having it here because if we changes happen in Madara/the default case, we don't want txns to be rejected because of the mempool being filled.

d-roak: I don't have any issues with having this here. The only thing I usually dislike and am against is using fine-tuned parameters when they are not required.

> There was a case during benchmarking where we were firing a lot of txns and the mempool was getting filled. In that case, the node rejects transactions immediately which is isn't ideal for the gomu gomu

This needs to be fixed, in a real-world scenario we need to properly handle these cases. Adding the flags to make it work, just makes it harder for us to detect such issues

apoorvsadana: > This needs to be fixed, in a real-world scenario we need to properly handle these cases. Adding the flags to make it work, just makes it harder for us to detect such issues

But immediately dropping the transaction is the ideal way to handle it for a node right? We can have queue and retry services that keep trying to make sure it eventually gets in but that's more of a job for RPC providers and not the node itself. If the node reaches the max cap at the mempool, it should not accept anymore transactions.

apoorvsadana: > I don't have any issues with having this here. The only thing I usually dislike and am against is using fine-tuned parameters when they are not required.

Fair enough, we can remove this parameter if it's not needed for now then @isavov. Let's just replace it with a comment on the gomu gomu config that if the number is increased, we should increase the pool limit accordingly so we don't lose this knowledge.

isavov: Just pushed an update and it appears that the custom flags are not currently needed with the default gomu gomu config.

apoorvsadana (approved):

PR #1451, opened 2024-02-17
add: from_address in l1 message calldata

anshalshukla:
Cairo contracts try to fetch from_address from the 0th index of call data, this was noticed while trying to integrate starkgate and was later confirmed on `Starknet Cairo Core Stars - Pharaohs` group 

- Bugfix

Currently the calldata directly parsed from the event in L1 without any modification.

Resolves: #NA

## What is the new behavior?

Adds from_data at 0th index of calldata


apoorvsadana (approved):

crates/client/l1-messages/src/contract.rs
```diff
@@ -32,12 +38,15 @@ pub fn parse_handle_l1_message_transaction(
         .try_into()
         .map_err(L1EventToTransactionError::InvalidNonce)?;
 
-    let calldata: Vec<Felt252Wrapper> = event
-        .payload
-        .iter()
-        .map(|param| Felt252Wrapper::try_from(sp_core::U256(param.0)))
-        .collect::<Result<Vec<Felt252Wrapper>, Felt252WrapperError>>()
-        .map_err(L1EventToTransactionError::InvalidCalldata)?;
+    let mut calldata: Vec<Felt252Wrapper> = vec![from_address];
```
tdelabro, line 41: you can preallocate the correct length there: event.payload.len() + 1

tdelabro: and use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect_into to populate the allocated vec

tdelabro (changes requested):

tdelabro (approved):

PR #1482, opened 2024-02-28
creating a local testnet with multiple nodes fails using only flags

### LINKED ISSUE #1468: bug: creating a local testnet with multiple nodes fails using only flags
# Bug Report

**Madara version:** v0.7.0

**Current behavior:** When trying to bootstrap a local testnet with multiple nodes, the nodes fail to connect with the validator node because of a genesis mismatch. Tested with chains `dev` and `local`.
image

**Expected behavior:** We should be able to run a local testnet while following the steps of this: https://docs.substrate.io/tutorials/build-a-blockchain/simulate-network/

**Steps to reproduce:** Check image in current behavior and the docs on the expected behavior

**Related code:** N/A

**Other information:** An alternative solution that works is creating a raw chainspec and pass it with the `--chain` flag. Another weird message we are getting is a genesis mismatch when passing a non-raw chainspec (the file is the same, so the genesis shouldn't mismatch)

hhamud:
# Pull Request type
* Closes #1468 

- Bugfix

Current behaviour is that genesis configuration hash changes upon each node instance. Therefore it is not possible to build a local testnet with multiple notes using only the cli flags. 

I am not sure why the chainspec is not deterministic, could it be the wasm runtime?.

Resolves: #NA

## What is the new behavior?
New behaviour is that it is possible to create a local testnet with multiple nodes using only cli flags as it just spits out the chainspec configuration into a json file and uses the same one.

No


scripts/run_testnet.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env sh
+
+# Function to kill background processes
+cleanup() {
```
apoorvsadana, line 4: is this not being called anywhere?

scripts/run_testnet.sh
```diff
@@ -0,0 +1,59 @@
...
+    find . -type f -name "validator_output.txt" -delete
+    find . -type f -name "node_output.txt" -delete
+    kill "$validator_pid" "$node_pid" 2>/dev/null
+}
+
+# build release
+cargo build --release
+
+# download configs
+./target/release/madara setup --from-remote
```
apoorvsadana, line 15: we can skip this if we're using --from-local later on

apoorvsadana (commented):
interesting, you mean writing it to a JSON file and reading it back makes it work fine? can you debug the differences between in the spec in both the cases, this seems to be a hacky solution for now but its definitely progress and has some insights

scripts/run_testnet.sh
hhamud: good point I forgot to call it.

hhamud:
@apoorvsadana  @tdelabro can we merge this?

scripts/run_testnet.sh
```diff
@@ -0,0 +1,59 @@
...
+    find . -type f -name "validator_output.txt" -delete
+    find . -type f -name "node_output.txt" -delete
+    kill "$validator_pid" "$node_pid" 2>/dev/null
+}
+
+# Trap EXIT signal to ensure cleanup runs on script exit
+trap cleanup EXIT
+
+# build release
+cargo build --release
```
tdelabro, line 15: In the CI there is already a build binary cached. You should use this one and no rebuild everything

crates/node/src/command.rs
```diff
@@ -10,6 +15,30 @@ use crate::constants::DEV_CHAIN_ID;
 use crate::constants::SHARINGAN_CHAIN_ID;
 use crate::{chain_spec, service};
 
+fn spit_spec_to_json(base_config: impl ChainSpec) -> Result<Box<dyn ChainSpec>, String> {
```
tdelabro, line 18: Not a big fan of this solution of writing to disk to read it again just after.
Is there a way to fix the solution in a more elegant way?
Did you manage to understand why there was incoherences between the two nodes?

tdelabro (changes requested):

scripts/run_testnet.sh
hhamud: I will keep the script as is for local testing but re-write the ci yml file instead.

crates/node/src/command.rs
hhamud: The plan was to split the issue into two different PRs as trying to find the original bug is taking longer than I originally thought that it would. so this PR fixes the initial issue #1468 while this PR #1498 will provide a better solution to this issue #1485

hhamud:
@tdelabro can you check this?

tdelabro (approved):

Issue #1485, opened 2024-02-29
bug: Genesis hash is only consistent upon using the raw chainspec file


hhamud:
# Bug Report

**Madara version:**
latest version

**Current behavior:**
Upon further investigation of #1468 and having provided a quicker fix to that issue on #1482. The main problem seems to be that the genesis hash is only consistent when using the raw chainspec file. 

**Expected behavior:**
The genesis hash should match if nodes are using the same chainspec and should be deterministic so the fault could either be within the forked subtrate crate or how we setup the chainspec. 

**Steps to reproduce:**
Run a testnest while using the `local` or `dev` chain flags.

hhamud:
its still being worked on, not stale.

hhamud:
Fixed by https://github.com/keep-starknet-strange/madara/pull/1498

Issue #1254, opened 2023-11-07
infra: Generate Terraform scripts to automate deployments of Madara based testnets 


abdelhamidbakhta:
# Description

We need an easy way to spin up and update Madara based testnet. For instance, currently we operate Sharingan testnet manually and it's painful. 

# Acceptance criteria

- Terraform scripts are working and deploy testnet correctly

# Notes

A follow up on this could be to integrate this in the CI to deploy automatically a release of Madara on Sharingan testnet.
Currently we don't care about losing data on this testnet, so we can take some shortcuts, like deleting data without backups and so on.

tdelabro:
@apoorvsadana what is your opinion on that? Did you developed it? Or an alternative solution? Is it really required?

apoorvsadana:
We don't have terraform scripts for deploying Madara. We do that with an inhouse service because we want to automate the process. And regarding DevX, I guess the madara cli should solve the development effort, maybe we can close this issue as it's not something that MUST be there on the Madara repo. The same way geth doesn't have infra deployment scripts in its repo (afaik(

tdelabro:
So I can close?

apoorvsadana:
Yes imo

PR #1498, opened 2024-03-05
Genesis hash is only consistent upon using the raw chainspec file 

hhamud:
# Pull Request type

- Bugfix

See #1485 

## What is the new behavior?
Genesis Hash does not change upon every new instance of node created. 

No


hhamud:
@tdelabro can you review this?

crates/pallets/starknet/src/tests/genesis_block.rs
apoorvsadana, line 1: let's add a comment explaining why we are doing these tests (because of inconsistencies which show up for some classes which we aren't aware of now but it breaks full node setup)

crates/pallets/starknet/src/tests/genesis_block.rs
```diff
@@ -0,0 +1,115 @@
...
+
+        let state_version = frame_system::Pallet::<MockRuntime>::runtime_version().state_version();
+        <MockRuntime as frame_system::Config>::Hash::decode(&mut &sp_io::storage::root(state_version)[..]).unwrap()
+    });
+
+    assert_eq!(b, t)
+}
+
+#[test]
+fn check_genesis_child_root() {
```
apoorvsadana, line 70: how's this test different than the storage root one?

crates/pallets/starknet/src/tests/genesis_block.rs
```diff
@@ -0,0 +1,115 @@
...
+        System::finalize();
+        let mut st: &[u8] = &[0u8; 32];
+        <MockRuntime as frame_system::Config>::Hash::decode(&mut st).unwrap()
+    });
+
+    assert_eq!(b, t)
+}
+
+#[test]
+fn check_genesis_storage() {
```
apoorvsadana, line 109: do we need this?

EvolveArt (approved):

EvolveArt:
We should probably investigate zombienet tho to make this more resilient and easily test with different numbers of validators/peers
cc https://github.com/paritytech/subxt/tree/master/examples/parachain-example

hhamud:
> We should probably investigate zombienet tho to make this more resilient and easily test with different numbers of validators/peers cc https://github.com/paritytech/subxt/tree/master/examples/parachain-example

I'll keep this in mind if the problem re-surfaces again. Also will that tool also help with this? https://github.com/keep-starknet-strange/madara/issues/1254

EvolveArt:
> > We should probably investigate zombienet tho to make this more resilient and easily test with different numbers of validators/peers cc https://github.com/paritytech/subxt/tree/master/examples/parachain-example
> 
> I'll keep this in mind if the problem re-surfaces again. Also will that tool also help with this? #1254

Not really no its only for local testing

PR #1500, opened 2024-03-06
chore: remove use of dispatchError in RuntimeAPI

tdelabro:
The use of `DispatchError` as error type is only required for the extrinsics. For the RuntimeAPI we can use whatever type we want (as long as it implements scale and type-info). 
Using such a custom type allow us to return richer information and participate into our will to move away from substrate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- chore(runtime_api): remove the use of `DispatchError`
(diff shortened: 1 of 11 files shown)


hhamud:
@tdelabro I can incorporate most of these changes into my PR as I am refactoring the error messaging in the rpc in #1555

tdelabro:
superseeded by #1555

PR #1522, opened 2024-03-19
Add tests for storage reversal for failed transactions

### LINKED ISSUE #1520: feat: test cases to check contract storage reverts for failed transactions
Currently, the Madara storage doesn't revert storage changes in case a smart contract transaction fails. I am referring to execution failures here, for example, you add a panic! inside the function call of a smart contract. Not other failures like missing fees etc. This should mostly be fixed in a few days. However, we need a good test case for this to ensure we don't miss this later on.

swetshaw:
# Tests

## Fixes
Fixes #1520 

- No check for storage updates for failed transactions

## What is the new behavior?
- Check if storage updates are reverted for failed transactions

No


crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +525,97 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+        );
+
+        // log account address
+        log::info!("Account address: {:?}", account_addr);
+
+        let deploy_transaction = InvokeTransactionV1{
+            sender_address: account_addr.into(),
+            signature: vec![],
+            nonce: Felt252Wrapper::ONE,
+            calldata: vec![
```
swetshaw, line 606: Hey @apoorvsadana , getting error (ref - screenshot) while running the deploy transaction here. 

Screenshot 2024-03-23 at 12 43 17 AM

I have an intuition that this is happening because of calldata. Could you please have a look?
The calldata is inspired from the network log for contract deployment transaction here - Screenshot 2024-03-23 at 12 47 35 AM

apoorvsadana: Can you try making the nonce as 2? Because the nonce was 1 for the declare txn

Also, are you sure this error message is for this test? Because the account you're using isn't an Argent account. The normal tests can have error messages if they are checking for failures so just confirming you're looking at the right one. You can test your specific case by specifying the test name after cargo test

swetshaw: - I had tried with nonce as 2, but it is still failing. 
- Yes you are right that error was getting logged from a different test case, my bad there.

Below is the error that is being thrown from this test. In summary the transaction execution is failing. I am trying to figure out the correct calldata for deploy transaction because only then the tx will go through. If you have any leads here please let me know

Screenshot 2024-03-23 at 3 53 50 PM

apoorvsadana: from the image you shared, the ngrok shows 8 elements in the calldata but your test case has only 7?

swetshaw: Yeah I missed adding the point that -

I deployed the same contract multiple times and according to my observation:
- All the values in the calldata array remains the same except for the 6th (5th index: calldata[5]) element. 
- This value changes for every transaction and I am unable to figure out what that could be

cairo-contracts/build/TransactionRevert.json
apoorvsadana, line 1: we can use the abi from the sierra file itself

cairo-contracts/src/cairo_1/storage/transactionRevert.cairo
apoorvsadana, line 1: let's rename the storage folder to test_contracts. let's also rename to `TransactionRevert.cairo`

cairo-contracts/src/cairo_1/storage/transactionRevert.cairo
```diff
@@ -0,0 +1,20 @@
+#[starknet::contract]
+mod HelloStarknet {
+    #[storage]
+    struct Storage {
+        balance: felt252, 
+    }
+
+    // Increases the balance by the given amount.
```
apoorvsadana, line 8: change the comment to panics

crates/pallets/starknet/Cargo.toml
```diff
@@ -92,6 +92,8 @@ hexlit = "0.5.5"
 assert_matches = "1.5.0"
 starknet-ff = { workspace = true }
 project-root = "0.2.2"
+env_logger = "*"
+test-log = "0.2.15"
```
apoorvsadana, line 96: do we still need this?

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -2,29 +2,36 @@ use blockifier::abi::abi_utils::get_storage_var_address;
...
-use super::utils::sign_message_hash;
+use super::utils::{get_contract_class, sign_message_hash};
+use crate::tests::utils::get_udc_deployed_contract_address;
 use crate::tests::{
     get_invoke_argent_dummy, get_invoke_braavos_dummy, get_invoke_dummy, get_invoke_emit_event_dummy,
     get_invoke_nonce_dummy, get_invoke_openzeppelin_dummy, get_storage_read_write_dummy, set_nonce,
 };
 use crate::{Call, Config, Error, StorageView};
 
+const INCREASE_BALANCE_FUNCTION_SELECTOR: &str = "0x362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320";
```
apoorvsadana, line 33: it's specific to a test case so we can move this constant there itself.

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -238,6 +245,7 @@ fn given_hardcoded_contract_run_storage_read_and_write_it_works() {
 
 #[test]
 fn test_verify_nonce() {
+    log::info!("Running contract_function_call_should_fail test");
```
apoorvsadana, line 248: let's remove this

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +533,144 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+        set_nonce::<MockRuntime>(&tx_sender, &Nonce(StarkFelt::from(1u64)));
+
+        assert_eq!(
+            Starknet::validate_unsigned(tx_source, &call),
+            Err(TransactionValidityError::Invalid(InvalidTransaction::Stale))
+        );
+
+        let storage_key = (tx_sender, StorageKey(PatriciaKey(StarkFelt::from(0u128))));
+
+        assert_eq!(Starknet::storage(storage_key), StarkFelt::from(0u128));
```
apoorvsadana, line 559: I suppose you probably want to check Starknet::nonces here?

apoorvsadana: if yes, then the test case should be `test_tx_fail_on_invalid_nonce_should_not_increment_nonce`

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +533,144 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+                Felt252Wrapper::from_hex_be("0x1").unwrap(),
+                Felt252Wrapper::from_hex_be("0xa").unwrap(),
+            ],
+        };
+
+        // the transaction reverts and returns Ok
+        assert_ok!(Starknet::invoke(RuntimeOrigin::none(), increase_balance_tx.clone().into()));
+
+        let storage_var_selector = StorageKey(PatriciaKey(
+            StarkFelt::try_from("0x0206f38f7e4f15e87567361213c28f235cccdaa1d7fd34c9db1dfe9489c6a091").unwrap(),
```
apoorvsadana, line 669: can we call the get_balance method instead so we're sure we're reading it correctly? because if the storage key is some random hex it will still return 0. also, as far as I remember, this bug exists in Madara today so this test case should actually fail.

swetshaw: Sure, will call the get_balance method. But the storage key is not just random hex -- it is the variable selector. The hex translates to `balance` in string. 

For better assertion, will add the get_balance method

swetshaw: Yes this test actually fails for current madara implementation.

cairo-contracts/build/TransactionRevert.json
swetshaw: removing the abi from here

cairo-contracts/src/cairo_1/storage/transactionRevert.cairo
swetshaw: done

cairo-contracts/src/cairo_1/storage/transactionRevert.cairo
swetshaw: done

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: done

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: done

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: hey this was a dummy test - I had assumed that even if transaction fails due to invalid nonce the storage should get reverted. But in practice that is not the case, in case of invalid nonce the transaction gets `failed` and not `reverted`

This test is not needed so removing it

crates/pallets/starknet/Cargo.toml
swetshaw: I don't think. Will remove

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: yet to add the get_balance function call. Pushing the changes in sometime

swetshaw: This is done

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +529,124 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+                .unwrap(),
+                udc_contract_address: FieldElement::from_hex_be(UDC_ADDRESS).unwrap(),
+            }),
+            &[],
+        );
+
+        // deploy contract
+        assert_ok!(Starknet::invoke(RuntimeOrigin::none(), deploy_transaction.into()));
+
+        let increase_balance_function_selector: &str = "0x362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320";
```
apoorvsadana, line 619: we can still keep it a const

apoorvsadana (approved):
Good job, thanks for this contribution!

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +529,124 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+        };
+
+        // the transaction reverts and returns Ok
+        assert_ok!(Starknet::invoke(RuntimeOrigin::none(), increase_balance_tx.clone().into()));
+
+        // the storage value should be 0 after the transaction reverts
+        let contract_address = ContractAddress(PatriciaKey(StarkFelt::try_from(contract_address).unwrap()));
+
+        let get_balance_function_selector = EntryPointSelector(
+            StarkFelt::try_from("0x039e11d48192e4333233c7eb19d10ad67c362bb28580c604d67884c85da39695").unwrap(),
```
apoorvsadana, line 644: starknet.rs has `get_selector_from_name` which can be used here and at other places where the selector is hardcoded

swetshaw:
> Good job, thanks for this contribution!

Thank you. Enjoyed it :)

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: Done

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: Done, using `get_selector_from_name`

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +529,127 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+        let get_balance_function_selector = get_selector_from_name("get_balance").unwrap();
+
+        let get_balance_function_selector_entrypoint =
+            EntryPointSelector(StarkFelt::try_from(get_balance_function_selector).unwrap());
+
+        let default_calldata = Calldata(Default::default());
+
+        let res = Starknet::call_contract(contract_address, get_balance_function_selector_entrypoint, default_calldata)
+            .unwrap();
+        assert_eq!(res, vec![Felt252Wrapper::from_hex_be("0x0").unwrap()])
```
EvolveArt, line 653:
```suggestion
        assert_eq!(res, vec![Felt252Wrapper::ZERO])
```

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +529,127 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+        let contract_address = ContractAddress(PatriciaKey(StarkFelt::try_from(contract_address).unwrap()));
+
+        let get_balance_function_selector = get_selector_from_name("get_balance").unwrap();
+
+        let get_balance_function_selector_entrypoint =
+            EntryPointSelector(StarkFelt::try_from(get_balance_function_selector).unwrap());
+
+        let default_calldata = Calldata(Default::default());
+
+        let res = Starknet::call_contract(contract_address, get_balance_function_selector_entrypoint, default_calldata)
```
EvolveArt, line 651:
```suggestion
        let balance_value = Starknet::call_contract(contract_address, get_balance_function_selector_entrypoint, default_calldata)
```

crates/pallets/starknet/src/tests/utils.rs
```diff
@@ -97,3 +98,12 @@ pub fn build_get_balance_contract_call(account_address: StarkFelt) -> (EntryPoin
 
     (balance_of_selector, calldata)
 }
+
+pub fn get_udc_deployed_contract_address(
```
EvolveArt, line 102: not really useful no?

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -525,3 +529,127 @@ fn test_verify_nonce_in_unsigned_tx() {
...
+            TransactionSource::InBlock,
+            &crate::Call::invoke { transaction: deploy_transaction.clone().into() },
+        ));
+
+        let contract_address: FieldElement = get_udc_deployed_contract_address(
+            salt.into(),
+            transaction_revert_class_hash.into(),
+            &UdcUniqueness::Unique(UdcUniqueSettings {
+                deployer_address: FieldElement::from_hex_be(
+                    "0x0642a8b9e2c6cc3a9ddb84575123f262a21415f78db453b0625d889e1e06ac32",
```
EvolveArt, line 609:
```suggestion
                    account_addr,
```

EvolveArt (approved):
smol changes but lgtm

crates/pallets/starknet/src/tests/invoke_tx.rs
swetshaw: I had tried using `account_adr` instead of hardcoding the value. But since `account_adr` is of type `ContractAddress` and `deployer_address` expects value of type `FieldElement`, I am not sure how to convert ContractAddress to FieldElement. Need help here @EvolveArt

apoorvsadana: Can you try account_adr.into(()

swetshaw: account_addr.into() doesn't work, it says the following error:

Screenshot 2024-04-04 at 12 21 07 AM

apoorvsadana: My bad, can you try account_adr.0.into()

swetshaw: Thanks, this `account_addr.0.0.into()` worked

PR #1530, opened 2024-03-23
Custom block import stage that verifies Sierra -> CASM class integrity for declare transactions

### LINKED ISSUE #789: feat: Check `compiled_class_hash` inside the pallet
# Feature Request

**Describe the Feature Request**

Currently, the`compiled_class_hash` is checked in the client in declare v2. This creates a vulnerability because a node can choose not to verify the `compiled_class_hash` and accept transactions. Other nodes will not be able to check this as there is no check in the runtime. 

Right now, the sierra compiler doesn't support `no_std` because of which it's not possible to have this check in the runtime. However, `no_std` for the compiler is in the roadmap and we must move this check to the runtime whenever that is implemented.

**Describe Preferred Solution**

I think we should implement this check `validate_unsigned` but open to discuss.

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Yes

m-kus:
This PR introduces a custom block import stage in the queue, that does additional checks before executing Grandpa consensus.
More specifically, upon every declare transaction it checks that Sierra/Casm class hashes specified in the transaction match the according Casm class (part of the extrinsic) and locally stored Sierra class.

Note that currently it only works for a single node setup, where Sierra classes DB is populated when add declare transaction RPC call is handled. For the multi-node setup we will need to query other nodes via P2P to get missing Sierra classes and finish block import.

Resolves #789


apoorvsadana:
Since this code only works for a single node because sierra classes aren’t communicated over p2p yet, does this provide any extra security at this point in time? Also, during a single node process, how does the block import work, does the validator that created the block also import it again or is it skipped?

I mainly trying to understand if there's an advantage of merging until we're sending sierra classes over p2p. It seems this would slower block production (if we need to wait for import before going to the next block) without any extra security.

crates/client/starknet-block-import/src/compilation.rs
```diff
@@ -0,0 +1,179 @@
...
+        match item {
+            MaybeRelocatable::Int(felt) => bytecode.push(Felt252Wrapper::from(felt.clone()).into()),
+            MaybeRelocatable::RelocatableValue(_) => return Err(CompilationError::UnexpectedRelocatable),
+        }
+    }
+    Ok(bytecode)
+}
+
+pub fn stark_felt_to_biguint(felt: &StarkFelt) -> BigUint {
+    BigInt::from_bytes_be(Sign::Plus, felt.bytes()).to_biguint().unwrap()
```
EvolveArt, line 131: let's handle the error here or write in comment why it's safe

EvolveArt (approved):

EvolveArt:
> > Since this code only works for a single node because sierra classes aren’t communicated over p2p yet, does this provide any extra security at this point in time?
> 
> Right, it does not provide extra security for a single node setup.
> 
> > Also, during a single node process, how does the block import work, does the validator that created the block also import it again or is it skipped?
> 
> Currently own blocks are checked too, but this can be disabled.
> 
> > I mainly trying to understand if there's an advantage of merging until we're sending sierra classes over p2p. It seems this would slower block production (if we need to wait for import before going to the next block) without any extra security.
> 
> There's no advantage except for being prepared :) We could probably make it feature enabled for now.
> 
> I will also do benchmarks to estimate the overhead.

As you mentionned let's disable this check for single node setup and put that under a feature flag for now

apoorvsadana:
> > Since this code only works for a single node because sierra classes aren’t communicated over p2p yet, does this provide any extra security at this point in time?
> 
> Right, it does not provide extra security for a single node setup.
> 
> > Also, during a single node process, how does the block import work, does the validator that created the block also import it again or is it skipped?
> 
> Currently own blocks are checked too, but this can be disabled.
> 
> > I mainly trying to understand if there's an advantage of merging until we're sending sierra classes over p2p. It seems this would slower block production (if we need to wait for import before going to the next block) without any extra security.
> 
> There's no advantage except for being prepared :) We could probably make it feature enabled for now.
> 
> I will also do benchmarks to estimate the overhead.

Yes let's move this behind a feature flag because decentralised seems a little far in the roadmap currently. Also, can we disable checking own blocks as well, is there any advantage to that?

crates/client/starknet-block-import/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    type Error = ConsensusError;
+
+    async fn check_block(&mut self, block: BlockCheckParams<Block>) -> Result<ImportResult, Self::Error> {
+        self.inner.check_block(block).await
+    }
+
+    async fn import_block(&mut self, block: BlockImportParams<Block>) -> Result<ImportResult, Self::Error> {
+        log::info!("🐺 Starknet block import: verifying declared CASM classes against local Sierra classes");
+        if let Some(extrinsics) = &block.body {
+            let prev_block_hash = *block.header.parent_hash();
```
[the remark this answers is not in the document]
apoorvsadana: understood

crates/client/starknet-block-import/src/compilation.rs
```diff
@@ -0,0 +1,179 @@
...
+    UnexpectedRelocatable,
+    #[error("Failed to parse felt from bytes: {0}")]
+    FeltFromBytes(#[from] FromByteArrayError),
+}
+
+pub(crate) fn blockifier_casm_class_to_compiled_class_hash(
+    casm_class: BlockifierCasmClass,
+) -> Result<FieldElement, CompilationError> {
+    match casm_class {
+        BlockifierCasmClass::V0(_) => Err(CompilationError::UnsupportedClassV0),
```
tdelabro, line 42: we can easily support that no?
It would be useful for full nodes working with the whole chain history like deoxys.
@antiyro wdyt?

crates/client/starknet-block-import/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+//! NOTES:
+//!
+//! Currently Sierra classes DB is populated when user submits RPC requests and therefore the
+//! approch works for single node setup only.
+//! In order to make it work in the multi-node setting, one needs to query missing Sierra classes
+//! from other nodes via P2P (see https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/class.proto)
+//!
+//! Cairo compiler version mismatch can be a problem, e.g. if the version used by Madara is lagging
+//! behind significantly it can fail to compile Sierra classes obtained from the recent compilers.
+//! Similarly, old Sierra classes might not compile because of the broken backward compatibility.
```
tdelabro, line 39: There is no other solution than to locally have all the compiler versions that have been used during the chain history. Those can only change on chain update so there won't be that many.

tdelabro (approved):

tdelabro:
Great work!! We are touching on some of the issues we have been pushing back since madara inception. That's great to see there is actually a good way to solve this.

PR #1555, opened 2024-04-06
FIx: starknet_call errs if contract nonexistent

### LINKED ISSUE #1537: bug: `starknet_call` errs if contract nonexistent
# Bug Report

**Madara version:**

Whichever version [52.7.206.208:9944](https://github.com/keep-starknet-strange/madara/discussions/553) is running

**Current behavior:**

Calling a nonexisting contract causes the API to return an internal server error instead of a user-level err msg.

**Expected behavior:**

The API should respond with a relatively descriptive err message instead of seeming like there's a bug in the server.

**Steps to reproduce:**

Executing the following:

```sh
curl -X POST --location "http://52.7.206.208:9944" \
    -H "Content-Type: application/json" \
    -d '{
          "id": 1,
          "jsonrpc": "2.0",
          "method": "starknet_call",
          "params": [
            {
              "contract_address": "0x1",
              "entry_point_selector": "0x1",
              "calldata": [
              ]
            },
            "latest"
          ]
        }'
```

Results in the following internal server error response:

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": 500,
    "message": "Internal server error"
  },
  "id": 1
}
```

Whereas executing the same request on Juno gives a descriptive error message:

```json
{
  "jsonrpc": "2.0",
  "error": {
    "code": 40,
    "message": "Contract error",
    "data": {
      "revert_error": "Requested contract address ContractAddress(PatriciaKey(StarkFelt(\"0x0000000000000000000000000000000000000000000000000000000000000001\"))) is not deployed."
    }
  },
  "id": 1
}
```

hhamud:
# Pull Request type
* Fix #1537
* Closes #1500

- Bugfix
- Feature
- Refactoring (API and functional changes)

Current behaviour is that the error messages received by users from the RPC is not detailed enough and users would prefer it if it included more information on what exactly caused the error

For example for this post request
```shell
curl -X POST --location "http://127.0.0.1:9944" \
    -H "Content-Type: application/json" \
    -d '{
          "id": 1,
          "jsonrpc": "2.0",
          "method": "starknet_call",
          "params": [
            {
              "contract_address": "0x1",
              "entry_point_selector": "0x1",
              "calldata": [
              ]
            },
            "latest"
          ]
        }'


```
You would get this response back from the RPC
```shell
{
  "jsonrpc": "2.0",
  "error": {
    "code": 40,
    "message": "Contract Error",
}
  "id": 1
}
```

Resolves: #1537

## What is the new behavior?
New api response for this post request would be

```shell
{
  "jsonrpc": "2.0",
  "error": {
    "code": 40,
    "message": "Contract Error",
    "data": {
      "revert_error": "Entry point EntryPointSelector(StarkFelt(\"0x0000000000000000000000000000000000000000000000000000000000000001\")) not found in contract."
    }
  },
  "id": 1
}
```

No


hhamud:
just waiting on this to be sorted out before continuting onwards https://github.com/paritytech/polkadot-sdk/issues/4181

hhamud:
@tdelabro @apoorvsadana Can you review this please?

crates/primitives/simulations/src/lib.rs
```diff
@@ -5,19 +5,33 @@ pub extern crate alloc;
...
 // The official rpc expect use to return the trace up to the point of failure.
 // Figuring out how to get that is a problem for later
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 #[cfg_attr(feature = "scale-info", derive(scale_info::TypeInfo))]
 #[cfg_attr(feature = "parity-scale-codec", derive(parity_scale_codec::Encode, parity_scale_codec::Decode))]
-pub struct PlaceHolderErrorTypeForFailedStarknetExecution;
+pub enum Error {
+    ContractNotFound(ContractAddress),
+    // TODO: replace string with Blockifier TransactionExecutionError
```
tdelabro, line 22: When will thins be possible?

tdelabro (approved):

crates/primitives/simulations/src/lib.rs
hhamud: When Blockifier and Cairo-vm add subtrate encode, decode and typeinfo to the error type. Right now I just wrap wrap the string formatting and pipe through to the RPC

crates/primitives/simulations/src/lib.rs
```diff
@@ -1,16 +1,37 @@
...
 // TODO: This is a placeholder
 // https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3919
 // The official rpc expect use to return the trace up to the point of failure.
 // Figuring out how to get that is a problem for later
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 #[cfg_attr(feature = "scale-info", derive(scale_info::TypeInfo))]
 #[cfg_attr(feature = "parity-scale-codec", derive(parity_scale_codec::Encode, parity_scale_codec::Decode))]
-pub struct PlaceHolderErrorTypeForFailedStarknetExecution;
+pub enum Error {
```
apoorvsadana, line 14: should we rename this to SimulationError to make it more readable when it's imported to other files?

crates/pallets/starknet/src/simulations.rs
```diff
@@ -156,54 +162,50 @@ impl<T: Config> Pallet<T> {
...
-        DispatchError,
-    > {
+    ) -> Result<Result<Vec<(TransactionExecutionInfo, CommitmentStateDiff)>, Error>, Error> {
+        let res = Self::re_execute_transactions_inner(transactions_before, transactions_to_trace);
         storage::transactional::with_transaction(|| {
-            let res = Self::re_execute_transactions_inner(transactions_before, transactions_to_trace);
             storage::TransactionOutcome::Rollback(Result::<_, DispatchError>::Ok(Ok(res)))
         })
-        .map_err(|_| Error::<T>::FailedToCreateATransactionalStorageExecution)?
+        .map_err(|_| Error::FailedToCreateATransactionalStorageExecution)?
```
apoorvsadana, line 192: will it be beneficial to catch the error here and pass it as an argument FailedToCreateATransactionalStorageExecution

crates/pallets/starknet/src/simulations.rs
```diff
@@ -98,55 +102,57 @@ impl<T: Config> Pallet<T> {
...
-            storage::TransactionOutcome::Rollback(Result::<_, DispatchError>::Ok(Self::estimate_message_fee_inner(
-                message,
-            )))
+            res = Some(Self::estimate_message_fee_inner(message));
+            storage::TransactionOutcome::Rollback(Result::<_, DispatchError>::Ok(()))
         })
-        .map_err(|_| Error::<T>::FailedToCreateATransactionalStorageExecution)?
+        .map_err(|_| Error::FailedToCreateATransactionalStorageExecution)?;
+
+        res.expect("`res` should have been set to `Some` at this point")
```
apoorvsadana, line 152: let's throw an Error instead of `expect`

crates/pallets/starknet/src/simulations.rs
```diff
@@ -45,22 +45,26 @@ impl<T: Config> Pallet<T> {
...
+                Err(e) => Err(Error::from(e)),
             })
             .map(|exec_info_res| {
                 exec_info_res.map(|exec_info| {
                     exec_info
                         .actual_resources
                         .0
                         .get("l1_gas_usage")
-                        .ok_or_else(|| DispatchError::from(Error::<T>::MissingL1GasUsage))
+                        .ok_or_else(|| DispatchError::from(ConfigError::<T>::MissingL1GasUsage))
```
apoorvsadana, line 59: should we make this an simulation error as well so its communicated to the client?

crates/client/rpc/src/trace_api.rs
```diff
@@ -99,11 +98,39 @@ where
             })?
             .map_err(|e| {
                 error!("Failed to call function: {:#?}", e);
-                StarknetRpcApiError::ContractError
+                StarknetRpcApiError::from(e)
             })?;
 
-        let simulated_transactions =
-            tx_execution_infos_to_simulated_transactions(tx_types, res).map_err(StarknetRpcApiError::from)?;
+        let mut simulated_transactions = vec![];
```
apoorvsadana, line 104: why did we remove the functions like `tx_execution_infos_to_simulated_transactions` and `blockifier_to_starknet_rs_ordered_events` and replace the code directly?

crates/client/rpc/src/errors.rs
```diff
@@ -1,67 +1,139 @@
...
+            StarknetRpcApiError::ClassAlreadyDeclared => 51,
+            StarknetRpcApiError::ValidationFailure => 55,
+            StarknetRpcApiError::UnsupportedTxVersion => 61,
+            StarknetRpcApiError::InternalServerError => 500,
+            StarknetRpcApiError::UnimplementedMethod => 501,
+            StarknetRpcApiError::ProofLimitExceeded => 10000,
+        };
+
+        let data = match &err {
+            StarknetRpcApiError::ContractError(ref error) => Some(serde_json::json!({
```
apoorvsadana, line 101: should we also implement Display for `ContractError` so we don't need to match here?

crates/primitives/simulations/src/lib.rs
tdelabro: created #1579

crates/pallets/starknet/src/simulations.rs
tdelabro: YOu don't have to use an outer variable, you can use the `with_transactional` return, can't you?

crates/client/rpc/src/trace_api.rs
tdelabro: Maybe it was the only place it was used?

crates/client/rpc/src/errors.rs
tdelabro: It's not really a display

crates/client/rpc/src/errors.rs
```diff
@@ -1,67 +1,139 @@
...
 // https://github.com/starkware-libs/starknet-specs/blob/0e859ff905795f789f1dfd6f7340cdaf5015acc8/api/starknet_write_api.json#L227
-#[derive(thiserror::Error, Clone, Copy, Debug)]
+#[derive(Error, Debug)]
 pub enum StarknetRpcApiError {
     #[error("Failed to write transaction")]
-    FailedToReceiveTxn = 1,
-    #[error("Contract not found")]
-    ContractNotFound = 20,
+    FailedToReceiveTxn,
+    #[error("Contract not found: {0}")]
```
tdelabro, line 15: This is not coherent with the specs defined here: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3912C13-L3915C14

crates/client/rpc/src/errors.rs
tdelabro: https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3948-L3963

The specs specify a different field, while the message stays `ContractError`

tdelabro (changes requested):

crates/client/rpc/src/trace_api.rs
hhamud: yep those functions were only used once.

crates/client/rpc/src/errors.rs
tdelabro: @apoorvsadana yep, the `data` field contains a json with a `revert_error` field

crates/primitives/simulations/src/lib.rs
```diff
@@ -1,16 +1,36 @@
...
 #[cfg_attr(feature = "scale-info", derive(scale_info::TypeInfo))]
 #[cfg_attr(feature = "parity-scale-codec", derive(parity_scale_codec::Encode, parity_scale_codec::Decode))]
-pub struct PlaceHolderErrorTypeForFailedStarknetExecution;
+pub enum SimulationError {
+    ContractNotFound,
+    TransactionExecutionFailed(String),
+    MissingL1GasUsage,
+    FailedToCreateATransactionalStorageExecution,
+    StateDiff,
+}
```
tdelabro, line 19: I don't think this is correct.

`FailedToCreateATransactionalStorageExecution` is an InternalError. It's our substrate code failing.
While all the others are the users input leading to the starknet protocol being unable to process the transaction.

For the first one we should `log::error` it and return InternalServerError, for the second ones, we should not log it (our code worked as expected, returning an error for an invalid input.

THose are two different levels of errors for two different levels of execution, they should be treated separately.

This is why the `simulation` code return a double error `Result<Result<TransactionExecutionInfo, SimulationError>, SimulationError>`, the outer one is on us, the inner one is on the user. They are managed differently because they are different types of errors

crates/primitives/simulations/src/lib.rs
```diff
@@ -1,16 +1,36 @@
+use blockifier::state::errors::StateError;
+use blockifier::transaction::errors::TransactionExecutionError;
 use blockifier::transaction::objects::TransactionExecutionInfo;
 use starknet_core::types::{SimulationFlag, SimulationFlagForEstimateFee};
 
 // TODO: This is a placeholder
 // https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3919
 // The official rpc expect use to return the trace up to the point of failure.
 // Figuring out how to get that is a problem for later
```
tdelabro, line 9: This comment is not up to date with the code anymore

tdelabro (changes requested):

tdelabro:
@hhamud a lil rebase and we will be good to go

crates/client/rpc/src/runtime_api.rs
```diff
@@ -43,29 +40,25 @@ where
         contract_address: ContractAddress,
         entry_point_selector: EntryPointSelector,
         calldata: Calldata,
-    ) -> RpcApiResult<Result<Vec<Felt252Wrapper>, sp_runtime::DispatchError>> {
-        self.client.runtime_api().call(best_block_hash, contract_address, entry_point_selector, calldata).map_err(|e| {
-            error!("Request parameters error: {e}");
-            StarknetRpcApiError::InternalServerError
-        })
+    ) -> RpcApiResult<Vec<Felt252Wrapper>> {
+        // simulation error
```
tdelabro, line 44: forgotten comment?

tdelabro (approved):

tdelabro (approved):

PR #1559, opened 2024-04-10
use trace instead of simulate to get txn receipt

apoorvsadana:
1. Uses trace instead of simulation for transaction receipt. Trace has re execution support.
2. Events and messages don't work in the case of pending receipts (didn't work before either)
3. 2 can be fixes right now but its easier to do when Massa labs implements pending blocks.


crates/client/rpc/src/lib.rs
```diff
@@ -1599,16 +1601,20 @@ where
 
         // TODO
         // Is any better way to get execution resources of processed tx?
-        let skip_validate = true;
-        // let parent_block_hash = Felt252Wrapper::from().into();
-        let parent_block_hash = self
-            .substrate_block_hash_from_starknet_block(BlockId::Hash(block_header.parent_block_hash.into()))
+        let parent_substrate_block_hash = self
+            .substrate_block_hash_from_starknet_block(BlockId::Hash(starknet_block.header().parent_block_hash.into()))
```
EvolveArt, line 1605:
```suggestion
            .substrate_block_hash_from_starknet_block(BlockId::Hash(block_header.parent_block_hash.into()))
```

crates/client/rpc/src/lib.rs
```diff
@@ -1599,16 +1601,20 @@ where
...
-        let simulation = self.simulate_tx(parent_block_hash, transaction.clone(), skip_validate, fee_disabled)?;
-        let execution_resources = actual_resources_to_execution_resources(simulation.actual_resources);
+        let execution_info = self.get_transaction_execution_info(
+            parent_substrate_block_hash,
+            starknet_block.transactions(),
+            chain_id,
+            transaction_hash,
+        )?;
+
+        let execution_resources = actual_resources_to_execution_resources(execution_info.actual_resources.clone());
```
EvolveArt, line 1617: need to clone?

starknet-rpc-test/Cargo.toml
```diff
@@ -120,9 +120,9 @@ path = "estimate_message_fee.rs"
 name = "starknet_simulate_transaction"
 path = "simulate_transaction.rs"
 
-[[test]]
-name = "starknet_get_state_update"
-path = "get_state_update.rs"
+#[[test]]
```
EvolveArt, line 123: why ?

crates/client/rpc/src/lib.rs
apoorvsadana: fixed

crates/client/rpc/src/lib.rs
apoorvsadana: apparently not, fixed

starknet-rpc-test/Cargo.toml
apoorvsadana: pushed it by mistake, fixed

EvolveArt (approved):

PR #1562, opened 2024-04-13
fix e2e tests workflow

apoorvsadana:


tdelabro:
entierly removed in #1580

PR #1577, opened 2024-04-26
re execute txs instead of simulate for txn receipts

apoorvsadana:


crates/client/rpc/src/lib.rs
```diff
@@ -1563,6 +1562,42 @@ where
...
+        let mut trace = self
+            .re_execute_transactions(parent_substrate_block_hash, transactions_before, transaction_to_trace)
+            .map_err(|e| {
+                log::error!("Failed to re-execute transactions: {e}");
+                StarknetRpcApiError::InternalServerError
+            })?;
+
+        let execution_info = trace.remove(0);
+
+        Ok(execution_info.0)
```
tdelabro, line 1598: `execution_info.1` are the state diffs, computed here:
crates/pallets/starknet/src/simulations.rs
```rust
let res = res.map(|r| (r, transactional_state.to_state_diff()));
``` 

We don't need them here, so it's wasted execution. Can you add a bool arg to `re_execute_transaction` runtime-api method to decide if we want the stateDiff, and change the signature so that the second ret value is an `Option`

tdelabro (changes requested):

crates/client/rpc/src/lib.rs
apoorvsadana: fixed

tdelabro (approved):

PR #1580, opened 2024-04-29
remove all da & settlement related content

tdelabro:


starknet-rpc-test/Cargo.toml
apoorvsadana, line 1: why remove this?

apoorvsadana (approved):

starknet-rpc-test/Cargo.toml
tdelabro: The way it is implemented they require the node to be run with --da-layer --da-conf

apoorvsadana: Hmm, should we move the build state update logic to a new worker then? (in a new PR maybe)_

tdelabro: Yep probably.
We will need to do so to provide SNOS with an input.
I will create it after this one is merged

PR #1589, opened 2024-05-04
version constants to pallet constants

apoorvsadana:


crates/node/Cargo.toml
```diff
@@ -128,3 +128,5 @@ sharingan = []
 sn-block-import = ["dep:mc-starknet-block-import"]
 # Enables Sierra class verification for both own and foreign blocks only + enabled for manual sealing mode
 sn-block-import-testing = ["sn-block-import"]
+# Enables dev features to improve DevX
+dev = []
```
tdelabro, line 132: how can we make sure that `dev` is never allowed in production?

tdelabro: Maybe with this: https://doc.rust-lang.org/stable/reference/conditional-compilation.html?highlight=debug_assertions#debug_assertions

tdelabro (changes requested):

crates/node/src/commands/run.rs
```diff
@@ -72,7 +72,10 @@ fn override_dev_environment(cmd: &mut ExtendedRunCmd) {
 
     cmd.base.force_authoring = true;
     cmd.base.alice = true;
-    cmd.base.tmp = true;
+
+    if cmd.base.shared_params.base_path.is_none() {
+        cmd.base.tmp = true;
+    }
```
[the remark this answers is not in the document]
apoorvsadana: oh forgot to mention this, it's unrelated to the PR but currently, if we specify a base-path and the --dev flag, then the base-path gets overridden --dev => --tmp. So I added a check to only include --tmp if there's no base-path provided.

crates/node/Cargo.toml
apoorvsadana: Do you mean we check if it's an optimized build and add an assertion that the dev flag does not exist? (like the opposite of debug_assert!). If yes, then I am not sure if that works well, because the Madara CLI will download optimized builds because they're way faster to run but it will have the dev flag enabled. I don't think there's a specific way to know that the user is running the chain in production, I think we can do

- add a big warning message whenever someone uses the Dev feature flag
- add another flag `production` and enfore that the user passes ONLY one

tdelabro: A solution would me to replace all `#[cfg(feature = "debug")]` buy `#[cfg(all(debug_assertions, feature = "debug"))]` this way we will have 3 situation

- not(debug): use prod code
- debug && debug_assertions: use debug guarded code
- debug without debug_assertions (aka: --release --features=debug): can't compile


Wdyt?

apoorvsadana: But we want `--release --features=debug` to compile for the CLI tool. We want to use release builds as debug builds are very slow but we want to enable dev features to make testing easier.

tdelabro: > we want --release --features=debug

Nope, that is the very thing we want to avoid

tdelabro: We have to find another way to make the CLI work, using the `not(debug)` settings

apoorvsadana: Ya I understand that in production we don't want people to be able to run with the --release flag. However, we want to make DevX easier with Madara. It's really bad right now where you need to clone Madara, build it locally so that you can finally test it. We need built binaries that people can directly download and start testing locally. Now, handing out debug binaries ins't good because they're really slow and won't show the power of Madara. So we need a way to hand out release binaries with debug features.

tdelabro: They are not that slow? Are they?

> won't show the power of Madara.

That is not what debug builds are for. They are for testing.


Test with debug binary, and prohibit dangerous features in release mode.
In the same idea I'm trying to protect against `testing` features being release out of `test` builds here: https://github.com/starkware-libs/blockifier/issues/430
I know it sounds a bit rigid, but I personally thinks it's for the greater good on the long run

tdelabro: We should not expose such an easy and convenient way to fuck up your chain. People will use it.

tdelabro: The DevX for fucking you chain should be as difficult as possible, so they really have to want it to do it.
Here they will do it inadvertently. 

And I put production security above the convenience of testing in the list of priorities. Especially for a blockchain project.

apoorvsadana: I get your point but the debug build is pretty slow when you compare it to the release build. I get your point that in the current context, it's very easy to pass the dev feature flag in release mode and start your chain. 

What if we go ahead with your idea of restricting the use of `dev` feature flag in `--release` builds. But we also add another feature flag called `dev-for-release` that let's you override this restriction. This way you specifically need to enable the feature `dev-for-release` to allow this  behaviour. We would like to use this in the CLI tool but it will have a huge warning message when you use it to switch off this feature flag for production.

tdelabro: > debug build is pretty slow when you compare it to the release build

How slow? And does it really matter? 
It's for local nodes and devs manually sending TXs, not for stress testing

apoorvsadana: > How slow?

It's about 10x slower. With the release build, Madara could do 120 TPS, in debug, it maxed at 12. This is just transfers by the way. Even getting responses from the RPC is much slower. Add invoke txs is <10ms in release and 100ms in debug. 
Screenshot 2024-05-06 at 7 20 46 PM

> And does it really matter

It depends on the application type I guess. If I am building a game with every interaction causing a transaction (and more complicated txs than transfer), then I could be getting 2-3 TPS in debug mode which can affect my development experience. The typing game we made in Madara, for example, might have lagged here. I completely get your security concern, and it makes sense. But I feel an additional feature flag to purposefully allow dev features in production + a BIG warning message on start can solve this. wdyt?

tdelabro: > But I feel an additional feature flag to purposefully allow dev features in production + a BIG warning message on start can solve this. wdyt?

To me, it's just back to step one with `feature = "dev"`

If people really need the perf for the development, they can compile to `--release` each time they modify their versionedConstants. It won't happen 100x.

Or they can compile to debug for the time they are trying different config values, and go back to `--release` when they go back to their regular dev.

But really, `--release --features=debug` this should be allowed.

apoorvsadana: The issue is the compiling it is difficult. I am imagining users who will use madaraup or madara cli and just download the binary. They might not even have Rust installed. And increasing the step limit is a common demand for app chains.

apoorvsadana: Hmm, cool I get your point. Let's do this then. I have pushed a commit where I check in the `build.rs` file itself and panic with a custom error message if someone uses the dev feature flag in release mode.

crates/runtime/src/pallets.rs
```diff
@@ -1,9 +1,16 @@
 //! Configuration of the pallets used in the runtime.
 //! The pallets used in the runtime are configured here.
 //! This file is used to generate the `construct_runtime!` macro.
+#[cfg(feature = "dev")]
```
tdelabro, line 4: But you declared the `dev = []` feature in the `node/Cargo.toml`, not in this crate.

I think you have to add `dev = []` to `runtime/Cargo.toml` and change the `node/Cargo.toml` to `dev = [madara-runtime/dev]`

tdelabro: Otherwise this will always be ignored

tdelabro (changes requested):

crates/runtime/src/pallets.rs
```diff
@@ -32,6 +40,29 @@ use crate::*;
 // --------------------------------------
 // CUSTOM PALLETS
 // --------------------------------------
+const EXECUTION_CONSTANTS_STR: &str = include_str!("../resources/versioned_constants.json");
+
+#[cfg(all(not(debug_assertions), not(feature = "dev")))]
```
tdelabro, line 45: `#[cfg(not(all(debug_assertions, feature = "dev")))]`

tdelabro (changes requested):

tdelabro (approved):

crates/node/build.rs
```diff
@@ -1,6 +1,18 @@
+use std::env;
+
 use substrate_build_script_utils::{generate_cargo_keys, rerun_if_git_head_changed};
 
 fn main() {
+    // Check if the feature flag is enabled
+    let feature_enabled = env::var("CARGO_FEATURE_DEV").is_ok();
+    // Check if we are in release mode
+    let release_mode = env::var("PROFILE").map(|p| p == "release").unwrap_or(false);
```
EvolveArt, line 9: we should also check for production mode which is the one actually used no?

crates/runtime/src/pallets.rs
```diff
@@ -151,12 +183,17 @@ impl pallet_timestamp::Config for Runtime {
     type WeightInfo = ();
 }
 
+fn get_execution_constants() -> Arc<VersionedConstants> {
+    EXECUTION_CONSTANTS.deref().clone()
```
EvolveArt, line 187: do we need to clone?

crates/node/build.rs
apoorvsadana: ahh good catch, changed it so that it works on debug mode only

crates/runtime/src/pallets.rs
apoorvsadana: yep, otherwise it's the ref and doesn't work

EvolveArt (approved):

Issue #1571, opened 2024-04-26
feat: support for STRK token as payment


tdelabro:
# Feature Request
Users should be able to pay their transaction using the STRK token

**Describe Preferred Solution**

1. Deploy the ERC20 smart contract at genesis, with correct storages values being set
2. Write pallet tests making sure that an user with 0 ETH but some STRK can submit TXs and be charged for it

**Related Code**

Edit the `set_infinite_tokens` (crates/pallets/starknet/src/tests/mod.rs) method so that it supports both ETH and STRK.

Both `configs/genesis-assets/genesis.json` and `crates/pallets/starknet/src/tests/mock/genesis.json` should be modified

**Additional Context**

**If the feature request is approved, would you be willing to submit a PR?**
_(Help can be provided if you need assistance submitting a PR)_

Ugo-X:
please can i take this?

j1mbo64:
Hi, can i take this one ? @tdelabro

tdelabro:
Sure @j1mbo64

tdelabro:
Ask for help if you need on the Madara channel :)

I think you can do some TDD for this one, coz I don't know exactly what the changes should be tbh

PR #1597, opened 2024-05-08
feat: support strk as fee token

j1mbo64:
# Pull Request type

- Feature

Resolves: #1571

## What is the new behavior?

- Invoke V1 transaction continue to consume ETH as fee token.
- Invoke V3 transaction consume now STRK as fee token.

No.


configs/genesis-assets/genesis.json
```diff
@@ -118,6 +118,10 @@
       "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
       "0x0372ee6669dc86563007245ed7343d5180b96221ce28f44408cff2898038dbd4"
     ],
+    [
+      "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
+      "0x0372ee6669dc86563007245ed7343d5180b96221ce28f44408cff2898038dbd4"
```
tdelabro, line 123: Can you confirm, "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d" is the address of a ERC20 contract "0x0372ee6669dc86563007245ed7343d5180b96221ce28f44408cff2898038dbd4", pre-deployed?

configs/genesis-assets/genesis.json
```diff
@@ -244,6 +248,90 @@
       ],
       "0xffffffffffffffffffffffffffffffff"
     ],
+    [
```
tdelabro, line 251: What are we doing here?

(damn I hate json for not supporting comments)

crates/pallets/starknet/src/tests/mock/genesis.json
```diff
@@ -150,6 +150,10 @@
       "0x00000000000000000000000000000000000000000000000000000000000000AA",
       "0x06232eeb9ecb5de85fc927599f144913bfee6ac413f2482668c9f03ce4d07922"
     ],
+    [
+      "0x00000000000000000000000000000000000000000000000000000000000000BB",
+      "0x06232eeb9ecb5de85fc927599f144913bfee6ac413f2482668c9f03ce4d07922"
```
tdelabro, line 155: Why isn't the class hash the same as in `genesis-assets/genesis.json`?

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -617,3 +620,285 @@ fn storage_changes_should_revert_on_transaction_revert() {
         assert_eq!(balance_value, vec![Felt252Wrapper::ZERO])
     })
 }
+
+#[test]
+fn given_hardcoded_contract_run_invoke_v1_without_fees_token_it_fails() {
```
tdelabro, line 625: this test covers something already covered by "given_hardcoded_contract_run_invoke_v1_without_eth_with_strk_fee_token_it_fails".

The only way it would behave differently would be the thing making it fail with the first one were having strk, and not having strk would be the thing making it succeed.
It doesn't make a lot of sense, so you can just remove this one

crates/pallets/starknet/src/tests/invoke_tx.rs
```diff
@@ -617,3 +620,285 @@ fn storage_changes_should_revert_on_transaction_revert() {
...
+
+        assert_ok!(Starknet::invoke(none_origin, transaction));
+        let final_balance_vec = get_balance_contract_call(sender_address, strk_fee_contract_address);
+        assert!(final_balance_vec[1] == initial_balance_vec[1]);
+        assert!(final_balance_vec[0] < initial_balance_vec[0]);
+    });
+}
+
+#[test]
+fn given_hardcoded_contract_run_invoke_v3_without_strk_with_eth_fee_token_it_fails() {
```
tdelabro, line 812: but my understranding was that currently on mainnet it is possible to pay with both ETH and STRK, why does it fails here?

crates/pallets/starknet/src/tests/mod.rs
```diff
@@ -78,6 +79,41 @@ pub fn get_invoke_dummy(
...
+    let tx = starknet_api::transaction::InvokeTransaction::V3(starknet_api::transaction::InvokeTransactionV3 {
+        resource_bounds: create_resource_bounds(),
+        tip: starknet_api::transaction::Tip::default(),
+        calldata,
+        sender_address,
+        nonce,
+        signature,
+        nonce_data_availability_mode: DataAvailabilityMode::L1,
+        fee_data_availability_mode: DataAvailabilityMode::L1,
+        paymaster_data: starknet_api::transaction::PaymasterData(vec![StarkFelt::ZERO]),
```
tdelabro, line 108: https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#v3_transaction_fields

The spes says it should be empty

tdelabro: same for `account_deployment_data`

tdelabro: and `paymaster_data`

crates/pallets/starknet/src/tests/utils.rs
```diff
@@ -112,3 +116,58 @@ pub fn build_get_balance_contract_call(account_address: StarkFelt) -> (EntryPoin
...
 }
+
+pub fn create_resource_bounds() -> starknet_api::transaction::ResourceBoundsMapping {
+    let mut map = BTreeMap::new();
+    map.insert(
+        starknet_api::transaction::Resource::L1Gas,
+        starknet_api::transaction::ResourceBounds { max_amount: 10000, max_price_per_unit: 12000 },
+    );
+    map.insert(
+        starknet_api::transaction::Resource::L1Gas,
```
tdelabro, line 127: You are inserting two times with key == L1Gas. You should probably use l2Gas for the second, no?

crates/pallets/starknet/src/tests/utils.rs
```diff
@@ -112,3 +116,58 @@ pub fn build_get_balance_contract_call(account_address: StarkFelt) -> (EntryPoin
...
+        starknet_api::transaction::ResourceBounds { max_amount: 50000, max_price_per_unit: 31000 },
+    );
+    starknet_api::transaction::ResourceBoundsMapping(map)
+}
+
+fn stark_felt_to_field_element(felt: &StarkFelt) -> FieldElement {
+    // Can only fail if the 32 byte number is not smaller than the field's modulus,
+    // but StarkFelt already enforces that
+    FieldElement::from_bytes_be(&felt.0).unwrap()
+}
```
tdelabro, line 137: Remove this and uses Felt252Wrapper for conversions instead. It implement all the From/Into you need

crates/pallets/starknet/src/tests/utils.rs
```diff
@@ -112,3 +116,58 @@ pub fn build_get_balance_contract_call(account_address: StarkFelt) -> (EntryPoin
...
+
+pub fn get_balance_contract_call(
+    sender_address: ContractAddress,
+    contract_address: ContractAddress,
+) -> Vec<Felt252Wrapper> {
+    let call_args = build_get_balance_contract_call(sender_address.0.0);
+    Starknet::call_contract(contract_address, call_args.0, call_args.1).unwrap()
+}
```
tdelabro, line 145: It is just used one in one test, no?
Just write the call there and don't factorize yet

crates/pallets/starknet/src/tests/utils.rs
```diff
@@ -112,3 +116,58 @@ pub fn build_get_balance_contract_call(account_address: StarkFelt) -> (EntryPoin
...
+
+pub fn set_sender_erc20_balance_to_zero(sender_address: ContractAddress, erc20_contract_address: ContractAddress) {
```
tdelabro, line 147: No reason to use the term `sender` rather than `account` in this context.

crates/pallets/starknet/src/tests/utils.rs
```diff
@@ -112,3 +116,58 @@ pub fn build_get_balance_contract_call(account_address: StarkFelt) -> (EntryPoin
...
+    // Set balance storage value to zero
+    StorageView::<MockRuntime>::insert(balance_low_storage_key, StarkFelt::try_from("0").unwrap());
+    StorageView::<MockRuntime>::insert(balance_high_storage_key, StarkFelt::try_from("0").unwrap());
+
+    // Ensure that balance is truely set to zero
+
+    pretty_assertions::assert_eq!(
+        get_balance_contract_call(sender_address, erc20_contract_address),
+        vec![Felt252Wrapper::from_hex_be("0x0").unwrap(), Felt252Wrapper::from_hex_be("0x0").unwrap()]
+    );
```
tdelabro, line 172: There is no need to ensure at everycall. Just to write a unit-test for your method. If it works once, it will will work the others times

tdelabro (changes requested):

configs/genesis-assets/genesis.json
j1mbo64: What do you mean by "pre-deployed" ? In Madara I understand the genesis.json content is deployed at start.
I took the Madara ETH address, i see it's the same on the Starknet mainnet, so i take the STRK address on the mainnet and i set it on genesis.
I misunderstand something ?

configs/genesis-assets/genesis.json
j1mbo64: I set the initial STRK balance for pre-deployed account as it was done for ETH.

crates/pallets/starknet/src/tests/invoke_tx.rs
j1mbo64: Ok i just wanted to be complete in my test, i totally agree

crates/pallets/starknet/src/tests/invoke_tx.rs
j1mbo64: In my understanding (I check it on mainnet with ArgentX), when you pay fees with STRK, TXv3 is always used, and if you pay with ETH, it's a TXv1.
Also, in transaction field, you can't specify the token used as fees.

configs/genesis-assets/genesis.json
tdelabro: It's fine, I think it can work

crates/pallets/starknet/src/tests/invoke_tx.rs
tdelabro: > In my understanding (I check it on mainnet with ArgentX), when you pay fees with STRK, TXv3 is always used, and if you pay with ETH, it's a TXv1.

Hoooooo that's how it works!!

@apoorvsadana you were aware of this behavior?

apoorvsadana: ya that was my doubt as well on the tg group, my next after this that when v1 is deprecated will STRK be on the only method of fee payment? or will we have proper paymasters by then?

tdelabro: I think paymaster is at the applicative level, can it be part of the protocol? Of a broader "fee market"?

apoorvsadana: I think paymaster will be in the protocol level but I haven't read a lot on this tbh. There's this post which shows the sequencer will need to be changed. -https://community.starknet.io/t/starknet-account-abstraction-model-part-2/839. And V3 txs also have paymaster data which isn't used now but might be used for this purpose eventually.

crates/pallets/starknet/src/tests/mock/genesis.json
j1mbo64: The ABI hash is used to compute the class hash. Madara use the ABI in `configs/genesis-assets/ERC20.json` and the Mock use `cairo-contracts/build/ERC20.json`. This files have the same content but the one used by the Mock have some more `\n`. You can check with `diff -u` Do you want i open an issue to uniform this ?

crates/pallets/starknet/src/tests/utils.rs
j1mbo64: `get_balance_contract_call` is used 10 times in `invoke_tx.rs`.

crates/pallets/starknet/src/tests/mod.rs
j1mbo64: It is set to zero but it's typed like it is in `starknet-api`. You can check it here https://github.com/starkware-libs/starknet-api/blob/77ff5a30e924cfe7672c773b140fe16af47a07b1/src/external_transaction_test.rs#L57

crates/pallets/starknet/src/tests/mock/genesis.json
tdelabro: Yeah you can open an issue and then close this conversation. 
Thanks a lot

crates/pallets/starknet/src/tests/mod.rs
tdelabro: What I meant is that the value should be `starknet_api::transaction::PaymasterData(vec![])`

crates/pallets/starknet/src/tests/utils.rs
tdelabro: okay fine then

crates/pallets/starknet/src/tests/utils.rs
j1mbo64: I have set this `FieldElement::from_bytes_be(&felt.0).unwrap()` in my function. The `get_storage_key` func want a `FieldElement`. I have a `ContractAddress`, the From Trait of `Felt252Wrapper` for `PatriciaKey` exist but i can't retrieve PatriciaKey from my ContractAddress but i can retrieve `StarkFelt`. I think i have do the best with existing code, say me if i can do better

tdelabro (approved):

Issue #1484, opened 2024-02-29
feat(logs): Display Madara logs instead of Substrate logs


antiyro:
# Feature Request

**Describe the Feature Request**

Currently the problem is that we display logs from Substrate on block sync and concensus session:

<img width="846" alt="Capture d’écran 2024-02-29 à 19 10 21" src="https://github.com/keep-starknet-strange/madara/assets/74653697/4738e259-a24c-40e9-9f56-6068b9dd9889">

We should display Madara logs directly instead. The problem is that those logs appear to be displayed within the substrate lib [here](https://github.com/paritytech/substrate/blob/033d4e86cc7eff0066cd376b9375f815761d653c/client/basic-authorship/src/basic_authorship.rs#L572), so we should either override them or do some changes within our Substrate fork.

akhercha:
Hey, happy to take this one!

antiyro:
https://github.com/KasarLabs/deoxys/issues/140

antiyro:
@akhercha yes ser, please check the issue above and lmk if you need any help

akhercha:
Hey guys!
Really sorry but it's pretty hard for me to allocate time for this - I'm un-assigning myself so you can move on with this one!

azurwastaken:
I can take it

jbcaron:
edit link: https://github.com/KasarLabs/deoxys-fork/issues/140

DADADAVE80:
Hello @antiyro , can I work on this if it isn't fixed yet?

PR #1609, opened 2024-05-24
feat(logs): Display Madara logs instead of Substrate logs

azurwastaken:
# Pull Request type

- Feature

Currently madara print substrate logs

Resolves: #1484

## What is the new behavior?

- Added the same logs as substrate but with Starknet related stuff

No

## Other information

I only added Starknet related logs because I don't know what to do with substrate logs, should I remove them directly from the massa fork ?


antiyro (approved):

tdelabro:
I don't know what was the goal here, but none of this makes sense.
Those are the wrong logs, to the wrong place.
This service just dump some useful values into our backend. It has nothing to do with block production or consensus. I don't know why you put those logs threre

azurwastaken:
> I don't know what was the goal here, but none of this makes sense. Those are the wrong logs, to the wrong place. This service just dump some useful values into our backend. It has nothing to do with block production or consensus. I don't know why you put those logs threre

Following #1484 , the goal is to provide starknet/madara related data instead of substrate ones because the substrate datas but maybe there is a better place to put them in madara. In fact i just did as discussed in Telegram. 

Could you please show me where you think it should be ? Because, as mentioned in telegram, its an absolute pain to integrate it in the polkadot sdk fork and it kill the perf to retrieve starknet data within it.

tdelabro:
>  its an absolute pain to integrate it in the polkadot sdk fork 

If you want log about the consensus, you should put them in the part of the code that handles the consensus, and that is the polkadot SDK

I don't see a good way to work around the fact that the logs should be where the things are happening.

tdelabro:
>  it kill the perf to retrieve starknet data within it.

Logs are informative stuff, you consult once in a while. Perf should not be at stake when we talk about reading logs.
What are you trying to achieve with those logs exactly?

PR #1611, opened 2024-05-27
feat(root): Adding state commitments computation using the bonsai lib

antiyro:
draft pr to implement commtiments


tdelabro (changes requested):

tdelabro:

PR #1617, opened 2024-05-31
feat: Declare V0 RPC call

ocdbytes:
Added a RPC method to declare a V0 contract using the program and declare transaction using the 0x1 account.

# Pull Request type
Feature

## What is the current behaviour?

Declaring V0 using the 0x1 account is not possible from empty state in madara

## What is the new behaviour?

RPC call is introduced for declaring the V0 contract using the program and declare call. Syntax is as follows : 

```rust
/// Broadcasted declare contract transaction v0.
#[derive(Debug, Eq, PartialEq)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub struct BroadcastedDeclareTransactionV0 {
    /// The address of the account contract sending the declaration transaction
    pub sender_address: FieldElement,
    /// The maximal fee that can be charged for including the transaction
    pub max_fee: FieldElement,
    /// Signature
    pub signature: Vec<FieldElement>,
    /// The class to be declared
    pub contract_class: Arc<CompressedLegacyContractClass>,
    /// If set to `true`, uses a query-only transaction version that's invalid for execution
    pub is_query: bool,
}

let contract_abi_artifact: CompressedLegacyContractClass = serde_json::from_reader(
                std::fs::File::open(env!("CARGO_MANIFEST_DIR").to_owned() + "/" + &artifact_path).unwrap(),
            ).unwrap();

let params: BroadcastedDeclareTransactionV0 = BroadcastedDeclareTransactionV0 {
    sender_address: FieldElement::from_hex_be("0x1").unwrap(),
    max_fee: FieldElement::from(482250u128),
    signature: Vec::new(),
    contract_class: Arc::new(contract_abi_artifact),
    is_query: false,
};

let json_body = &json!({
    "jsonrpc": "2.0",
    "method": "madara_addDeclareTransactionV0",
    "params": [params],
    "id": 4
});

let req_client = reqwest::Client::new();
let raw_txn_rpc = req_client.post(url).json(json_body).send().await;
```

No


.gitignore
```diff
@@ -44,3 +44,7 @@ cairo_venv
 
 # solidity cache
 tmp/
+
+# Setup deps
+setup_and_launch.sh
+./setup_dir
```
apoorvsadana, line 50: do we need this in the madara gitignore? looks like specific files for your use case instead of common ignored files

crates/client/l1-gas-price/src/worker.rs
```diff
@@ -66,6 +66,8 @@ async fn update_gas_price(
         .json()
         .await?;
 
+    // log::info!(">>>>>> gas_prices : {:?}", fee_history);
```
apoorvsadana, line 69: let's delete comment

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +41,25 @@ pub struct PredeployedAccountWithBalance {
...
+    pub class_hash: ClassHash,
+}
+
 /// Madara rpc interface for additional features.
 #[rpc(server, namespace = "madara")]
 pub trait MadaraRpcApi: StarknetReadRpcApi {
     #[method(name = "predeployedAccounts")]
     fn predeployed_accounts(&self) -> RpcResult<Vec<PredeployedAccountWithBalance>>;
+
+    #[method(name = "declarev0")]
```
apoorvsadana, line 55:
```suggestion
    #[method(name = "declareV0")]
```

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +41,25 @@ pub struct PredeployedAccountWithBalance {
...
+    async fn declare_v0_contract(
```
apoorvsadana, line 56: let's add a comment for the issue we were facing when sending the class info directly

crates/client/rpc-core/src/lib.rs
```diff
@@ -189,3 +208,16 @@ pub trait StarknetTraceRpcApi {
...
     async fn trace_transaction(&self, transaction_hash: FieldElement) -> RpcResult<TransactionTrace>;
 }
+
+pub enum DeclareTransactionCommonInput {
+    V0(DeclareTransactionV0V1, Vec<u8>, IndexMap<EntryPointType, Vec<EntryPoint>>, usize),
+    V1(DeclareTransaction),
+}
+
+#[async_trait]
+pub trait StarknetRpcApiCommonFuncs {
```
apoorvsadana, line 218: I was thinking we can skip the trait for this altogether and just do `impl Starknet`. The reason was that we already have other implementations which are commonly accessed and just want this to move to a new impl block because it has different constraints on the generics

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,101 @@ where
...
+    BE: Backend<B> + 'static,
+    C: HeaderBackend<B> + BlockBackend<B> + StorageProvider<B, BE> + 'static,
+    C: ProvideRuntimeApi<B>,
+    C::Api: StarknetRuntimeApi<B> + ConvertTransactionRuntimeApi<B>,
+    G: GenesisProvider + Send + Sync + 'static,
+    H: HasherT + Send + Sync + 'static,
+{
+    async fn declare_txn_common(
+        &self,
+        transaction_inputs: DeclareTransactionCommonInput,
```
apoorvsadana, line 244: we can move `DeclareTransactionCommonInput` in `rpc/types` if we remove the trait as mentioned aboved

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,101 @@ where
...
+                }
+
+                let extrinsic =
+                    self.convert_tx_to_extrinsic(best_block_hash, AccountTransaction::Declare(transaction)).unwrap();
+
+                submit_extrinsic(self.pool.clone(), best_block_hash, extrinsic).await.ok()?;
+
+                Some((tx_hash, class_hash))
+            }
+            DeclareTransactionCommonInput::V0(declare_txn, program_vec, entrypoints, abi_length) => {
```
apoorvsadana, line 270: what if we made a custom `DeclareV0Transaction` type which has all these fields? i think it might be cleaner/maintainable, wdyt?

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,101 @@ where
...
+                    &blockifier::execution::contract_class::ContractClass::V0(ContractClassV0(Arc::from(
+                        ContractClassV0Inner {
+                            program: program_decoded,
+                            entry_points_by_type: entrypoints,
+                        },
+                    ))),
+                    0,
+                    abi_length,
+                )
+                .unwrap();
```
apoorvsadana, line 282: we should avoid all unwraps here, try to throw a specific StarknetRpcError wherever possible and if nothing matches then InternalServerError. do note, we can't add new error fields to `StarknetRpcApiError` unless they don't match the starknet specn here - https://github.com/starkware-libs/starknet-specs

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,101 @@ where
...
+
+                let extrinsic =
+                    self.convert_tx_to_extrinsic(best_block_hash, AccountTransaction::Declare(transaction)).unwrap();
+
+                submit_extrinsic(self.pool.clone(), best_block_hash, extrinsic).await.ok()?;
+
+                Some((tx_hash, class_hash))
+            }
+            DeclareTransactionCommonInput::V0(declare_txn, program_vec, entrypoints, abi_length) => {
+                let txn_hash: TransactionHash = TransactionHash(StarkHash { 0: FieldElement::ONE.to_bytes_be() });
```
apoorvsadana, line 271: we need to fix this i guess?

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,101 @@ where
...
+                .unwrap();
+
+                let declare_transaction = DeclareTransaction::new(
+                    starknet_api::transaction::DeclareTransaction::V0(declare_txn),
+                    txn_hash,
+                    class_info,
+                )
+                .unwrap();
+
+                let current_block_hash = self.get_best_block_hash();
```
apoorvsadana, line 291: everything after this is common for both V0 and V1 I guess. if we handle the V0 and V1 entirely differently, then it technically beats the purpose of the common function

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,101 @@ where
...
+                }
+
+                let extrinsic = self
+                    .convert_tx_to_extrinsic(best_block_hash, AccountTransaction::Declare(declare_transaction.clone()))
+                    .unwrap();
+
+                let res = submit_extrinsic(self.pool.clone(), best_block_hash, extrinsic).await.unwrap();
+
+                log::info!(">>>>> extrinsic res : {:?}", res);
+                log::info!(">>>>> class hash declared: {:?} ✅ | txn hash : {:?}", declare_transaction.class_hash(), declare_transaction.tx_hash);
```
apoorvsadana: we can remove these logs. if you really want it, we can maybe make it a trace log.

apoorvsadana: also, I just noticed we don't add a log target to any logs in the client like we do here - https://github.com/apoorvsadana/madara/blob/main/crates/pallets/starknet/src/lib.rs#L101-L101

can you create a new issue to do this for madara logs in client as well, thanks!

crates/client/rpc/src/lib.rs
```diff
@@ -262,6 +360,20 @@ where
...
     }
+
+    async fn declare_v0_contract(
+        &self,
+        declare_transaction: DeclareTransactionV0V1,
+        program_vec: Vec<u8>,
+        entrypoints: IndexMap<EntryPointType, Vec<EntryPoint>>,
+        abi_length: usize
+    ) -> RpcResult<DeclareV0Result> {
+        let (_txn_hash, class_hash) = self.declare_txn_common(DeclareTransactionCommonInput::V0(declare_transaction, program_vec, entrypoints, abi_length))
```
apoorvsadana: i guess we should return the txn hash as well

crates/pallets/starknet/src/lib.rs
```diff
@@ -566,11 +566,17 @@ pub mod pallet {
...
             );
-            // Check if contract is deployed
-            ensure!(
-                ContractClassHashes::<T>::contains_key(transaction.tx().sender_address()),
-                Error::<T>::AccountNotDeployed
-            );
+
+            let tx_version = transaction.tx().version();
+
+            // Checking if transaction version is zero. Because it was causing issues with contract declaration.
```
apoorvsadana, line 572:
```suggestion
            // Avoiding account checks for V0 declares as they don't need to originate from an existing account
```

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +41,36 @@ pub struct PredeployedAccountWithBalance {
...
 
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DeclareV0Result {
+    pub txn_hash: TransactionHash,
+    pub class_hash: ClassHash,
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum DeclareV0Transaction {
```
apoorvsadana, line 52:
```suggestion
pub enum DeclareTransaction {
```

My idea was actually

1. Have a CustomDeclareV0Transaction type like this

```rust
struct CustomDeclareV0Transaction {
  declare_tx: DeclareTransactionV0V1,
  program: Vec<u8>
  ...
```

Since this is our custom type it will mostly work with serde (if it doesn't we can skip). We will use `CustomDeclareV0Transaction` as an input for `declare_v0_contract` RPC.

2. Have a `DeclareV0Transaction ` type inside `rpc` not `rpc-core` as its not needed in RPC. Also we can rename it to `DeclareTransaction` and instead of a tuple for V0 it will contain CustomDeclareV0Transaction

crates/client/rpc/src/lib.rs
apoorvsadana: this is still pending

crates/client/rpc/src/lib.rs
apoorvsadana: this is still pending

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +41,41 @@ pub struct PredeployedAccountWithBalance {
...
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct CustomDeclareV0Transaction {
+    pub declare_transaction: DeclareTransactionV0V1,
+    pub program_vec: Vec<u8>,
+    pub entrypoints: IndexMap<EntryPointType, Vec<EntryPoint>>,
+    pub abi_length: usize
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum DeclareTransactionWithV0 {
```
apoorvsadana, line 60: this type doesn't need to be in core as it's not being used here right?

crates/client/rpc/src/lib.rs
apoorvsadana: still pending

apoorvsadana: bumping this up again, we were going to do the transaction hash thing right?

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,97 @@ where
...
+        transaction_inputs: DeclareTransactionWithV0,
+    ) -> Result<(TransactionHash, ClassHash), StarknetRpcApiError> {
+        match transaction_inputs {
+            DeclareTransactionWithV0::V1(transaction) => {
+                self._declare_tx(transaction).await
+            }
+            DeclareTransactionWithV0::V0(declare_txn, program_vec, entrypoints, abi_length) => {
+                let txn_hash: TransactionHash = TransactionHash(StarkHash { 0: FieldElement::ONE.to_bytes_be() });
+
+                let program_decoded = Program::decode_all(&mut &*program_vec).unwrap();
```
apoorvsadana, line 252: unwrap

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,97 @@ where
...
+                    ))),
+                    0,
+                    abi_length,
+                );
+
+                match try_class_info {
+                    Ok(val) => {
+                        class_info = val;
+                    }
+                    Err(_e) => return Err(StarknetRpcApiError::InternalServerError),
```
apoorvsadana, line 269:
```rust
let class_info = ClassInfo::new(
                    &blockifier::execution::contract_class::ContractClass::V0(ContractClassV0(Arc::from(
                        ContractClassV0Inner { program: program_decoded, entry_points_by_type: entrypoints },
                    ))),
                    0,
                    abi_length,
                ).map_err(|e| {
                  log::debug!("error: ", e);
                  return Err(StarknetRpcApiError::InternalServerError
                })?
```

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,97 @@ where
...
+                }
+
+                let try_declare_transaction = DeclareTransaction::new(
+                    starknet_api::transaction::DeclareTransaction::V0(declare_txn),
+                    txn_hash,
+                    class_info,
+                );
+
+                match try_declare_transaction {
```
apoorvsadana, line 278: same as above

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +226,97 @@ where
...
+        }
+
+        let extrinsic = self
+            .convert_tx_to_extrinsic(best_block_hash, AccountTransaction::Declare(txn.clone()))
+            .unwrap();
+
+        let res = submit_extrinsic(self.pool.clone(), best_block_hash, extrinsic).await;
+
+        match res {
+            Ok(_val) => Ok((txn.tx.compute_hash(Felt252Wrapper::from(self.chain_id().unwrap().0), false), txn.class_hash())),
```
apoorvsadana, line 310: txn already has transaction_hash inside it right? we should call that and instead of hard coding it to ONE for V0 above, we can compute it there

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +41,41 @@ pub struct PredeployedAccountWithBalance {
...
+    V0(DeclareTransactionV0V1, Vec<u8>, IndexMap<EntryPointType, Vec<EntryPoint>>, usize),
```
apoorvsadana, line 61:
```suggestion
    V0(CustomDeclareV0Transaction),
```

crates/client/rpc/src/lib.rs
apoorvsadana: still pending

crates/client/rpc/src/lib.rs
ocdbytes: transaction hash ban rha h h but shuru me to build a transaction we need to provide some value to ye temp value h actual me transaction hash ban rha h if you check neeche

apoorvsadana: is it not possible to get the actual txn hash here? like we need `DeclareV0V1` to get the hash right?

crates/client/rpc/src/types.rs
```diff
@@ -25,6 +29,14 @@ pub enum ParseTokenError {
...
 }
 
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct CustomDeclareV0Transaction {
+    pub declare_transaction: DeclareTransactionV0V1,
+    pub program_vec: Vec<u8>,
+    pub entrypoints: IndexMap<EntryPointType, Vec<EntryPoint>>,
+    pub abi_length: usize,
+}
+
```
tdelabro, line 39: Don't you already declare this type in `crates/client/rpc-core/src/lib.rs`

crates/pallets/starknet/src/lib.rs
```diff
@@ -581,11 +581,17 @@ pub mod pallet {
...
+            let tx_version = transaction.tx().version();
+
+            // Avoiding account checks for V0 declares as they don't need to originate from an existing account
+            if tx_version != TransactionVersion::ZERO {
+                // Check if contract is deployed
+                ensure!(
+                    ContractClassHashes::<T>::contains_key(transaction.tx().sender_address()),
+                    Error::<T>::AccountNotDeployed
+                );
+            }
```
tdelabro, line 594: I think I did removed this check in #1627

tdelabro (changes requested):

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +40,32 @@ pub struct PredeployedAccountWithBalance {
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct DeclareV0Result {
+    pub txn_hash: TransactionHash,
+    pub class_hash: ClassHash,
+}
+
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub struct CustomDeclareV0Transaction {
+    pub declare_transaction: DeclareTransactionV0V1,
+    pub program_vec: Vec<u8>,
```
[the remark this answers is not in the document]
ocdbytes: Hey so at first I tried with ContractClassV0 only but there was issue in deserialising the program field so then after trial and error we thought that giving the vector and then converting the program will work.

crates/client/rpc/src/lib.rs
```diff
@@ -220,9 +224,55 @@ where
...
+    C: HeaderBackend<B> + BlockBackend<B> + StorageProvider<B, BE> + 'static,
+    C: ProvideRuntimeApi<B>,
+    C::Api: StarknetRuntimeApi<B> + ConvertTransactionRuntimeApi<B>,
+    G: GenesisProvider + Send + Sync + 'static,
+    H: HasherT + Send + Sync + 'static,
+{
+    fn _get_txn_hash(&self, transaction: DeclareTransactionV0V1) -> TransactionHash {
+        let txn = starknet_api::transaction::DeclareTransaction::V0(transaction);
+        txn.compute_hash(Felt252Wrapper::from(self.chain_id().unwrap().0), false)
+    }
```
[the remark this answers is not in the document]
ocdbytes: It is used I have changed it to `get_txn_hash`. Missed it my bad.

crates/client/rpc/src/types.rs
ocdbytes: I have this type in `crates/client/rpc-core/src/lib.rs` but it can't be imported into `crates/client/rpc/src/lib.rs` so had to define it explicitly here only.

crates/client/rpc-core/src/lib.rs
```diff
@@ -36,11 +40,32 @@ pub struct PredeployedAccountWithBalance {
...
+    pub entrypoints: IndexMap<EntryPointType, Vec<EntryPoint>>,
```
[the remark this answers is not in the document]
ocdbytes: added

crates/pallets/starknet/src/lib.rs
ocdbytes: I after rebasing it will be removed automatically

crates/client/rpc-core/src/lib.rs
```diff
@@ -47,6 +47,16 @@ pub struct DeclareV0Result {
     pub class_hash: ClassHash,
 }
 
+/// CustomDeclareV0Transaction
+/// This will be given as an input to the rpc body
+///
+/// **declare_transaction** : DeclareTransactionV0V1 struct constructed before calling.
+///
+/// **program_vec** : Program vector is program from `ClassInfo` struct converted into Vector.
```
tdelabro, line 55: `program_bytes`

crates/client/rpc-core/src/lib.rs
```diff
@@ -47,6 +47,16 @@ pub struct DeclareV0Result {
...
+///
+/// **entrypoints** : Entrypoints are taken from `ClassInfo` struct.
+///
+/// **abi_length** : ABI length calculated from abi in `LegacyContractClass`.
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
```
tdelabro, line 60: Do you really need `Eq, PartialEq`?
What for?

crates/client/rpc-core/src/lib.rs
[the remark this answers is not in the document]
tdelabro: What was the issue?

crates/client/rpc/src/types.rs
tdelabro: There is no reason it cannot be imported from rpc-core to rpc. 
Rpc depends on rpc-core

tdelabro (changes requested):

crates/client/rpc-core/src/lib.rs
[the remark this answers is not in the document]
ocdbytes: I don't know the exact issue. I tried to debug it by putting the logs It was saying that some field in program is not deserialised properly thus causing the error, so then only we moved to this solution.

crates/client/rpc/src/types.rs
ocdbytes: wait sorry It was other way around. fixed this in new commit.

crates/client/rpc-core/src/lib.rs
ocdbytes: do you want me to rename `program_vec` to `program_bytes` ?

tdelabro: vec is not really descriptive, bytes add a bit of information. So yes.

ocdbytes: done

tdelabro:
hey @ocdbytes, can you review this: https://github.com/ocdbytes/madara/pull/2
and then merge it into your branch please
It implements the rpc in a more coherent way, in regard to the rest of the codebase

crates/client/rpc-core/src/lib.rs
```diff
@@ -40,42 +37,17 @@ pub struct PredeployedAccountWithBalance {
...
 pub trait MadaraRpcApi: StarknetReadRpcApi {
     #[method(name = "predeployedAccounts")]
     fn predeployed_accounts(&self) -> RpcResult<Vec<PredeployedAccountWithBalance>>;
 
-    // There is an issue in deserialisation when we try to send the class info directly
-    // That's why we are sending the components saperately here and then building the
-    // transaction here in madara and executing the function in the pallet.
-    #[method(name = "declareV0")]
-    async fn declare_v0_contract(&self, params: CustomDeclareV0Transaction) -> RpcResult<DeclareV0Result>;
+    #[method(name = "addDeclareTransactionV0")]
```
tdelabro, line 46: Add doc

crates/client/rpc/Cargo.toml
```diff
@@ -37,6 +37,7 @@ sc-network-sync = { workspace = true }
 # Starknet
 blockifier = { workspace = true }
 cairo-vm = { workspace = true }
+indexmap = { workspace = true }
```
tdelabro, line 40: probably not used

crates/pallets/starknet/src/lib.rs
```diff
@@ -736,6 +736,23 @@ pub mod pallet {
 
             let transaction = Self::convert_runtime_calls_to_starknet_transaction(call.clone())
                 .map_err(|_| InvalidTransaction::Call)?;
+
```
tdelabro, line 739: Add a lil comment explaining that transactions v0 don't have Nonce or validation rules.

starknet-rpc-test/Cargo.toml
```diff
@@ -10,9 +10,13 @@ anyhow = "1.0.80"
 assert_matches = "1.5.0"
 async-lock = "3.1.0"
 async-trait = { workspace = true }
+blockifier = { workspace = true }
```
tdelabro, line 13: Not sure we nee all those new crates. Can you give it a look and cleanup the one that we are not using

tdelabro (changes requested):

tdelabro (approved):

PR #1621, opened 2024-05-31
dev: Make HasherT methods accept iterator

### LINKED ISSUE #1613: dev: make `HasherT` methods accept iterator rather than `&[Felt]` as argument
True for `compute_hash_on_wrapper` and `compute_hash_on_elements`.
Then check everywhere if it can allow us to remove some `.collect()` that were only there to respect the arg type.

Also pleases add a default impl for `compute_hash_on_wrappers` and make it call `compute_hash_on_elements` and remove the concrete impl of it for `poseidon` and `pedersen`.

fishseabowl:
# Pull Request type

- Feature

Resolves: #1613 

## What is the new behavior?

1. HasherT methods accept iterator
2. Add a default impl for compute_hash_on_wrappers
-
-
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next release
 
+- dev: make HasherT methods accept iterator rather than &[Felt] as argument
 - feat: actual estimate_fee added, brought back l1 messages and refactored
   simulate tx
 - dev: impl get_state_updates using get_transaction_re_execution_state_diff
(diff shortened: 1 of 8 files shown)


tdelabro (changes requested):
Let's wait for https://github.com/xJonathanLEI/starknet-rs/pull/594
before merging this one

tdelabro:
Okay, mb I didn't remember we were internally calling starknet-rs, which takes an `&[FieldElement]`.

For now let's just keep the default impl part of the PR and revert the IntoIterator one.
Once this is merged https://github.com/xJonathanLEI/starknet-rs/pull/594, we will add it

fishseabowl:
> Okay, mb I didn't remember we were internally calling starknet-rs, which takes an `&[FieldElement]`.
> 
> For now let's just keep the default impl part of the PR and revert the IntoIterator one. Once this is merged [xJonathanLEI/starknet-rs#594](https://github.com/xJonathanLEI/starknet-rs/pull/594), we will add it

Sure. Thank you for your review.

PR #1627, opened 2024-06-05
fix: remove unecessary checks before tx execution

tdelabro:
By removing those checks we will return same errors are Starknet rather than our customs ones (when we will be able to return complex errors from runtime)

### DIFF
--- a/crates/pallets/starknet/src/lib.rs
+++ b/crates/pallets/starknet/src/lib.rs
@@ -509,14 +509,6 @@ pub mod pallet {
             // This ensures that the function can only be called via unsigned transaction.
             ensure_none(origin)?;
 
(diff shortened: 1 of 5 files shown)


EvolveArt (approved):

crates/pallets/starknet/src/lib.rs
```diff
@@ -576,17 +568,6 @@ pub mod pallet {
             // This ensures that the function can only be called via unsigned transaction.
             ensure_none(origin)?;
 
-            // Check class hash is not already declared
-            ensure!(
```
apoorvsadana, line 580: so what happens if its already declared, does it fail or it just overrides?

apoorvsadana (approved):

PR #1637, opened 2024-06-24
Add getter pending block extrinsics

sydhds:
Fix retrieving pending transactions

# Pull Request type

- Bugfix

When retrieving pending transactions using RPC call, every transactions in transactionPool are retrieved.

Resolves: #NA

## What is the new behavior?

Now, only the transactions added by the BlockBuilder in order to create the next futur bllck are retrieved.

No

## Other information

N/A


tdelabro:

